PR #2, opened 2021-12-06
Merge overlapping ranges from input during stats collection

rkenmi:


pdames (approved):
Looks good, thanks!

pdames:
One corner case to either handle here or in a subsequent PR is that `None` specified an either side of the interval when calling `list_deltas` effectively represents an unbounded interval on that side (i.e. `None` at the start of the interval is equivalent to -infinity, while `None` at the end of the interval is equivalent to +infinity).

rkenmi:
> One corner case to either handle here or in a subsequent PR is that `None` specified on either side of the interval when calling `list_deltas` effectively represents an unbounded interval on that side (i.e. `None` at the start of the interval is equivalent to `-math.inf`, while `None` at the end of the interval is equivalent to `math.inf`).

Sounds good! Added a TODO for this corner case.

PR #10, opened 2022-01-31
[Storage] API Updates to add Ray Options Provider

pdames:
This PR changes existing APIs to inject `ray.remote` options providers when invoking storage APIs that will subsequently invoke concurrent Ray tasks for all blocks of a dataset.


rkenmi (approved):

Zyiqin-Miranda (approved):
LGTM, thanks!

Issue #3, opened 2021-12-10
Add infinity range support for stats collection range intervals


rkenmi:
https://github.com/ray-project/deltacat/blob/18c6c5f93a1e9c77bd3f794502a3dedbc179a0f1/deltacat/compute/stats/utils/intervals.py#L5

Currently the delta stats collector does not have any support for ranges with infinity. For example if we want to grab stream positions from (-inf, inf), then ideally the range would encompass all stream positions for this particular delta.

Ideas:
- Allow `None` as an input to represent the infinity range. For example `(None, None)` can represent (-inf, inf), and `(5, None)` can represent (5, inf).
- Document the behavior of how `None` is interpreted across `deltacat_storage` implementations

PR #12, opened 2022-04-01
[Stats] Add support for infinite range with delta stream positions

rkenmi:
Addresses #3 to allow clients to conveniently fetch stats for unbounded intervals

### DIFF
--- a/deltacat/compute/stats/basic.py
+++ b/deltacat/compute/stats/basic.py
@@ -9,7 +9,7 @@ from deltacat.compute.stats.models.stats_result import StatsResult
 from deltacat.compute.stats.types import StatsType
 from deltacat.compute.stats.utils.io import read_cached_delta_stats, cache_delta_column_stats, get_delta_stats, \
     get_deltas_from_range
-from deltacat.compute.stats.utils.intervals import merge_intervals, UnboundedRange
+from deltacat.compute.stats.utils.intervals import merge_intervals, DeltaRange
 
 from deltacat.storage import PartitionLocator, DeltaLocator, Delta
 from deltacat.storage import interface as unimplemented_deltacat_storage
@@ -21,7 +21,7 @@ from deltacat.storage import interface as unimplemented_deltacat_storage
 
 def collect(
         source_partition_locator: PartitionLocator,
-        delta_stream_position_range_set: Optional[Set[Tuple[UnboundedRange, UnboundedRange]]] = {(None, None)},
+        delta_stream_position_range_set: Optional[Set[DeltaRange]] = None,
         columns: Optional[List[str]] = None,
         stat_results_s3_bucket: Optional[str] = None,
         deltacat_storage=unimplemented_deltacat_storage) -> Dict[int, DeltaStats]:
@@ -36,7 +36,7 @@ def collect(
 
     Args:
         source_partition_locator: Reference to the partition locator tied to the given delta stream positions
-        delta_stream_position_range_set: A set of open bounded intervals with an int type representing finite,
+        delta_stream_position_range_set: A set of intervals with an int type representing finite,
             closed bounded values, and a None type representing unbounded infinity.
         columns: Columns can be optionally included to collect stats on specific columns.
             By default, all columns will be calculated.
@@ -46,6 +46,8 @@ def collect(
     Returns:
         A mapping of stream positions to their corresponding delta stats.
     """
+    if delta_stream_position_range_set is None:
+        delta_stream_position_range_set = {(None, None)}
     delta_stream_range_stats: Dict[int, DeltaStats] = {}
     delta_range_lookup_pending: List[ObjectRef[List[Delta]]] = []
 
--- a/deltacat/compute/stats/basic.py
+++ b/deltacat/compute/stats/basic.py
@@ -9,7 +9,7 @@ from deltacat.compute.stats.models.stats_result import StatsResult
 from deltacat.compute.stats.types import StatsType
 from deltacat.compute.stats.utils.io import read_cached_delta_stats, cache_delta_column_stats, get_delta_stats, \
     get_deltas_from_range
-from deltacat.compute.stats.utils.intervals import merge_intervals
+from deltacat.compute.stats.utils.intervals import merge_intervals, UnboundedRange
 
 from deltacat.storage import PartitionLocator, DeltaLocator, Delta
 from deltacat.storage import interface as unimplemented_deltacat_storage
@@ -21,7 +21,7 @@ from deltacat.storage import interface as unimplemented_deltacat_storage
 
 def collect(
         source_partition_locator: PartitionLocator,
-        delta_stream_position_range_set: Set[Tuple[int, int]],  # TODO: R&D on RangeSet impl
+        delta_stream_position_range_set: Optional[Set[Tuple[UnboundedRange, UnboundedRange]]] = {(None, None)},
         columns: Optional[List[str]] = None,
         stat_results_s3_bucket: Optional[str] = None,
         deltacat_storage=unimplemented_deltacat_storage) -> Dict[int, DeltaStats]:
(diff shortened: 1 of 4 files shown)


deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
-from typing import Set, Tuple
-from collections import Iterable
+from typing import Set, Tuple, List, Iterable, Union
 
+UnboundedRange = Union[int, None]
```
pdames, line 3:
```suggestion
UnboundedRange = Optional[int]
```

deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
...
+NumericalRange = Union[int, float]
```
pdames, line 5:
```suggestion
DeltaRange = Union[UnboundedRange, NumericalRange]
```

deltacat/compute/stats/basic.py
```diff
@@ -21,7 +21,7 @@
 
 def collect(
         source_partition_locator: PartitionLocator,
-        delta_stream_position_range_set: Set[Tuple[int, int]],  # TODO: R&D on RangeSet impl
+        delta_stream_position_range_set: Optional[Set[Tuple[UnboundedRange, UnboundedRange]]] = {(None, None)},
```
pdames, line 24:
```suggestion
        delta_stream_position_range_set: Optional[Set[Tuple[DeltaRange, DeltaRange]]] = {(None, None)},
```
Also, we should avoid assigning mutable collections as defaults since this can lead to some hard-to-troubleshoot bugs. One simple resolution is to just set this to `None` and assign the default in one of the first lines of the function impl.

deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
...
```
pdames, line 4: Minor: Maybe add a comment stating that float is only added to support math.inf?

deltacat/compute/stats/basic.py
```diff
@@ -36,8 +36,8 @@ def collect(
 
     Args:
         source_partition_locator: Reference to the partition locator tied to the given delta stream positions
-        delta_stream_position_range_set: Each delta stream position range is a closed bounded interval of integer
-            values. Currently, there is no support for infinity ranges but this is subject to change.
+        delta_stream_position_range_set: A set of open bounded intervals with an int type representing finite,
```
pdames, line 39:
```suggestion
        delta_stream_position_range_set: A set of open or closed bounded intervals with an int type representing finite,
```

deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
...
+    Each input interval is an open unbounded interval with infinite ranges.
+    Intervals will be merged into a minimal number of output intervals in O(N log N) time.
 
     Example:
         >>> merge_intervals((3, 9), (8, 12), (15, 19))
         ((3, 12), (15, 19))
 
+    Example:
+        >>> merge_intervals((3, 9), (None, 15), (13, 30))
+        ((None, 30))
```
pdames, line 20: Maybe add one more example showing that `math.inf` is also equivalent to None?

deltacat/compute/stats/utils/intervals.py
pdames: On second thought, I think it would make this code and subsequent code much more clear to just drop `math.inf` support which allows us to remove `float` from the signature here.

Then your whole typing here is also reduced down to just one type:
`DeltaRange = Optional[int]`

deltacat/compute/stats/utils/intervals.py
pdames: See above comment - let's just drop `math.inf` support. Seems to overcomplicate the code without giving back much of any tangible benefits.

deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
...
+
+def _to_numeric_values(intervals_list: List[Tuple[UnboundedRange, UnboundedRange]]):
+    for i, interval in enumerate(intervals_list):
+        start, end = _get_validated_interval(interval)
+        if start is None:
+            start = float('-inf')
+        if end is None:
+            end = float('inf')
+
+        intervals_list[i] = (start, end)
```
pdames, line 75: Let's drop this together with our support for `math.inf`.

rkenmi: I think we'll still need this conversion to sort and merge the intervals for a given None type value

deltacat/compute/stats/utils/intervals.py
rkenmi: I think we would still need the `float` to account for the numerical representation of None, to do the actual sort/merge. But I agree that the input and output signature of `merge_intervals` can be reduced down to `DeltaRange = Optional[int]`

deltacat/compute/stats/basic.py
rkenmi: Good catch, thanks 🙂

deltacat/compute/stats/utils/intervals.py
rkenmi: Yeah I agree that from a client caller perspective, `math.inf` doesn't seem as intuitive to pass in, as opposed to `None` to represent all values for a start/end interval.

deltacat/compute/stats/utils/intervals.py
rkenmi: The `merge_intervals` function internally converts `None` values to numerical values, to make use of the sort. After the merging process, the float (inf) types are converted back to `None`. This is done to be compliant with the `list_deltas` API, which accepts `Optional[int]` for starting and ending stream positions. 

So I think we could keep the `NumericalRange` type for internal conversions (None/infinity) but keep the function input/output types to be a simple `Optional[int]`.

deltacat/compute/stats/utils/intervals.py
pdames: Also, since all of these are really just one side of a range, it may be more clear to rename `UnboundedRange` to `DeltaPosition`, `NumericalRange` to `NumericDeltaPosition`, and then define `DeltaRange = Tuple[DeltaPosition, DeltaPosition]` for use in the top-level, user-facing APIs.

deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
...
-# TODO: Allow users to add infinite ranges in intervals: i.e. (5, None) => (5, math.inf)
-def merge_intervals(intervals: Set[Tuple[int, int]]):
+
+def merge_intervals(intervals: Set[DeltaRange]) -> Set[DeltaRange]:
     """Merges a set of N input intervals into a minimal number of output intervals.
 
-    Each input interval must be closed bounded intervals. Intervals will be merged into
-    a minimal number of closed bounded output intervals in O(N log N) time.
+    Each input interval is an open unbounded interval with infinite ranges.
+    Intervals will be merged into a minimal number of output intervals in O(N log N) time.
```
pdames, line 12:
```suggestion
    All input intervals will be merged into a minimal number of output intervals in O(N log N) time.
```

deltacat/compute/stats/utils/intervals.py
```diff
@@ -1,62 +1,88 @@
...
         ((3, 12), (15, 19))
 
+    Example:
+        >>> merge_intervals((3, 9), (None, 15), (13, 30))
+        ((None, 30))
+
     Args:
-        intervals: A list of closed bounded intervals
+        intervals: A list of open bounded intervals with an int type representing finite, closed bounded values and
+            a None type representing infinity.
```
pdames, line 24:
```suggestion
        intervals: A list of intervals with an int type representing finite, closed bounded values and
            a None type representing infinity.
```

pdames (approved):
Looks good overall! Just a couple minor comments on the docstrings. Feel free to merge when you're ready.

Zyiqin-Miranda:
LGTM, thanks for the supporting the unbounded delta stream positions!

Zyiqin-Miranda (approved):

PR #13, opened 2022-04-20
Add job event state listener and dispatcher

rkenmi:
This PR introduces general purpose utilities (event state listener and dispatcher) to transition Ray jobs from one state to another. It also includes an example driver script for deltacat compaction runs.

### Concepts

Event State Listener: Latest events are polled from the Event Summary DynamoDB table. Callbacks can be defined for an event's job state, with granularity down to the job state sequence level. In its simplest form, these callbacks can be used to invoke calls with the Event Dispatcher to transition from a previous job state to another job state.

Event Dispatcher: Publishes new events via SNS with the job state name and job state sequence.

--

### Pre-requisites
- [Amazon Ray with SNS Event Callback support](https://github.com/rkenmi/amazon-ray/tree/sns_events_with_script_support) (May be moved to a separate branch of the [Ray repo](https://github.com/ray-project/ray) in the future)

### Tested with
- Ray 1.11.0

### DIFF
--- a/deltacat/autoscaler/events/dispatcher.py
+++ b/deltacat/autoscaler/events/dispatcher.py
@@ -1,0 +1,141 @@
+import logging
+from typing import Dict, Any, Optional, Union, Callable
+
+from ray.autoscaler._private.aws.events import AwsEventManagerBase
+from ray.autoscaler._private.event_system import RayEvent, ScriptStartedEvent, ScriptInProgressEvent, \
+    ScriptInProgressCustomEvent, ScriptCompletedEvent, event_enum_values, States
+
+from deltacat import logs
+
+stats_metadata_collection_started_event = ScriptInProgressCustomEvent("STATS_METADATA_COLLECTION_STARTED", 2)
+stats_metadata_collection_completed_event = ScriptInProgressCustomEvent("STATS_METADATA_COLLECTION_COMPLETED", 3)
+compaction_session_started_event = ScriptInProgressCustomEvent("COMPACTION_SESSION_STARTED", 4)
+compaction_session_completed_event = ScriptInProgressCustomEvent("COMPACTION_SESSION_COMPLETED", 5)
+custom_events = [stats_metadata_collection_started_event,
+                 stats_metadata_collection_completed_event,
+                 compaction_session_started_event,
+                 compaction_session_completed_event]
+
+
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
+
+
+class EventDispatcher:
+    def __init__(self, events_manager: AwsEventManagerBase):
+        """Constructor for the event dispatcher.
+
+        Intended for usage by Ray parent and child clusters running BDT compaction workflows.
+        This helper class is a wrapper around the events_manager for dispatching compaction workflow events
+        annotated with event metadata for tracking purposes, such as parent and child Ray session IDs.
+
+        Args:
+            events_manager: Events manager for publishing events through a cloud provider
+        """
+        self.events_manager = events_manager
+
+    def dispatch_event(self,
+                       event: RayEvent,
+                       parent_session_id: Optional[str] = None,
+                       session_id: Optional[str] = None,
+                       event_data: Dict[str, Any] = None):
+        if event_data is None:
+            event_data = {}
+
+        event_data["event"] = event
+        event_data["rayParentSessionId"] = parent_session_id
+        event_data["raySessionId"] = session_id
+
+        self._publish_event({
+            **self.events_manager.config["parameters"],
+            **event_data
+        })
+
+    def start_initializing(self):
+        """Publish a job state event
+        """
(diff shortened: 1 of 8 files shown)


deltacat/compute/compact_driver.py
```diff
@@ -0,0 +1,58 @@
...
+    manager = AwsEventManager(events_config) if events_config else None
+    source_table_dict = events_config["parameters"]["sourceTable"]
+    b64_encoded_pkv = events_config["parameters"].get("partitionsToCompact")
+    if b64_encoded_pkv:
+        pkv = gzip.decompress(base64.b64decode(b64_encoded_pkv))
+        pkv_json = json.loads(pkv)
+        partition_values = [kv["value"] for kv in pkv_json[0]]
+    else:
+        logger.info(f"No partition key-values found to compact. Attempting partition discovery.")
+        partition_values = partition_discovery()
```
pdames, line 52: Is "job runner" or "job manager" maybe a more descriptive name than "driver"? 

Also, we should try to separate this into a generic job runner (e.g. for any type of job that may run against your data catalog), a generic compaction job runner (e.g. that can run a change-data-capture job for any type of data catalog that supports deltas), and a specific compaction job runner (e.g. the implementation that we have here that has very specific expectations about the event model, payload content encoding, etc.). 

At a high-level, it seems like a generic job runner just needs to take some sort of event manager, event store, and event dispatcher as input and it should return a "handle" to the job runner as output (e.g. ideally a Ray cluster session ID, distinct cluster name, or some other ID that can be used to monitor/manage this job runner instance through the Ray CLI or another external app/service).

deltacat/compute/compact_driver.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # TODO: pass partition_values to child jobs
+    ddb_client = boto3.client("dynamodb", "us-east-1")
+    event_store = EventStoreClient(ddb_client, COMPACTION_EVENT_SUMMARY_TABLE_NAME)
+    event_dispatcher = EventDispatcher(manager)
+    poll_events(manager, event_store, event_dispatcher)
```
pdames, line 58: While a single process event poller may be usable for a PoC, we should consider transitioning this into an autoscaling event poller running on a Ray cluster ASAP.

deltacat/autoscaler/events/workflow.py
```diff
@@ -0,0 +1,142 @@
...
+from deltacat.autoscaler.events.exceptions import EventNotFoundException
+
+logging.basicConfig(level=logging.INFO)
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
+
+QUERY_EVENTS_MAX_RETRY_COUNTER = 100
+SLEEP_PERIOD_SECONDS = 10
+
+
+def poll_events(events_manager: AwsEventManagerBase,
```
pdames, line 23: Can we separate the AWS Event Manager impl out from a version of this that works for any event manager? As part of this, we may also want to consider (1) putting this inside of a class that provides an expected set of state transitions per implementing job type (and making the state transition map a member of that class) and (2) renaming this from "workflow" to something more specific like "event processor" or "event poller".

deltacat/autoscaler/events/dispatcher.py
```diff
@@ -0,0 +1,141 @@
...
+        event_data["raySessionId"] = session_id
+
+        self._publish_event({
+            **self.events_manager.config["parameters"],
+            **event_data
+        })
+
+    def start_initializing(self):
+        """Publish a job state event
+        """
```
pdames, line 55: Can these docstrings include a description of the intent of each state being published (e.g. the initializing state indicates that a request to start a job run has been successfully received, but the job run has not yet finished prerequisite initialization steps required to make progress on the requested job).

pdames (changes requested):

deltacat/autoscaler/events/dispatcher.py
```diff
@@ -0,0 +1,141 @@
...
+                                "customEventName": compaction_session_completed_event.name,
+                                "customDescription": "Finished compaction run",
+                            })
+
+    def complete_job(self):
+        """Publish a job state event
+        """
+        self.dispatch_event(ScriptCompletedEvent.complete_success)
+
+    def build_state_transitions(self) -> Dict[str, Union[Callable[[], None], Dict]]:
```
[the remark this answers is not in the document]
rkenmi: It shouldn't be the dispatcher's responsibility to determine the workflow, so I'll separate this out.

deltacat/compute/compact_driver.py
rkenmi: These files are slightly outdated - they no longer exist in the [newer DeltaCAT fork](https://github.com/pdames/deltacat/pull/1/files) as I've moved these scripts into SungatePython.

rkenmi:
Closing due to stale PR

PR #23, opened 2022-12-14
Scaling test

valiantljk:
* added node group and placement group manager
* hardcoded num_cpus to 0.5 for most of the remote tasks (run into resource not available issue in scaling test when default request of cpus is 1)


requirements.txt
```diff
@@ -1,9 +1,12 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
-s3fs >= 2021.7.0
-tenacity >= 8.0.1
-ray == 1.10.0
-pandas >= 1.3.1
-pyarrow >= 5.0.0
-numpy >= 1.21.1
-boto3 == 1.20.40
+"s3fs == 2022.1.0",
```
pdames, line 3: Can we make sure we include all of the changes from https://github.com/ray-project/deltacat/commit/40900c81df315280e3e1cceecc51886677ec9fab here and in setup.py (we can also set the mainline version to 0.2.0 in __init__.py for now).

setup.py
```diff
@@ -35,15 +36,22 @@ def find_version(*paths):
...
+        "ray[default] == 2.1.0",
         "pandas >= 1.3.1",
-        "pyarrow >= 5.0.0",
+        "pyarrow == 8.0.0",
+        "pydantic == 1.10.2",
         "numpy >= 1.21.1",
-        "boto3 == 1.20.40"
+        "boto3 == 1.20.24",
+        "typing-extensions == 4.4.0",
+        "binpacking == 1.5.2"
```
pdames, line 49: It seems like this is only used for tests - can we move it into a separate requirements.txt under scalability tests?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -139,32 +176,58 @@ def _execute_compaction_round(
...
     logger.info(f"Total cluster resources: {cluster_resources}")
-    logger.info(f"Available cluster resources: {ray.available_resources()}")
-    cluster_cpus = int(cluster_resources["CPU"])
-    logger.info(f"Total cluster CPUs: {cluster_cpus}")
-
+    if pg_config: # use resource in each placement group
+        node_resource_keys=None
+        cluster_resources = pg_config[1]
+        cluster_cpus = cluster_resources['CPU']
+    elif node_group_res: # use resource in each node group
```
pdames, line 193: Is there value in hanging onto all of this code related to node groups? Are there any benefits here that we don't get from placement groups? Should it be broken out into a separate branch to help clean the code up here a bit?

pdames (changes requested):
As a high-level comment, it seems like a lot of the code under the "autoscaler" module is still in an experimental and/or broken state. Should we separate that out into a separate branch to be merged in once it's more stable? Also, I think we probably want to rethink the module name a bit. Most of the code there seems to be related to more generic, high-level data catalog job management, and has little or nothing to do with Ray's underlying autoscaler logic (aside from perhaps some pieces of code that help initiate a Ray cluster launch or predict cluster resources).

deltacat/compute/compactor/compaction_session.py
valiantljk: Yeah, the code can be simplified. Node group currently is not used. It could be useful when we want to design an autoscalable placement group, where we can leverage node group to set min and max for each placement group. But let me remove it for now. As an autoscalable pg may be designed in a different way.

setup.py
valiantljk: yes, binpacking can be removed for now.

requirements.txt
valiantljk: sure

requirements.txt
```diff
@@ -1,9 +1,12 @@
...
+"memray == 1.3.1",
```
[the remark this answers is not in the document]
valiantljk: memray is only used for profiling. Yes, it can be removed.

deltacat/autoscaler/events/compaction/dispatcher.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+from deltacat.autoscaler.events.compaction.session_manager import CompactionSessionManager
+from deltacat.autoscaler.events.dispatcher import EventDispatcher
+from ray.autoscaler._private.aws.events import EventPublisher
```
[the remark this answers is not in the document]
valiantljk: Just removed the autoscaler module, and added the placement group module into deltacat.utils

pdames (approved):
LGTM - thanks!

PR #26, opened 2022-12-22
[chore] Added first iteration of build and publish to pypi workflow

pfaraone:
To Be Written

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,44 @@
+name: Publish Python distributions to PyPI and TestPyPI
+# on: push - Runs your workflow when you push a commit or tag.
+on:
+  release:
+    types: [published] # triggered whenever a new GitHub release is published
+jobs:
+  build-n-publish:
+    name: Build and publish Python distributions to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@main
+      with:
+          fetch-depth: 0
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+    - name: Install pypa/build
+      run: >-
+        python -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    # - name: Publish distribution to Test PyPI
+    #   uses: pypa/gh-action-pypi-publish@release/v1 # PyPI publish GitHub Action
+    #   with:
+    #     password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+    #     repository_url: https://test.pypi.org/legacy/
+    #     skip_existing: true
+    - name: Publish distribution to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
+        
--- a/deltacat/storage/model/types.py
+++ b/deltacat/storage/model/types.py
@@ -1,13 +1,11 @@
-import pyarrow as pa
-import pandas as pd
-import numpy as np
-
 from enum import Enum
+from typing import Any, List, Union
 
+import numpy as np
+import pandas as pd
+import pyarrow as pa
 from ray.data.dataset import Dataset
 from ray.data.impl.arrow_block import ArrowRow
-
-from typing import Any, List, Union
 
 LocalTable = Union[pa.Table, pd.DataFrame, np.ndarray]
 LocalDataset = List[LocalTable]
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,8 @@
-import setuptools
+import logging
 import os
 import re
-import logging
+
+import setuptools
 
 logger = logging.getLogger(__name__)
 


deltacat/storage/model/types.py
```diff
@@ -1,13 +1,11 @@
...
-
 from enum import Enum
+from typing import Any, List, Union
 
+import numpy as np
+import pandas as pd
+import pyarrow as pa
+from ray.data._internal.arrow_block import ArrowRow
 from ray.data.dataset import Dataset
-from ray.data.impl.arrow_block import ArrowRow
```
pfaraone, line 8: from ray.data.impl.arrow_block import ArrowRow throws a module not found error when running the unit test cases

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+        python -m unittest
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    # - name: Publish distribution to Test PyPI
```
pfaraone, line 41: I don't currently have maintainer access to TestPyPI DeltaCAT [access to ](https://test.pypi.org/project/deltacat/)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
+name: Publish Python distributions to PyPI and TestPyPI
+on: push # Runs your workflow when you push a commit or tag.
+# on:
+#   release:
+#     types: [published] # triggered whenever a new GitHub release is published
```
raghumdani, line 5: Can we remove redundant comments?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,52 @@
...
+        build
+        --user
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Run unit tests
+      run: >-
+        python -m unittest
```
raghumdani, line 32: we are using pytest module, but here you are running using unittest module. Can you check?

.github/workflows/publish.yml
raghumdani: Maybe we don't need this right now.

setup.py
```diff
@@ -35,13 +36,15 @@ def find_version(*paths):
...
-        "boto3 == 1.20.40"
+        "s3fs == 2022.11.0",
+        "tenacity == 8.1.0",
+        "ray[default] == 2.0.0",
+        "pandas == 1.5.2",
+        "pyarrow == 10.0.1",
+        "pydantic == 1.10.2",
+        "numpy == 1.24.0",
+        "boto3 == 1.24.59",
+        "typing-extensions >= 4.4.0"
```
raghumdani, line 47:
```suggestion
  "s3fs >= 2022.1.0",
  "tenacity >= 8.0.1",
  "ray[default] == 2.0.0",
  "pandas >= 1.3.1",
  "pyarrow >= 8.0.0",
  "pydantic >= 1.10.2",
  "numpy >= 1.21.1",
  "boto3 >= 1.20.24",
  "typing-extensions >= 4.4.0"
```

raghumdani (commented):
Looks good overall. Let's see if we can make `requirements.txt` changes.

.github/workflows/publish.yml
pfaraone: Yes - sorry should have give more context here. Currently we're running on push to main. This commented out block only triggers the workflow on release (below)  We could only trigger the workflow to only run if if we use the Github release feature
Screen Shot 2022-12-23 at 7 42 30 AM

.github/workflows/publish.yml
pfaraone: Ok - removing this commented out block then

.github/workflows/publish.yml
pfaraone: Changed to ```python -m pytest```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
+name: Publish Python distributions to PyPI and TestPyPI
+on: 
+  push: # Runs your workflow when you push a commit or tag to main or release branches.
+      branches:
+        - main
+        - 'releases/**'
+jobs:
+  build-n-publish:
+    name: Build and publish Python distributions to PyPI and TestPyPI
```
raghumdani, line 9: We are not publishing to TestPyPI now, hence we modify the name to reflect that.

raghumdani (approved):
Looks good

requirements.txt
```diff
@@ -1,9 +1,11 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
-s3fs >= 2021.7.0
-tenacity >= 8.0.1
-ray == 1.10.0
+s3fs >= 2022.11.0
+tenacity >= 8.1.0
+ray[default] >= 2.0.0
```
rkenmi, line 5: Just curious but any reason for setting *all* dependencies to `>= {version}`? I encountered some slow backtracking time with pip's new dependency resolver (see [here](https://stackoverflow.com/questions/65122957/resolving-new-pip-backtracking-runtime-issue)), which I think was partly due to us not tightening these version requirements.

pfaraone: These loose dependency ranges are an artifact of modifying our AWS Glue on Ray tests (uses ```Ray 2.0.0```)  if I recall correctly. In the latest commit, I've now pinned most of the dependencies to what ```pip list``` indicates is actually installed - reducing the ```Install Dependencies``` job runtime from 56 to 35 seconds in one run (https://github.com/pfaraone/deltacat/actions/runs/3836451986/jobs/6530636255).

rkenmi (approved):
LGTM

valiantljk (approved):
lgtm

Issue #28, opened 2023-01-03
Import issues with Ray Datasets


rkenmi:
@pfaraone 
> Import is broken in deltacat/storage/model/types.py with the ray version bump
Old ray: from ray.data.impl.arrow_block import ArrowRow
Curr ray :  from ray.data._internal.arrow_block import ArrowRow

PR #29, opened 2023-01-04
Update imports used for Ray Datasets / Ray 2.0+

rkenmi:
#28 

Tested via direct imports:

```
python

>>> import deltacat.io.aws.redshift.redshift_datasource
>>> import deltacat.io.dataset
>>> import deltacat.io.read_api
```

### DIFF
--- a/deltacat/io/dataset.py
+++ b/deltacat/io/dataset.py
@@ -4,14 +4,15 @@ from __future__ import annotations
 import pyarrow as pa
 import s3fs
 
-from typing import Optional, Union, Callable, Dict, Any, cast
+from typing import Optional, Union, Callable, Dict, Any, cast, TypeVar
 
 from ray.data import Dataset
 from ray.data.datasource import DefaultBlockWritePathProvider, \
     BlockWritePathProvider
-from ray.data._internal.arrow_block import T
 
 from deltacat.io.aws.redshift.redshift_datasource import RedshiftDatasource
+
+T = TypeVar('T')
 
 
 class DeltacatDataset(Dataset[T]):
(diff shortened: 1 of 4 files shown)


deltacat/io/read_api.py
```diff
@@ -3,7 +3,7 @@
 from deltacat.utils.common import ReadKwargsProvider
 
 from ray.data import read_datasource
-from ray.data.impl.arrow_block import ArrowRow
+from ray.data._internal.arrow_block import ArrowRow
```
raghumdani, line 6: Ideally, we should not rely on imports from ray _internal modules as they are subject to change without notice. Is ArrowRow not part of public module?

deltacat/io/dataset.py
```diff
@@ -9,7 +9,7 @@
 from ray.data import Dataset
 from ray.data.datasource import DefaultBlockWritePathProvider, \
     BlockWritePathProvider
-from ray.data.impl.arrow_block import T
+from ray.data._internal.arrow_block import T
```
raghumdani, line 12: You can just create a new TypeVar for generics instead of importing from ray _internal module.

raghumdani (commented):
Overall, looks good. Can you also add details about how these changes were tested?

deltacat/io/read_api.py
rkenmi: Agreed, but unfortunately `ArrowRow` isn't part of any ray public module and is only located in `_internal`

deltacat/io/dataset.py
rkenmi: Yeah, probably a good idea to reduce as many imports from `_internal` as we can

deltacat/io/read_api.py
raghumdani: Ok, maybe you want to add a TODO to clean up this once we have this type available in public module?

raghumdani (approved):

PR #31, opened 2023-01-05
Creating a release to trigger Github Actions workflow to build and publish to PyPI

pfaraone:
## Reason
In [26](https://github.com/ray-project/deltacat/pull/26) pushing a commit will trigger the workflow but pushing a tag (pointer to a specific commit) will not. As the downstream `Publish distribution to PyPI` job checks if the in current push event is for a tag and tag would not be present (job only triggered on a pushed commit) the distribution will not be published.

Expected behavior: Pushed tag will trigger workflow
Actual behavior: Pushed tag does not trigger workflow

A better trigger for the workflow to execute is to publish a release via the GitHub UI (also used in [ray core](https://github.com/ray-project/ray/releases)). When you publish a release on GitHub it tags the main branch with the version of the release that you specify. Each execution of the workflow is therefore guaranteed to be a tagged commit on the main branch.

Additionally the `setup.py` file dependencies are brought in sync with `requirements.txt`


rkenmi (commented):
Overall LGTM, could we possibly squash the commits? Seems like some pretty old commits are also being included (~2 weeks ago)

PR #34, opened 2023-01-06
Add pre-commit config file

pfaraone:
## Reason
Pre-commit lets you identify and fix some simple code (e.g., json validity) and styling issues (e.g., PEP8 compliance) before each commit. Depending on the check, a failure can stop your Git commit from happening, thus permitting you to fail fast and save unnecessary CI runner compute time. 

Install all developer dependencies
```
pip install -r dev-requirements.txt
```
### Setting up pre-commit
[`pre-commit`](https://pre-commit.com/) is a framework for managing and maintaining multi-language pre-commit hooks:
### Prerequisites
1. Install pre-commit
```
pip install pre-commit
```
or
```
conda install -c conda-forge pre-commit
```
or just install the dev-requirements file.
```
pip install -r dev-requirements.txt
```
2. Install all mentioned hooks/packages in the `.pre-commit-config.yaml` file
```
pre-commit install
```
#### Usage
Test files modified by your commit
```
git add deltacat/__init__.py
git commit -m "this commit will be verified"
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
black....................................................................Passed
pycln....................................................................Passed
isort....................................................................Passed
...
```
#### Testing all files
For file-based checks, the only files checked are the ones modified by your commit. To  force pre-commit to check all files in your repo once run
```
pre-commit run --all-files
```
#### Skipping hooks
If you need to bypass the pre-commit hooks (not recommended), you can use the -n or --no-verify flag when committing.
```
git commit -nm "This commit should succeed"
```

## Discussion
- Need to align formatting standards with [ray](https://github.com/ray-project/ray)


requirements.txt
```diff
@@ -1,11 +1,11 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
-s3fs == 2022.11.0
-tenacity == 8.1.0
-ray[default] == 2.0.0
+boto3 == 1.24.59
+numpy >= 1.21.1
```
raghumdani, line 4: can we pin these to working versions?

.flake8
```diff
@@ -0,0 +1,36 @@
+[flake8]
+exclude =
+max-line-length = 88
+inline-quotes = "
+ignore =
```
raghumdani, line 5: Any reason why we are ignoring these rules? Some context here would help.

.flake8
```diff
@@ -0,0 +1,36 @@
+[flake8]
+exclude =
+max-line-length = 88
```
raghumdani, line 3: In Intellij we generally use 120, do you have any reference library?

.flake8
```diff
@@ -0,0 +1,36 @@
+[flake8]
+exclude =
```
raghumdani, line 2: Is this line required?

dev-requirements.txt
```diff
@@ -0,0 +1 @@
+pre-commit == 2.20.0
```
raghumdani, line 1: don't we want to also add isort, flake and black?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+fail_fast: true
+minimum_pre_commit_version: '0'
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-added-large-files
+    -   id: check-case-conflict
+    -   id: check-json
+    # -   id: check-yaml # skip for now
```
raghumdani, line 12: Any reason? I am not a fan of commented code, as it does not serve any purpose. It can be replace via human readable documentation though.

raghumdani (commented):
Looks mostly good. I have left some comments:

1. I saw you have added `dev-requirements.txt`. Do you want to add the README.md about installing dev dependencies and setup of pre-commit?
2. How did we make sure pre-commit-hooks like `trailing-whitespace` etc. follow same convention as black code formatter?

.flake8
rkenmi: I think this is partially inspired from [ray's flake8 file](https://github.com/ray-project/ray/blob/master/.flake8)?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,29 @@
...
+    -   id: end-of-file-fixer
+    -   id: requirements-txt-fixer
+    -   id: trailing-whitespace
```
rkenmi, line 15: I'm curious if some of these hooks overlap with the functionality of flake8/black, for example flake8 looks like it already has a [rule](https://www.flake8rules.com/rules/W291.html) for trailing whitespaces.

.flake8
pfaraone: Yes it is copied verbatim from Ray's flake8 file

.flake8
pfaraone: Copied verbatim from (Ray)[https://github.com/ray-project/ray/blob/master/.flake8]. Should have made that clear

dev-requirements.txt
pfaraone: I've added these dependencies as well as `pytest` and `pytest-cov`

pfaraone:
> Looks mostly good. I have left some comments:
> 
> 1. I saw you have added `dev-requirements.txt`. Do you want to add the README.md about installing dev dependencies and setup of pre-commit?
> 2. How did we make sure pre-commit-hooks like `trailing-whitespace` etc. follow same convention as black code formatter?

1. Will do 
2. Black is run after the out of the box pre-commit (hooks)[https://github.com/pre-commit/pre-commit-hooks]. Therefore black will overwrite any trailing-whitespace fixed implemented by the earlier hooks.

.isort.cfg
```diff
@@ -0,0 +1,10 @@
+[settings]
+# This is to make isort compatible with Black. See
+# https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines.
```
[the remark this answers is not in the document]
pfaraone: I'll review this. It's copied verbatim from (Ray)[https://github.com/ray-project/ray/blob/master/.isort.cfg]

.flake8
pfaraone: Removed

requirements.txt
pfaraone: Pinned: `numpy == 1.23.4`

.github/workflows/publish-to-pypi.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       run: >-
         python -m pytest
     - name: Echo release tag
-      run: echo ${{ github.ref_name }} >> $RELEASE_TAG
```
pfaraone, line 33: Removed redirect as getting an ambiguous redirect error

.flake8
```diff
@@ -0,0 +1,35 @@
+[flake8]
```
raghumdani, line 1: You may want to exclude `build` and `.eggs`.

README.md
```diff
@@ -2,10 +2,62 @@
...
+Its data storage model allows you to define and manage fast, scalable,
+ACID-compliant data catalogs through git-like stage/commit APIs, and has been
 used to successfully host exabyte-scale enterprise data lakes.
 
 DeltaCAT uses the Ray distributed compute framework together with Apache Arrow
-for common table management tasks, including petabyte-scale 
+for common table management tasks, including petabyte-scale
 change-data-capture, data consistency checks, and table repair.
+
+## Getting Started
```
valiantljk, line 13: a readme file for a public lib should aim for non-developers; development section can be marked as How to contribute, etc. take a look of other oss projects to get some template.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,31 @@
...
+    rev: v4.4.0
+    hooks:
+    -   id: check-added-large-files
+    -   id: check-case-conflict
+    -   id: check-json
+    -   id: check-yaml
+    -   id: end-of-file-fixer
+    -   id: requirements-txt-fixer
+    -   id: trailing-whitespace
+    -   id: no-commit-to-branch
```
valiantljk, line 14: is there a way to set the `protected` branch in github?

valiantljk (commented):
lgtm

raghumdani (approved):
LGTM

.pre-commit-config.yaml
pfaraone: Yes there is. You can create a branch protection rule (brazil ws use --package SunGatePython). In that case this hook may be redundant

README.md
pfaraone: Updated the README

.flake8
pdames: +1 to maintaining a consistent style with Ray. I also prefer shorter lines (e.g. 70-90 characters) since it makes it easier to develop/review multiple files side-by-side, is helpful to anyone w/o great eyesight who reads the code (since they can enlarge the text while still fitting whole lines on their monitor), and encourages writing more diff-friendly code that is more (e.g. by forcing you to not pack too many arguments to a function on one line of code).

setup.py
```diff
@@ -13,9 +13,8 @@ def find_version(*paths):
     version_file_path = os.path.join(ROOT_DIR, *paths)
     with open(version_file_path) as file_stream:
         version_match = re.search(
-            r"^__version__ = ['\"]([^'\"]*)['\"]",
-            file_stream.read(),
-            re.M)
+            r"^__version__ = ['\"]([^'\"]*)['\"]", file_stream.read(), re.M
```
pdames, line 16: Minor: Going back to the discussion on diff-friendly code, why consolidate all 3 args on one-line instead of one arg per line? Did the current linter config do this automatically (in which case, assuming it's 100% consistent with Ray, maybe we could recommend a change to the linter config for the main Ray repo)?

pdames (approved):

pdames:
As a high-level comment, it would be good to submit another PR after this that includes all changes that result from applying this linter to the entire project, so that we don't continue gradually eating a bunch of unrelated linter changes in every subsequent PR.

setup.py
pdames: Hmmm... was it because of the last line ending in `re.M)`? I wonder if it would have done the same if it was instead split into two lines with a trailing comma like:
```
    re.M,
)
```

Either way, I guess this is one case where I don't feel like the current config is necessarily improving code, but that could just be me.

PR #40, opened 2023-01-13
Bumped dependencies down to support Python 3.7

pfaraone:
## Why 
In #26  we pinned some dependency versions which pinned us to py38+ even though we've explicitly stated we are supporting py37 and above

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,11 +1,11 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
-boto3 == 1.24.59
-numpy == 1.23.4
-pandas == 1.5.2
+boto3 == 1.20.24
+numpy == 1.21.5
+pandas == 1.3.5
 pyarrow == 10.0.1
 pydantic == 1.10.4
 ray[default] == 2.0.0
-s3fs == 2022.11.0
+s3fs == 2022.2.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,11 +1,11 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
-boto3 == 1.24.59
-numpy == 1.23.4
-pandas == 1.5.2
+boto3 == 1.20.24
```
raghumdani, line 3: Glue uses Python 3.9, did we test this on Glue?

setup.py
```diff
@@ -25,24 +25,24 @@ def find_version(*paths):
 
 
 setuptools.setup(
-    name="deltacat",
+    name="deltacat-fork",
```
raghumdani, line 28: Why renaming to deltacat-fork?

pfaraone: Sorry this carried over from my fork version testing - I'll change it

pfaraone: Good catch

requirements.txt
pdames: I haven't encountered any issues with these dependencies on Glue.

pdames (approved):
LGTM - thanks!

PR #41, opened 2023-01-17
[compaction] Enable compaction based on compacted source partition; Add support for primary key index building

Zyiqin-Miranda:
This PR adds support for compaction based on compacted source partition locator by:
1. Building initial primary key indices on compacted source partition locator.
2. Producing round completion file containing all initial primary key indices and stats about primary key indices.
3. In subsequent compaction run, reading high watermark from round completion file and discover new deltas.
4. In subsequent compaction run, consuming all primary key indices during dedupe correctly.

Testing:
For correctness check, tested with the following scenario:
delta 1: 100 rows, building primary key indices on.
delta 2: 40 rows, 20 distinct rows. 40 rows having primary keys same as in delta 1.
delta 3: 200 rows, 200 distinct rows. 100 rows having primary keys same as in delta 1.

Scenario 1: Build primary key indices on delta 1 first, subsequent compaction on delta 2. Compaction dedupe step taking 140 rows as input, producing 100 rows as output.
Scenario 2: Build primary key indices on delta 1 first, subsequent compaction on delta 2 and delta 3. Compaction dedupe step taking 340 rows as input, producing 200 rows as output.

### DIFF
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -56,12 +56,47 @@ def check_preconditions(
     )
 
 
+def check_build_initial_primary_key_indices_preconditions(
+            build_initial_primary_key_index: bool,
+            source_partition_locator: PartitionLocator,
+            source_compacted_partition_locator: PartitionLocator,
+            deltacat_storage):
+    if build_initial_primary_key_index:
+        assert source_compacted_partition_locator is not None
+    assert source_partition_locator.partition_values \
+        == source_compacted_partition_locator.partition_values, \
+        "Building initial primary key indices must use the same partition values " \
+        "for the source and compacted partition locator."
+    stream_locator = source_compacted_partition_locator.stream_locator
+    namespace = stream_locator.namespace
+    table_name = stream_locator.table_name
+    table_version = stream_locator.table_version
+    partition_values = source_compacted_partition_locator.partition_values
+    deltas_list_result = deltacat_storage.list_deltas(
+        namespace,
+        table_name,
+        partition_values,
+        table_version,
+        None,
+        None,
+        False,
+    )
+    delta_res = deltas_list_result.all_items()
+    assert len(delta_res) == 1, \
+        "Only one delta is expected when building initial primary key indices from compacted partition locator"
+    return delta_res[0].stream_position
+
+
 def compact_partition(
         source_partition_locator: PartitionLocator,
         compacted_partition_locator: PartitionLocator,
         primary_keys: Set[str],
         compaction_artifact_s3_bucket: str,
         last_stream_position_to_compact: int,
+        source_compacted_partition_locator: Optional[PartitionLocator] = None,
+        build_initial_primary_key_indices: Optional[bool] = False,
+        compaction_based_on_initial_primary_key_indices: Optional[bool] = False,
+        primary_key_source_stream_position: Optional[int] = None,
         hash_bucket_count: Optional[int] = None,
         sort_keys: List[SortKey] = None,
         records_per_primary_key_index_file: int = 38_000_000,
@@ -71,7 +106,7 @@ def compact_partition(
         min_hash_bucket_chunk_size: int = 0,
         compacted_file_content_type: ContentType = ContentType.PARQUET,
         delete_prev_primary_key_index: bool = False,
-        read_round_completion: bool = False,
+        read_round_completion: bool = True,
         pg_config: Optional[List[Dict[str, Any]]] = None,
         schema_on_read: Optional[pa.schema] = None,  # TODO (ricmiyam): Remove this and retrieve schema from storage API
         deltacat_storage=unimplemented_deltacat_storage):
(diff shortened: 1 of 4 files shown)


deltacat/compute/compactor/compaction_session.py
```diff
@@ -71,7 +106,7 @@ def compact_partition(
         min_hash_bucket_chunk_size: int = 0,
         compacted_file_content_type: ContentType = ContentType.PARQUET,
         delete_prev_primary_key_index: bool = False,
-        read_round_completion: bool = False,
+        read_round_completion: bool = True,
```
raghumdani, line 109: changing defaults is often backward incompatible. I'd recommend document motivation for the defaults as a doc.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -56,12 +56,47 @@ def check_preconditions(
...
 def compact_partition(
         source_partition_locator: PartitionLocator,
         compacted_partition_locator: PartitionLocator,
         primary_keys: Set[str],
         compaction_artifact_s3_bucket: str,
         last_stream_position_to_compact: int,
+        source_compacted_partition_locator: Optional[PartitionLocator] = None,
+        build_initial_primary_key_indices: Optional[bool] = False,
+        compaction_based_on_initial_primary_key_indices: Optional[bool] = False,
+        primary_key_source_stream_position: Optional[int] = None,
```
raghumdani, line 99: Throughout this pkg, I often had to look into implementation to understand the purpose of a parameter. A documentation could ease readers in understanding at-least for this important method this is exposed.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -56,12 +56,47 @@ def check_preconditions(
     )
 
 
+def check_build_initial_primary_key_indices_preconditions(
```
raghumdani, line 59: I see this method is retrieving compacted snapshot from a different partition locator. Maybe we can rename accordingly.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -56,12 +56,47 @@ def check_preconditions(
...
+    deltas_list_result = deltacat_storage.list_deltas(
+        namespace,
+        table_name,
+        partition_values,
+        table_version,
+        None,
+        None,
+        False,
+    )
+    delta_res = deltas_list_result.all_items()
```
raghumdani, line 84: This method call retrieves all the deltas which can be time-consuming. If you just want to pick one, why not only retrieve the first page?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -56,12 +56,47 @@ def check_preconditions(
...
+    partition_values = source_compacted_partition_locator.partition_values
+    deltas_list_result = deltacat_storage.list_deltas(
+        namespace,
+        table_name,
+        partition_values,
+        table_version,
+        None,
+        None,
+        False,
+    )
```
raghumdani, line 83: I generally prefer named args as they are easy to read and backward compatible when you add a new param to the method. Any thoughts?

deltacat/compute/compactor/compaction_session.py
Zyiqin-Miranda: Agree with more document about default values. 
For `read_round_completion`, should be set to `True` because we need to enable multi-round compaction to limit input deltas when the partition is large. I think this flag set to `False` in `scaling-test` branch initially cause we don't want to read from previous test result which should be set back when merging back to `main`.

deltacat/compute/compactor/compaction_session.py
Zyiqin-Miranda: Thanks Raghavendra, will add explanation on these added arguments.

deltacat/compute/compactor/compaction_session.py
Zyiqin-Miranda: This is a precondition check before building primary key indices that the partition locator passed in should only contain one delta. So all_items() should contain only one delta.

deltacat/compute/compactor/compaction_session.py
Zyiqin-Miranda: Agreed, will change to named args

deltacat/compute/compactor/compaction_session.py
Zyiqin-Miranda: This method is checking the preconditions before building the primary key indices, like the partition value we build primary key indices on should be the same as the partition value of destination partition we written to. [Here is](https://github.com/ray-project/deltacat/blob/76b49d818e310412347e971a1f470a1d4fc38822/deltacat/compute/compactor/compaction_session.py#L35) the precondition check method for actual compaction.

Zyiqin-Miranda: Would you mind elaborate more on what should be the method name?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -83,54 +118,114 @@ def compact_partition(
...
+            input_deltas_stats=input_deltas_stats,
+            min_pk_index_pa_bytes=0,
+            min_hash_bucket_chunk_size=0,
+            compacted_file_content_type=ContentType.PARQUET,
+            schema_on_read=schema_on_read,
+            delete_prev_primary_key_index=False,
+            read_round_completion=True,
+            deltacat_storage=deltacat_storage)
+    else:
+        while has_next_compaction_round:
```
valiantljk, line 172: It looks like the code can be simplified into:
```
def compact_partition():
      while has_next_compaction_round:
           has_next_compaction_round_obj, new_partition_obj, new_rci_obj = \
                _execute_compaction_round.options(**opts).remote()
                build_initial_primary_key_indices = False
```

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -64,13 +65,15 @@ def union_primary_key_indices(
...
             prior_pk_index_table = pa.concat_tables(tables)
-            prior_pk_index_table = sc.append_stream_position_column(
-                prior_pk_index_table,
-                repeat(
-                    prev_compacted_delta_stream_pos,
-                    len(prior_pk_index_table),
-                ),
-            )
+            if not sc.if_stream_position_column_exists(prior_pk_index_table):
+                logger.info(f"peach: append stream position column when pk: {build_initial_primary_key_indices}")
```
valiantljk, line 69: might need better logger info to make it more readable?

Zyiqin-Miranda: Nice catch! Sorry I left the debug lines here

deltacat/compute/compactor/compaction_session.py
rkenmi: It sounds like this is about avoiding some risk with latency - if we got a bad table that didn't have a single delta, and somehow happened to have more than 1000+ deltas, then it would take an expensive network call before we actually decide to fail the validation check.

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -198,14 +202,22 @@ def write_new_primary_key_index(
...
+                sc._PARTITION_STREAM_POSITION_COLUMN_NAME,
+                sc._ORDERED_FILE_IDX_COLUMN_NAME,
+                sc._ORDERED_RECORD_IDX_COLUMN_NAME,
+            ])
+            table = sc.append_file_idx_column(table, dest_file_idx_col)
+            table = sc.append_record_idx_col(table, dest_file_row_idx_col)
+        else:
+            # Force primary key file idx and record idx point to source compacted partition
+            # when building initial primary key indices
+            table = table.drop([
```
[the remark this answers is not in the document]
Zyiqin-Miranda: Build PK indices flow needs one more column which is the delta stream position associated with source compacted partition. In normal flow, this column is appended as part of reading primary key indices, which is the delta stream position associated with our compacted partition. So it needs some kind of additional info here to make subsequent compaction works correctly.

deltacat/compute/compactor/utils/io.py
```diff
@@ -207,3 +226,77 @@ def limit_input_deltas(
     logger.info(f"Input uniform delta count: {len(rebatched_da_list)}")
 
     return rebatched_da_list, hash_bucket_count, latest_stream_position
+
+
+def annotate_delta_for_primary_key_indices_building(
+        input_deltas: List[Delta],
```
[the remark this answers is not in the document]
Zyiqin-Miranda: Only one delta is expected, otherwise we should raise an exception.

PR #46, opened 2023-01-28
 Ray’s shared memory object store not garbage collecting after job completes

pfaraone:
> When a worker calls `ray.put()` or returns from a task, it copies the provided values into Ray’s shared memory object store. Ray will then [make these objects available](https://docs.google.com/document/d/1tBw9A4j62ruI5omIJbMxly-la5w4q_TjyJgJL_jN2fI/preview#heading=h.rtepjltxoeb7) throughout the cluster, attempt to [recover](https://docs.google.com/document/d/1tBw9A4j62ruI5omIJbMxly-la5w4q_TjyJgJL_jN2fI/preview#heading=h.tlm9n2vnrd30) them in case of a failure, [spill](https://docs.google.com/document/d/1tBw9A4j62ruI5omIJbMxly-la5w4q_TjyJgJL_jN2fI/preview#heading=h.7ce0zjt0shwh) them if the object store exceeds its configured capacity, and [garbage-collect](https://docs.google.com/document/d/1tBw9A4j62ruI5omIJbMxly-la5w4q_TjyJgJL_jN2fI/preview#heading=h.pyvgsp25rn5q) them once all ObjectRefs have gone out of scope.

https://docs.ray.io/en/latest/ray-core/scheduling/memory-management.html#objectref-reference-counting

Ray’s object store ([Plasma Object Store](https://docs.ray.io/en/latest/ray-core/objects/serialization.html#plasma-store)) only contains data stored via ray.put(), task args and return values. Ray implements distributed reference counting so that any ObjectID in scope in the cluster is pinned in the object store. This includes local python references, arguments to pending tasks, and IDs serialized inside of other objects. However, even after the job completes, the plasma store usage remains elevated as indicated when running `ray memory` in the head node.

In this PR, these changes result in the object emptying after the job completes: 
1. Manually deleting the cloudpickled object references that are pinned in memory. Calling `del` on a Python object decreases the local reference count by 1.

These changes seem to fix Ray distributed reference counting of the object references (allowing it to hit 0) and thereby allowing the object to be garbage collection. 

Additionally, imports in files touched were sorted using `isort`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,20 +10,20 @@ repos:
     -   id: check-yaml
     -   id: end-of-file-fixer
     -   id: requirements-txt-fixer
-   #  -   id: trailing-whitespace
+    -   id: trailing-whitespace
 -   repo: https://github.com/hadialqattan/pycln
     rev: v2.1.2
     hooks:
     -   id: pycln
-# -   repo: https://github.com/PyCQA/isort
-#     rev: 5.11.4
-#     hooks:
-#     -   id: isort
-# -   repo: https://github.com/psf/black
-#     rev: 22.12.0
-#     hooks:
-#       - id: black
-# -   repo: https://github.com/pycqa/flake8
-#     rev: 6.0.0
-#     hooks:
-#     -   id: flake8
+-   repo: https://github.com/PyCQA/isort
+    rev: 5.11.4
+    hooks:
+    -   id: isort
+-   repo: https://github.com/psf/black
+    rev: 22.12.0
+    hooks:
+      - id: black
+-   repo: https://github.com/pycqa/flake8
+    rev: 6.0.0
+    hooks:
+    -   id: flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,20 +10,20 @@ repos:
     -   id: check-yaml
     -   id: end-of-file-fixer
     -   id: requirements-txt-fixer
-    -   id: trailing-whitespace
+   #  -   id: trailing-whitespace
 -   repo: https://github.com/hadialqattan/pycln
     rev: v2.1.2
     hooks:
     -   id: pycln
--   repo: https://github.com/PyCQA/isort
-    rev: 5.11.4
-    hooks:
-    -   id: isort
--   repo: https://github.com/psf/black
-    rev: 22.12.0
-    hooks:
-      - id: black
--   repo: https://github.com/pycqa/flake8
-    rev: 6.0.0
-    hooks:
-    -   id: flake8
+# -   repo: https://github.com/PyCQA/isort
+#     rev: 5.11.4
+#     hooks:
+#     -   id: isort
+# -   repo: https://github.com/psf/black
+#     rev: 22.12.0
+#     hooks:
+#       - id: black
+# -   repo: https://github.com/pycqa/flake8
+#     rev: 6.0.0
+#     hooks:
+#     -   id: flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,8 +11,6 @@ repos:
     -   id: end-of-file-fixer
     -   id: requirements-txt-fixer
     -   id: trailing-whitespace
-    -   id: no-commit-to-branch
-        args: [--branch, main]
 -   repo: https://github.com/hadialqattan/pycln
     rev: v2.1.2
     hooks:
--- a/deltacat/compute/compactor/steps/dedupe.py
+++ b/deltacat/compute/compactor/steps/dedupe.py
@@ -364,7 +364,7 @@ def dedupe(
             dedupe_task_index,
             object_ref,
         )
-        del ray.put
+        del object_ref
     logger.info(f"Count of materialize buckets with object refs: "
                 f"{len(mat_bucket_to_dd_idx_obj_id)}")
 
--- a/deltacat/compute/compactor/steps/dedupe.py
+++ b/deltacat/compute/compactor/steps/dedupe.py
@@ -364,7 +364,6 @@ def dedupe(
             dedupe_task_index,
             object_ref,
         )
-        del object_ref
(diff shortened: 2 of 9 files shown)


deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -363,8 +362,9 @@ def dedupe(
         src_file_records_obj_refs.append(object_ref)
         mat_bucket_to_dd_idx_obj_id[mat_bucket] = (
             dedupe_task_index,
-            cloudpickle.dumps(object_ref),
+            object_ref,
         )
+        del ray.put
```
raghumdani, line 367: del ray.put?

deltacat/compute/compactor/utils/primary_key_index.py
```diff
@@ -209,7 +208,8 @@ def group_hash_bucket_indices(
         if obj is not None:
             obj_ref = ray.put(obj)
             object_refs.append(obj_ref)
-            hash_bucket_group_to_obj_id[hb_group] = cloudpickle.dumps(obj_ref)
+            hash_bucket_group_to_obj_id[hb_group] = obj_ref
+            del obj_ref
```
raghumdani, line 212: Though this approach is cleaning up object store, its not cleaning up main memory.

deltacat/compute/compactor/steps/dedupe.py
pfaraone: Type - fixed now

deltacat/compute/compactor/utils/primary_key_index.py
pfaraone: I've experimented with the free internal API but it's not having any impact on cleaning up main memory.

valiantljk:
just reviewed the changes. overall, looks good. One question, are there any other places in the code that use cloudpickle? should we change all?

pfaraone:
> just reviewed the changes. overall, looks good. One question, are there any other places in the code that use cloudpickle? should we change all?

In this PR, all `cloudpickle` API calls have been removed.

deltacat/compute/compactor/utils/primary_key_index.py
pfaraone: Running  `ray globalgc` has no impact on the main memory usage as well.

pdames (commented):
Looks OK at a high-level, but a couple overall comments/questions:

1. Have we run benchmarks with the new code to assess its performance, memory consumption, runtime latency, stability etc. vs. the old? What were the results?

2. I feel like, if we really want to ensure a clean state at the beginning of the next round, our efforts would be better spent in launching a new cluster per round, and then focusing more on removing unnecessary memory consumed between each high-level parallel step (hash bucket, dedupe, materialize). Even if we make some incremental improvements to object store memory consumption between rounds, we're still left with a dirty cluster at the start of the next round. The persistence of anything written to disk, metadata written to the GCS, or any other untracked artifacts on any of our cluster nodes seem to make per-round cluster cleanup a moving target.

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -363,8 +362,9 @@ def dedupe(
         src_file_records_obj_refs.append(object_ref)
         mat_bucket_to_dd_idx_obj_id[mat_bucket] = (
             dedupe_task_index,
-            cloudpickle.dumps(object_ref),
+            object_ref,
         )
+        del object_ref
```
pdames, line 367: This and our other manual object ref deletes will look confusing to future maintainers who don't have all the context that preceded this PR. Can you include a comment stating why we're including the delete here, what it is meant to help with, and/or a link to an existing discussion about the same on GitHub or the Ray Discussion forums?

deltacat/compute/compactor/utils/primary_key_index.py
raghumdani: Makes sense. Let's revisit if it becomes a blocker.

deltacat/compute/compactor/steps/dedupe.py
pfaraone: Resolving this conversation as I'm no longer using the `del` for manual object deletion in my latest commit. Solely removing the `cloudpickle` API calls allows the object memory to free up properly after the job run finishes.

.pre-commit-config.yaml
```diff
@@ -11,8 +11,6 @@ repos:
     -   id: end-of-file-fixer
     -   id: requirements-txt-fixer
     -   id: trailing-whitespace
-    -   id: no-commit-to-branch
```
pfaraone, line 14: There are alternative ways to protect specific branches from direct checkins using only Github settings.

raghumdani (approved):
LGTM

deltacat/compute/compactor/utils/primary_key_index.py
```diff
@@ -206,11 +206,13 @@ def group_hash_bucket_indices(
...
-            object_refs.append(obj_ref)
-            hash_bucket_group_to_obj_id[hb_group] = cloudpickle.dumps(obj_ref)
-
+        if obj is None:
+            continue
+        obj_ref = ray.put(obj)
+        pickled_obj_ref = cloudpickle.dumps(obj_ref)
+        object_refs.append(pickled_obj_ref)
+        hash_bucket_group_to_obj_id[hb_group] = pickled_obj_ref
+        del obj_ref
```
raghumdani, line 215: Let's add documentation around why we are deleting this object reference.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -90,9 +92,9 @@ def _materialize(
     logger.info(f"Starting materialize task...")
     dedupe_task_idx_and_obj_ref_tuples = [
         (
-            t[0],
-            cloudpickle.loads(t[1])
-        ) for t in dedupe_task_idx_and_obj_id_tuples
```
pfaraone, line 95: Refactor only

rkenmi (approved):

Issue #37, opened 2023-01-11
Retroactively Apply New Linter Config to Entire Project


pdames:
Issue for unaddressed TODO on https://github.com/ray-project/deltacat/pull/34:

> As a high-level comment, it would be good to submit another PR after this that includes all changes that result from applying this linter to the entire project, so that we don't continue gradually eating a bunch of unrelated linter changes in every subsequent PR.

pdames:
To help prevent linter backfills in the future, this should ideally be preceded by: https://github.com/ray-project/deltacat/issues/60

PR #47, opened 2023-01-28
#37- Retroactively applying linter (black, isort) + Lint on CI

pfaraone:
# Issue
#37


.flake8
```diff
@@ -35,4 +35,6 @@ ignore =
   B015
   B016
   B017
+  F541
```
pfaraone, line 38: ignore 
- f-string without placeholder

pdames (approved):
LGTM!

valiantljk (approved):
LGTM, thanks.

raghumdani (commented):
I assume there is no manual changes after or before applying linter, however `README about how to run linter and integrate it with vscode or pycharm` would be helpful for other maintainers.

raghumdani (approved):

pdames:
To ensure that styling remains consistent across subsequent PRs, please provide a simple script that can be run to setup Git commit hooks. For example:

https://github.com/ray-project/ray/blob/master/setup_hooks.sh

We should also ensure that our CI workflow runs a linter check on all PRs. For example, here's a similar check that ensures that Apache license header is present in all new files added to a PR:

https://github.com/ray-project/ray_beam_runner/blob/master/.github/workflows/ci.yml#L74-L85

pfaraone:
> To ensure that styling remains consistent across subsequent PRs, please provide a simple script that can be run to setup Git commit hooks. For example:
> 
> https://github.com/ray-project/ray/blob/master/setup_hooks.sh
> 
> We should also ensure that our CI workflow runs a linter check on all PRs. For example, here's a similar check that ensures that Apache license header is present in all new files added to a PR:
> 
> https://github.com/ray-project/ray_beam_runner/blob/master/.github/workflows/ci.yml#L74-L85

CI workflow now runs a linter check on all PRs against main.

> To ensure that styling remains consistent across subsequent PRs, please provide a simple script that can be run to setup Git commit hooks. For example:

I don't think this is necessary at the moment as pre-commit comes with all the batteries included to set itself and the linters up  (just run `pre-commit install` - https://pre-commit.com/#3-install-the-git-hook-scripts) . I could add to a README-dev.md that you have to `pip install dev-requirements.txt` then `pre-commit install` to run pre-commit locally

setup.py
```diff
@@ -25,7 +25,7 @@ def find_version(*paths):
 
 
 setuptools.setup(
-    name="deltacat",
+    name="deltacat-fork",
```
pdames, line 28:
```suggestion
    name="deltacat",
```

pdames (approved):
LGTM! I applied a quick fix to revert the test package name left in `setup.py`

PR #52, opened 2023-01-31
Eagerly materialize PyArrow tables up to records_per_compacted_file

rkenmi:
Tested on tables with delta entry files > 4000000 records.

Issue: https://github.com/ray-project/deltacat/issues/53

With `records_per_compacted_file = 4_000_000` (default):
```
Max in-memory byte size of any concatenated table groups: 8_188_431_980
Max record count of any concatenated table groups: 3_999_154
```

With `records_per_compacted_file = 20_000_000`:
```
Max in-memory byte size of any concatenated table groups: 13_510_287_893
Max record count of any concatenated table groups: 6_834_441
```

### DIFF
--- a/deltacat/compute/compactor/steps/materialize.py
+++ b/deltacat/compute/compactor/steps/materialize.py
@@ -51,14 +51,13 @@ def materialize(
                      f"and record count: {compacted_tables_record_count}")
         compacted_table = pa.concat_tables(compacted_tables)
         if compacted_file_content_type in DELIMITED_TEXT_CONTENT_TYPES:
-            # convert to pandas since pyarrow doesn't support custom delimiters
-            # and doesn't support utf-8 conversion of all types (e.g. Decimal128)
+            # TODO (ricmiyam): Investigate if we still need to convert this table to pandas DataFrame
             # TODO (pdames): compare performance to pandas-native materialize path
             df = compacted_table.to_pandas(
                 split_blocks=True,
                 self_destruct=True,
+                zero_copy_only=True
             )
-            del compacted_table
             compacted_table = df
         delta = deltacat_storage.stage_delta(
             compacted_table,
@@ -85,7 +84,6 @@ def materialize(
                 len(compacted_table),
             ),
         )
-        del compacted_table
         logger.info(f"Materialize result: {materialize_result}")
         return materialize_result
 
@@ -201,10 +199,10 @@ def materialize(
         f"Expected at least one materialized result in materialize step."
 
     # Merge all new deltas into one for this materialize bucket index
-    merged_mr = MaterializeResult.of(merged_delta,
-                                     materialized_results[0].task_index,
-                                     PyArrowWriteResult.union([mr.pyarrow_write_result
-                                                               for mr in materialized_results]))
+    merged_materialize_result = MaterializeResult.of(merged_delta,
+                                                     materialized_results[0].task_index,
+                                                     PyArrowWriteResult.union([mr.pyarrow_write_result
+                                                                               for mr in materialized_results]))
 
     logger.info(f"Finished materialize task...")
-    return merged_mr
+    return merged_materialize_result
--- a/deltacat/compute/compactor/steps/materialize.py
+++ b/deltacat/compute/compactor/steps/materialize.py
@@ -42,11 +42,13 @@ def materialize(
         deltacat_storage=unimplemented_deltacat_storage) -> MaterializeResult:
 
     def _materialize(
-            compacted_tables: List[pa.Table]) -> MaterializeResult:
+            compacted_tables: List[pa.Table],
+            compacted_tables_record_count: int) -> MaterializeResult:
         compacted_tables_size = sum([TABLE_CLASS_TO_SIZE_FUNC[type(tbl)](tbl)
                                      for tbl in compacted_tables])
         logger.debug(f"Uploading {len(compacted_tables)} compacted tables "
-                     f"with size: {compacted_tables_size} bytes")
+                     f"with size: {compacted_tables_size} bytes "
+                     f"and record count: {compacted_tables_record_count}")
         compacted_table = pa.concat_tables(compacted_tables)
         if compacted_file_content_type in DELIMITED_TEXT_CONTENT_TYPES:
             # convert to pandas since pyarrow doesn't support custom delimiters
@@ -182,7 +184,7 @@ def materialize(
         # Write manifests up to max_records_per_output_file
         if compacted_tables and \
                 total_record_count + record_count > max_records_per_output_file:
-            materialized_results.append(_materialize(compacted_tables))
+            materialized_results.append(_materialize(compacted_tables, total_record_count))
             # Free up written tables in memory
             compacted_tables.clear()
             total_record_count = 0
@@ -190,7 +192,7 @@ def materialize(
         total_record_count += record_count
         compacted_tables.append(pa_table)
 
-    materialized_results.append(_materialize(compacted_tables))
+    materialized_results.append(_materialize(compacted_tables, total_record_count))
     # Free up written tables in memory
     compacted_tables.clear()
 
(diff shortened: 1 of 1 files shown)


deltacat/compute/compactor/steps/materialize.py
```diff
@@ -101,69 +151,60 @@ def materialize(
...
+    materialized_results.append(_materialize(compacted_tables, total_record_count))
+    # Free up written tables in memory
+    compacted_tables.clear()
+
+    merged_delta = Delta.merge_deltas([mr.delta for mr in materialized_results])
+    assert materialized_results and len(materialized_results) > 0, \
+        f"Expected at least one materialized result in materialize step."
+
+    # Merge all new deltas into one for this materialize bucket index
+    merged_mr = MaterializeResult.of(merged_delta,
```
raghumdani, line 204: minor: rename `merged_materialize_result`

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -101,69 +151,60 @@ def materialize(
...
+            total_record_count = 0
+
+        total_record_count += record_count
+        compacted_tables.append(pa_table)
+
+    materialized_results.append(_materialize(compacted_tables, total_record_count))
+    # Free up written tables in memory
+    compacted_tables.clear()
+
+    merged_delta = Delta.merge_deltas([mr.delta for mr in materialized_results])
```
raghumdani, line 199: Minor: rename mr -> `materialized_result`

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -41,6 +41,54 @@ def materialize(
...
+            # TODO (pdames): Generalize WriteResult to contain in-memory-table-type
+            #  and in-memory-table-bytes instead of tight coupling to paBytes
+            PyArrowWriteResult.of(
+                len(manifest.entries),
+                TABLE_CLASS_TO_SIZE_FUNC[type(compacted_table)](compacted_table),
+                manifest.meta.content_length,
+                len(compacted_table),
+            ),
+        )
+        del compacted_table
```
raghumdani, line 88: I assume del would not be needed here as this goes out of scope anyway.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -41,6 +41,54 @@ def materialize(
...
+            compacted_tables_record_count: int) -> MaterializeResult:
+        compacted_tables_size = sum([TABLE_CLASS_TO_SIZE_FUNC[type(tbl)](tbl)
+                                     for tbl in compacted_tables])
+        logger.debug(f"Uploading {len(compacted_tables)} compacted tables "
+                     f"with size: {compacted_tables_size} bytes "
+                     f"and record count: {compacted_tables_record_count}")
+        compacted_table = pa.concat_tables(compacted_tables)
+        if compacted_file_content_type in DELIMITED_TEXT_CONTENT_TYPES:
+            # convert to pandas since pyarrow doesn't support custom delimiters
+            # and doesn't support utf-8 conversion of all types (e.g. Decimal128)
```
raghumdani, line 55: latest version of pyarrow now supports utf-8 conversion of decimal.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -41,6 +41,54 @@ def materialize(
         schema: Optional[pa.Schema] = None,
         deltacat_storage=unimplemented_deltacat_storage) -> MaterializeResult:
 
+    def _materialize(
```
raghumdani, line 44: why not define this in global namespace?

rkenmi: Mainly for readability and convenience. This function is specific to the materialize step, and moving this out to the global namespace means more variables to pass around.

deltacat/compute/compactor/steps/materialize.py
rkenmi: Maybe a fresh TODO here would be a good idea, since changing this would be a bit out-of-scope.

deltacat/compute/compactor/steps/materialize.py
rkenmi: Yeah, the intention here was to make it more explicit that we want to to free this table from memory.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -41,6 +41,54 @@ def materialize(
...
+                                     for tbl in compacted_tables])
+        logger.debug(f"Uploading {len(compacted_tables)} compacted tables "
+                     f"with size: {compacted_tables_size} bytes "
+                     f"and record count: {compacted_tables_record_count}")
+        compacted_table = pa.concat_tables(compacted_tables)
+        if compacted_file_content_type in DELIMITED_TEXT_CONTENT_TYPES:
+            # convert to pandas since pyarrow doesn't support custom delimiters
+            # and doesn't support utf-8 conversion of all types (e.g. Decimal128)
+            # TODO (pdames): compare performance to pandas-native materialize path
+            df = compacted_table.to_pandas(
```
raghumdani, line 57: `zero_copy_only` can be set to this method to avoid copying it altogether to pandas.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -101,69 +151,60 @@ def materialize(
...
-            TABLE_CLASS_TO_SIZE_FUNC[type(compacted_table)](compacted_table),
-            manifest.meta.content_length,
-            len(compacted_table),
-        ),
-    )
-    logger.info(f"Materialize result: {materialize_result}")
+
+        # Write manifests up to max_records_per_output_file
+        if compacted_tables and \
+                total_record_count + record_count > max_records_per_output_file:
```
raghumdani, line 186: what if single delta manifest entry has more than max_records_per_output_file record?

raghumdani (commented):
Overall, changes look good.

deltacat/compute/compactor/steps/materialize.py
raghumdani: Ok.

valiantljk (commented):
LGTM. Thanks for the change. 
Any performance comparison after the change, e.g., memory consumption reduction rate.

deltacat/compute/compactor/steps/materialize.py
raghumdani: Makes sense..

deltacat/compute/compactor/steps/materialize.py
raghumdani: redundant code always causes confusion. The next reader here would be thinking why do we have to del here, and are there implications of removing this.

deltacat/compute/compactor/steps/materialize.py
rkenmi: We raise an exception asking the client to increase the `records_per_compacted_file` parameter. This should generally be okay since this parameter is adjustable by the caller on a per-compaction job/round basis.

deltacat/compute/compactor/steps/materialize.py
rkenmi: Agreed, we could probably get rid of all `del` usages in this function altogether.

deltacat/compute/compactor/steps/materialize.py
raghumdani: got it.

raghumdani (approved):
LGTM

valiantljk (approved):

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -101,69 +149,60 @@ def materialize(
             src_file_idx_np.item(),
             file_reader_kwargs=read_kwargs_provider,
         )
-        mask_pylist = list(repeat(False, len(pa_table)))
+        record_count = len(pa_table)
+        if record_count > max_records_per_output_file:
```
pdames, line 153: I think we'll want to get rid of this in a follow-up PR since some regressive tables will present 1 or more overly large files from amongst their deltas (meaning that their file size is a bottleneck for bulk parallel consumption) and our job in general should be to always produce a read-optimized view of the table by the end of materialize by not only merging small files up to `max_records_per_output_file`, but also splitting large files down to `max_records_per_output_file`.

PR #54, opened 2023-02-01
Refactor execute_compaction_round and placement group utils

valiantljk:
Tested with one partition. Using bundle id in pg will be submitted in a separate pr

* make the _execute_compaction_round as a normal python function
* modify placement group to take node resource id
* modify round robin options provider to take node resource id as well as placement group strategy
* update round robin option provider that is passed in hb/dd/mat


deltacat/compute/compactor/compaction_session.py
```diff
@@ -172,35 +166,26 @@ def _execute_compaction_round(
...
-    # cluster_cpus = int(cluster_resources["CPU"])
-    # logger.info(f"Total cluster CPUs: {cluster_cpus}")
-
-    # collect node group resources
-
     cluster_resources = ray.cluster_resources()
     logger.info(f"Total cluster resources: {cluster_resources}")
     if pg_config: # use resource in each placement group
-        node_resource_keys=None
         cluster_resources = pg_config[1]
```
raghumdani, line 172: The implicit assumption that `pg_config[1]` is `cluster_resources`, and `pg_config[0]` is remote options makes it difficult to understand the code. Can we either document as method docstrings or create a separate class representing the data structure?

deltacat/utils/placement.py
```diff
@@ -223,12 +228,13 @@ def _config(instance_cpus: int, instance_type: int, time_out: Optional[float] =
 				pg_res['CPU']+=node_detail['resources_total']['CPU']
 				pg_res['memory']+=v['memory']
 				pg_res['object_store_memory']+=v['object_store_memory']
-				pg_res['node_id'].append(node_id)
 		cluster_resources['CPU'] = int(pg_res['CPU'])
 		cluster_resources['memory'] = float(pg_res['memory'])
 		cluster_resources['object_store_memory'] = float(pg_res['object_store_memory'])
-		cluster_resources['node_id'] = pg_res['node_id']
-		pg_config=[opts,cluster_resources]
+		cluster_resources['node_id'] = node_res_keys # bundle_id
```
raghumdani, line 234: If the cluster has 100 nodes but num_bundles is 90, we just want to round robin on 90 nodes. Hence, you may want to trim this list to take only num_bundles items or pass in only num_bundles items to this method.

deltacat/utils/ray_utils/concurrency.py
```diff
@@ -107,4 +107,11 @@ def round_robin_options_provider(
     assert resource_keys, f"No resource keys given to round robin!"
     resource_key_index = i % len(resource_keys)
     key = resource_keys[resource_key_index]
-    return {"resources": {key: resource_amount_provider(resource_key_index)}}
+    if kwargs.get("pg_config"):
```
raghumdani, line 110: we also need to update bundle index within the pg_config right?

deltacat/utils/ray_utils/concurrency.py
```diff
@@ -107,4 +107,11 @@ def round_robin_options_provider(
...
+        #use pg and bundle id for fault-tolerant round-robin
+        opt = {"resources": {key: resource_amount_provider(resource_key_index)}}
```
raghumdani, line 112: can this line be omitted or providing bundle index in `resources` behave same way as specifying a new scheduling strategy object?

setup.py
```diff
@@ -35,14 +35,14 @@ def find_version(*paths):
...
+        "boto3 == 1.20.40",
         "numpy == 1.21.5",
         "pandas == 1.3.5",
         "pyarrow == 10.0.1",
         "pydantic == 1.10.4",
         "ray[default] == 2.0.0",
-        "s3fs == 2022.2.0",
-        "tenacity == 8.1.0",
+        "s3fs == 2022.1.0",
+        "tenacity == 8.0.1",
```
raghumdani, line 45: why are we modifying these version again? we found the old versions to be functional on Manta.

raghumdani (commented):
Overall looks good. Please address high level comments.

deltacat/compute/compactor/compaction_session.py
valiantljk: makes sense.

deltacat/utils/placement.py
valiantljk: the current `node_res_keys` is only the available node ids in this placement group, so should be fine.

deltacat/utils/ray_utils/concurrency.py
valiantljk: yeah, that will be next pr. This pr only tries to refactor the execute_compaction_round and make it working with existing pg which assumes node_id as resource key.

deltacat/utils/ray_utils/concurrency.py
valiantljk: same as previous comment

setup.py
valiantljk: okey, let me change it back

pdames (commented):
Closing in favor of https://github.com/ray-project/deltacat/pull/58. Feel free to re-open if needed.

PR #56, opened 2023-02-01
Optimized the dedupe step and writing primary key index

raghumdani:
The actor RecordCountsPendingMaterialize takes up huge memory on a node. Moreover, it takes forever to process additional record counts after a certain threshold. This commit gets rid of the actor. A separate PR will introduce rebuilding primary key index that can be leveraged by the current compaction round. Also fixed the materialize to process dedupe results. Tested on two different tables.

### DIFF
--- a/deltacat/aws/s3u.py
+++ b/deltacat/aws/s3u.py
@@ -35,8 +35,6 @@ from typing import Any, Callable, Dict, List, Optional, Generator, Union
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-
-@ray.remote
 class CapturedBlockWritePaths:
     def __init__(self):
         self._write_paths: List[str] = []
@@ -70,14 +68,14 @@ class UuidBlockWritePathProvider(BlockWritePathProvider):
     """Block write path provider implementation that writes each
     dataset block out to a file of the form: {base_path}/{uuid}
     """
-    def __init__(self, capture_actor: CapturedBlockWritePaths):
+    def __init__(self, capture_object: CapturedBlockWritePaths):
         self.write_paths: List[str] = []
         self.block_refs: List[ObjectRef[Block]] = []
-        self.capture_actor = capture_actor
+        self.capture_object = capture_object
 
     def __del__(self):
         if self.write_paths or self.block_refs:
-            self.capture_actor.extend.remote(
+            self.capture_object.extend(
                 self.write_paths,
                 self.block_refs,
             )
@@ -357,8 +355,8 @@ def upload_table(
     if s3_table_writer_kwargs is None:
         s3_table_writer_kwargs = {}
 
-    capture_actor = CapturedBlockWritePaths.remote()
-    block_write_path_provider = UuidBlockWritePathProvider(capture_actor)
+    capture_object = CapturedBlockWritePaths()
+    block_write_path_provider = UuidBlockWritePathProvider(capture_object)
     s3_table_writer_func(
         table,
         s3_base_url,
@@ -369,8 +367,8 @@ def upload_table(
     )
     # TODO: Add a proper fix for block_refs and write_paths not persisting in Ray actors
     del block_write_path_provider
-    block_refs = ray.get(capture_actor.block_refs.remote())
-    write_paths = ray.get(capture_actor.write_paths.remote())
+    block_refs = capture_object.block_refs()
+    write_paths = capture_object.write_paths()
     metadata = _get_metadata(table, write_paths, block_refs)
     manifest_entries = ManifestEntryList()
     for block_idx, s3_url in enumerate(write_paths):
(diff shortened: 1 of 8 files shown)


rkenmi:
Thanks for also including some minor bug fixes for https://github.com/ray-project/deltacat/pull/52

- Unused `enumerate()` over collection and not unpacking tuple values
- Removing exceptions for delta entry records > `records_per_compacted_file`, since this can cause a regression with uploading gigantic deltas if clients specify a incredibly large `records_per_compacted_file` value

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -29,7 +29,7 @@
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 
-@ray.remote(num_cpus=0.5)
+@ray.remote
 def materialize(
```
pdames, line 33: This isn't a result of this PR, but I noticed that the recent change to eagerly evict tables from materialize memory has introduced a side-effect that, whenever we have more than `max_records_per_output_file` to materialize, we materialize `max_records_per_output_file` to file 1-N (where N will also usually be 1), and then write the remainder to the next file.

This (1) erodes our ability to produce a read-optimized output table and (2) broke the source -> dest primary key index record mapping performed by the actor removed in this PR (since it assumed that materialize will write `max_records_per_output_file` for each file, and then only chop off the remainder once across files assigned to a materialize task). The second doesn't matter if we don't ever go back to mapping records from src -> dest at the end of the `dedupe` step, but the first is still an issue.

We should also capture an issue to fix this in a follow-up PR where, instead of chopping off the remainder over `max_records_per_output_file` every time we met or exceed this record count, we write one file and then continue appending records to the remainder. A remainder with less than `max_records_per_output_file` should only be written when no more records are left for the task to materialize.

deltacat/compute/compactor/utils/primary_key_index.py
```diff
@@ -8,10 +8,10 @@
 
 from deltacat.utils.common import ReadKwargsProvider
 from ray import cloudpickle
+from deltacat.constants import PRIMARY_KEY_INDEX_WRITE_BOTO3_CONFIG
```
pdames, line 11: We may want to capture a TODO here that also helps decouple us from S3 in general by just writing the primary key index files back to the catalog using the existing DeltaCAT client APIs. This should defer all writes to the underlying file system in use by the catalog, which should already have considered strategies to better avoid/mitigate throttling issues from cloud storage providers like S3.

pdames (approved):
Looks good overall - thanks! We'll just want to make sure to capture and address all of the follow-up items in subsequent PRs.

deltacat/compute/compactor/steps/materialize.py
raghumdani: Agree. The implicit coupling makes the materialize implementation fragile. As long we enforce a contract through some unit tests, any new developer will be reminded of following the contract. For now, have added a TODO.

deltacat/compute/compactor/utils/primary_key_index.py
raghumdani: Makes sense.

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -126,87 +101,18 @@ def drop_duplicates_by_primary_key_hash(table: pa.Table) -> pa.Table:
     return table.take(list(value_to_last_row_idx.values()))
 
 
-def write_new_primary_key_index(
+def _write_new_primary_key_index(
```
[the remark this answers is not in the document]
raghumdani: Right, I have a doc that summarizes the next steps related to this.

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -36,14 +36,14 @@
...
-def union_primary_key_indices(
+def _union_primary_key_indices(
         s3_bucket: str,
         round_completion_info: RoundCompletionInfo,
         hash_bucket_index: int,
         df_envelopes_list: List[List[DeltaFileEnvelope]]) -> pa.Table:
 
-    logger.info(f"Reading dedupe input for {len(df_envelopes_list)} "
-                f"delta file envelope lists...")
+    logger.info(f"[Hash bucket index {hash_bucket_index}] Reading dedupe input for "
```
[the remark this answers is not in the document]
raghumdani: Yes, the issue seems to be already linked to this PR.

deltacat/aws/s3u.py
```diff
@@ -35,8 +35,6 @@
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-
-@ray.remote
 class CapturedBlockWritePaths:
```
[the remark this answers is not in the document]
raghumdani: Agreed. Added a TODO

PR #58, opened 2023-02-03
Improve compaction fault-tolerance with refactored placement group api

valiantljk:
(https://github.com/ray-project/deltacat/issues/45)

With this PR
* the placement group options is updated with bundle id instead of using node id
* tasks will be round-robinly placed onto different bundles
* bundle strategy default is SPREAD, but can be PACK or others specified by users
* new API takes `number of placement group to be created`, `total cpus per pg`, `cpus per bundle` , and `bundle strategy`, `capture_child_tasks ` as parameters
* new way to figure out head id with ray.experimental api
* avoided creating pg on head by launching the pg creation remote task on head


setup.py
```diff
@@ -35,7 +35,7 @@ def find_version(*paths):
     packages=setuptools.find_packages(where=".", include="deltacat*"),
     install_requires=[
         # any changes here should also be reflected in requirements.txt
-        "boto3 == 1.20.24",
+        "boto3 == 1.20.40",
```
pdames, line 38: Was this kept intentionally? When testing with Python 3.7 on my laptop, `pip install` fails to resolve compatible dependencies when `boto3 == 1.20.40`.

deltacat/utils/ray_utils/concurrency.py
```diff
@@ -104,7 +104,13 @@ def round_robin_options_provider(
         foo.options(**opt).remote()
     ```
     """
-    assert resource_keys, f"No resource keys given to round robin!"
-    resource_key_index = i % len(resource_keys)
-    key = resource_keys[resource_key_index]
-    return {"resources": {key: resource_amount_provider(resource_key_index)}}
+    opts = kwargs.get("pg_config")
```
pdames, line 107: Can we just add this as an additional named argument instead to make it easier to document going forward (e.g. add `pg_config` together with `resource_amount_provider` and `lambda i`)?

deltacat/utils/ray_utils/concurrency.py
```diff
@@ -104,7 +104,13 @@ def round_robin_options_provider(
...
+    if opts: #use pg and bundle id for fault-tolerant round-robin
```
pdames, line 108: Nit: space after `#`.

```suggestion
    if opts: # use pg and bundle id for fault-tolerant round-robin
```

pdames, line 108: We can just continue use the more specific name of `pg_config` if we move it into a named keyword argument in `round_robin_options_provider`.

pdames (approved):
LGTM - just a few minor comments. Feel free to merge when ready.

deltacat/utils/placement.py
```diff
@@ -158,7 +159,17 @@ def get_group_by_name(self, gname: str) -> Optional[Dict[str, Union[str, float]]
 			logger.info(f"There is no available resources for {gname}")
 			return None
 		return group_res
-
+		
+class PlacementGroupResource():
```
raghumdani, line 163: Minor: `PlacementGroupConfig` would be more accurate?

raghumdani (approved):

setup.py
pdames:
```suggestion
        "boto3 == 1.20.24",
```

PR #68, opened 2023-02-04
Fix placement group to resolve variables

raghumdani:
Fix below issues:

1. rename `instance_type`.
2. call class method using self. 
3. rename to `PlacementGroupConfig`.


pdames (approved):
LGTM!

Issue #69, opened 2023-02-04
Support limiting deltas entries in a compaction round


raghumdani:
Currently, we only limit deltas in a compaction round based on total object store memory available in a cluster. When there is a very large delta that contains many manifest files, we still have to limit them and perform re-batching.

pdames:
From https://github.com/ray-project/deltacat/pull/70:

> ... the current contract of compaction assumes that each round must be able to compact at least one delta. To work with extremely large deltas we'll need to drive that down to at least file-level granularity (which will drive subsequent changes into the Round Completion File and each round that reads it to determine a starting point). Future improvements would then include driving each round down to record-level granularity to work with files that are too large to complete in a single round.

raghumdani:
Primary key index building is a pre-requisite to running multiple rounds: #63

PR #70, opened 2023-02-04
Support table compaction source rebases and refactor compact_partition

pdames:
See user guide in included `README.md` for compaction source rebase use-cases and invocation examples.
As part of this PR, the `compact_partition` API has also been refactored for better backwards-compatibility going forward by using keyword arguments for all optional args.


deltacat/compute/compactor/README.md
```diff
@@ -0,0 +1,107 @@
...
+`source_partition_locator` and `rebase_source_partition_locator` to the same
+value:
+```python
+compact_partition(
+  source_partition_locator=source_partition_to_compact,
+  compacted_partition_locator=deltacat_compacted_partition,
+  primary_keys=delta_source_primary_keys,
+  compaction_artifact_s3_bucket=deltacat_s3_bucket,
+  last_stream_position_to_compact=delta_source_last_stream_position,
+  rebase_source_partition_locator=source_partition_to_compact,
```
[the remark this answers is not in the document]
raghumdani: I'd prefer having `rebase_source_partition_locator` as it's more generic and allows for both rebases and backfills. Also, there is an inherent coupling between `rebase locator` and `source locator` as one is derived from other in most cases.

deltacat/compute/compactor/README.md
```diff
@@ -0,0 +1,107 @@
...
+the new compactor will read and compact deltas from, then your first call to
+`compact_partition` should look something like this:
+```python
+compact_partition(
+  source_partition_locator=old_compacted_partition,
+  compacted_partition_locator=deltacat_compacted_partition,
+  primary_keys=delta_source_primary_keys,
+  compaction_artifact_s3_bucket=deltacat_s3_bucket,
+  last_stream_position_to_compact=delta_source_last_stream_position,
+  rebase_source_partition_locator=delta_source_partition,
```
yiqinzhu, line 40: Could we relate the partitions mentioned here to  `S1`, `S2`, `D` as stated in the above explanation? For example, something like `delta_source_partition(S2)`, `deltacat_compacted_partition(D)`.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -68,9 +69,10 @@ def compact_partition(
         min_hash_bucket_chunk_size: int = 0,
         compacted_file_content_type: ContentType = ContentType.PARQUET,
         delete_prev_primary_key_index: bool = False,
-        read_round_completion: bool = False,
         pg_config: Optional[PlacementGroupConfig] = None,
         schema_on_read: Optional[pa.schema] = None,  # TODO (ricmiyam): Remove this and retrieve schema from storage API
+        rebase_source_partition_locator: Optional[PartitionLocator] = None,
+        rebase_source_partition_high_watermark: Optional[int] = None,
```
raghumdani, line 75: can we get this info directly from `rebase_source_partition_locator` and deltacat storage API?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -213,9 +216,10 @@ def _execute_compaction_round(
...
     # a compatible primary key index
     round_completion_info = None
-    if read_round_completion:
-        logger.info(f"Reading round completion file for compatible "
-                    f"primary key index root path {compatible_primary_key_index_root_path}")
+    if not rebase_source_partition_locator:
+        logger.info(
+            f"Reading round completion file for compatible "
+            f"primary key index root path: {compatible_primary_key_index_root_path}")
         round_completion_info = rcf.read_round_completion_file(
```
raghumdani, line 223: How would we handle the case when `rebase` itself requires multiple compaction rounds? I think it will be cleaner to separate out `round_completion_files` from `compaction_completion_files`. If the user passes same params, it will re-run all annotated deltas again in an infinite loop.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -68,9 +69,10 @@ def compact_partition(
         min_hash_bucket_chunk_size: int = 0,
         compacted_file_content_type: ContentType = ContentType.PARQUET,
         delete_prev_primary_key_index: bool = False,
-        read_round_completion: bool = False,
         pg_config: Optional[PlacementGroupConfig] = None,
         schema_on_read: Optional[pa.schema] = None,  # TODO (ricmiyam): Remove this and retrieve schema from storage API
+        rebase_source_partition_locator: Optional[PartitionLocator] = None,
+        rebase_source_partition_high_watermark: Optional[int] = None,
         deltacat_storage=unimplemented_deltacat_storage):
```
raghumdani, line 76: Could we return the final result as the location to the file so the orchestrator could cache the results for easy discoverability?

raghumdani (commented):
Overall looks good.

deltacat/compute/compactor/compaction_session.py
pdames: Eventually, I think so - I'll add a TODO for this and open a related issue to propose a storage API for this and leverage it for simplicity of management of catalogs that can support and implement it today, since it will simplify operations in cases where the compactor can determine the old high watermark automatically.

Right now we're lacking any official or consistent specification of how to do this across various data catalogs, but we could consider adding a storage API that simply agrees to return the latest copy-on-write compaction high-watermark for any partition locator, and leave it up to the catalog to determine whether they have a supported implementation for this API.

This currently just defers to the simple answer of "the caller figured out the high watermark somehow" over the more user-friendly ideal of "the compactor will figure this out for you as long as the catalog has implemented an API to automatically retrieve the high watermark."

deltacat/compute/compactor/compaction_session.py
pdames: I'll create a separate issue to address this in a follow-up PR, since the current contract of compaction assumes that each round must be able to compact at least one delta. To work with extremely large deltas we'll need to drive that down to at least file-level granularity (which will drive subsequent changes into the Round Completion File and each round that reads it to determine a starting point). Future improvements would then include driving each round down to record-level granularity to work with files that are too large to complete in a single round.

deltacat/compute/compactor/compaction_session.py
pdames: The location of the round completion file? Sure, that makes sense.

raghumdani: yes

valiantljk:
LGTM, with this pr, we also need to modify the `io.discovery_deltas` to combine both rebased and source tables' deltas in case there is no primary key index from prior round. https://github.com/ray-project/deltacat/issues/76

deltacat/compute/compactor/compaction_session.py
pdames: Related Issue: https://github.com/ray-project/deltacat/issues/69

deltacat/compute/compactor/compaction_session.py
pdames: Related Issue: https://github.com/ray-project/deltacat/issues/69

pdames:
> LGTM, with this pr, we also need to modify the `io.discovery_deltas` to combine both rebased and source tables' deltas in case there is no primary key index from prior round. #76

Will address this in a follow-up PR that branches compactor code into implementations that run w/ and w/o PKI.

raghumdani (approved):
LGTM

Issue #65, opened 2023-02-03
Maximize file sizes produced by each materialize task


pdames:
This isn't a result of this PR, but I noticed that the recent change to eagerly evict tables from materialize memory has introduced a side-effect that, whenever we have more than `max_records_per_output_file` to materialize, we materialize `max_records_per_output_file` to file 1-N (where N will also usually be 1), and then write the remainder to the file N+1.

This (1) erodes our ability to produce a read-optimized output table and (2) broke the source -> dest primary key index record mapping performed by the actor removed in this PR (since it assumed that materialize will only have at most one file with a remainder less than `max_records_per_output_file` per task). 

(2) doesn't matter if we never go back to mapping records from src -> dest to resolve a primary key index at the end of the `dedupe` step, but (1) will remain an issue regardless of our primary key index resolution strategy.

We should also capture an issue to fix this in a follow-up PR where, instead of chopping off the remainder over `max_records_per_output_file` every time we met or exceed this record count, we write one file and then continue appending records to the remainder. A remainder with less than `max_records_per_output_file` should only be written when no more records are left for the task to materialize.

_Originally posted by @pdames in https://github.com/ray-project/deltacat/pull/56#discussion_r1095335528_

pdames:
For example, if the input to a `materialize` task run is 32-million input records from 4 ordered input files with respective record counts [8MM, 4MM, 18MM, 2MM] then the current code would write 3 output files with suboptimal record counts (shown in **bold**) [10MM, **2MM**, 10MM, **8MM**, **2MM**] instead of the optimal result of only writing 1 output file with a suboptimal record count [10MM, 10MM, 10MM, **2MM**].

PR #71, opened 2023-02-05
Maximize file sizes produced by each materialize task

rkenmi:
Issue: #65 

Create read-optimized number of output files in materialize while retaining minimal heap memory usage.

Example compaction job with `records_per_output_file = 100_000`, each debug output = one materialize task:
```csv
  | 2023-02-08T17:52:29.725-08:00 | 2023-02-09 01:52:29,530 DEBUG materialize.py:226 -- 3 files written with records: [700000, 100000, 53352]
  | 2023-02-08T17:52:32.038-08:00 | 2023-02-09 01:52:32,021 DEBUG materialize.py:226 -- 3 files written with records: [700000, 200000, 58439]
  | 2023-02-08T17:52:43.315-08:00 | 2023-02-09 01:52:43,143 DEBUG materialize.py:226 -- 3 files written with records: [300000, 100000, 33628]
  | 2023-02-08T17:52:43.342-08:00 | 2023-02-09 01:52:43,160 DEBUG materialize.py:226 -- 3 files written with records: [700000, 100000, 12591]
  | 2023-02-08T17:52:43.757-08:00 | 2023-02-09 01:52:43,508 DEBUG materialize.py:226 -- 3 files written with records: [700000, 100000, 59279]
  | 2023-02-08T17:52:47.608-08:00 | 2023-02-09 01:52:47,587 DEBUG materialize.py:226 -- 3 files written with records: [200000, 100000, 41779]
```


deltacat/compute/compactor/steps/materialize.py
```diff
@@ -154,8 +149,7 @@ def _materialize(
             src_file_idx_np.item(),
             file_reader_kwargs_provider=read_kwargs_provider,
         )
-        record_count = len(pa_table)
-        mask_pylist = list(repeat(False, record_count))
+        mask_pylist = list(repeat(False, len(pa_table)))
```
raghumdani, line 152: We don't need to create a mask pylist after my latest changes. We could directly pass pyarrow array here to `take` indices that we need.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -181,27 +175,52 @@ def _materialize(
             [sc._DEDUPE_TASK_IDX_COLUMN_NAME]
         )
```
raghumdani, line 176: Sorting must be done after merging all the slices that go into a single file. However, we would not need this sorting if we are not predicting the indices using record counts.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -181,27 +175,52 @@ def _materialize(
...
 
-        # Write manifests up to max_records_per_output_file
-        # TODO(raghumdani): Write exactly the same number of records into each file to
-        # produce a read-optimized view of the tables.
-        if compacted_tables and \
-                total_record_count + record_count > max_records_per_output_file:
-            materialized_results.append(_materialize(compacted_tables, total_record_count))
+        if remainder_tables:
+            prev_remainder_record_count = sum([len(tbl) for tbl in remainder_tables])
```
raghumdani, line 179: This may become inefficient if we end up iterating through all the remainder tables especially if there are lot of small tables and our max_records_per_output_file is too huge. Basically, increasing the complexity from O(N) to skim through this to O(N^2).

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -181,27 +175,52 @@ def _materialize(
...
+                f" to be fitted to {max_records_per_output_file}, but got {remainder_records} instead."
+            materialized_results.append(_materialize(remainder_tables))
             # Free up written tables in memory
-            compacted_tables.clear()
-            total_record_count = 0
+            remainder_tables.clear()
+
+            # Update the starting record offset of the current file
+            if new_file_record_offset < len(pa_table):
+                pa_table = pa_table.slice(offset=new_file_record_offset)
```
raghumdani, line 203: Have you tested for the case where this table turns out to be empty?

deltacat/utils/pyarrow.py
```diff
@@ -133,6 +133,24 @@ def slice_table(
     return tables
 
 
+def slice_table_generator(
+        table: pa.Table,
+        max_len: int) -> Generator[pa.Table, None, None]:
+    """
```
raghumdani, line 139: You could document this public method, and mention this wouldn't yield any results when the table is empty.

deltacat/utils/pyarrow.py
```diff
@@ -133,6 +133,24 @@ def slice_table(
...
+    Generator for creating 0-copy table slices.
+    """
+    offset = 0
+    records_remaining = len(table)
+    while records_remaining > 0:
```
raghumdani, line 144: When the max records per file is say 100 (N) but your table has 100000 (M) records, this would yield 1000 smaller pyarrow tables. The 1000 compute iterations could be saved if you just split it into two parts one of size `floor(M/N)*N` and second of `M%N`.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -154,8 +149,7 @@ def _materialize(
             src_file_idx_np.item(),
             file_reader_kwargs_provider=read_kwargs_provider,
         )
-        record_count = len(pa_table)
```
raghumdani, line 157: This was a bug initially which counted record counts of original `pa_table` and not the filtered table. We may need to setup unit test and integration to catch these kind of issues going forward.

raghumdani (commented):
Overall, as advertised. However, you may need to do some more optimizations to merge.

deltacat/utils/pyarrow.py
raghumdani: The idea here is to save the 1000 compute iterations everytime we add a new entry to remainder table.

rkenmi: The slicing only occurs on `stage_delta` if the table being passed in has greater record counts than `max_records_per_output_file`. So by effectively passing in a concatenated table with a guaranteed size of `max_records_per_output_file` (or less than, for the final remainder), we can guarantee that each concatenated table would result in one file written to S3 (and not have any further splitting).

Basically, pre-splitting to `[100, 100, 100, ..., 100, 1]` or just bundling upfront to `[100_000, 1]` doesn't really change the fact that we compute 1000 iterations. It might be a bit nuanced behavior, but pre-splitting is convenient here since we are also slicing and merging table records based on the offset of previous remainders.

raghumdani: Correct me if I am wrong, don't we end up calling data catalog multiple times to create new "in-progress" delta for each file instead of calling data catalog once for all the files uploaded during a single `stage_delta` call.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -45,16 +47,20 @@ def materialize(
...
+        logger.debug(f"Concatenating {len(compacted_tables)} compacted tables"
+                     f" with total size: {compacted_tables_size} bytes")
         compacted_table = pa.concat_tables(compacted_tables)
+        compacted_table = compacted_table.take(
+            pc.sort_indices(compacted_table, sort_keys=PA_SORT_KEYS),
+        )
+        compacted_table = compacted_table.drop(
+            [sc._DEDUPE_TASK_IDX_COLUMN_NAME]
+        )
+
```
raghumdani, line 63: To be precise, this sorting can be removed as we are not predicting the indices during dedupe.

deltacat/utils/pyarrow.py
rkenmi: Currently, the data catalog API internally slices a given table into groups of `max_records_per_output_file` tables, and performs a single S3 call per table.

https://github.com/ray-project/deltacat/blob/d4b307cae12c8d8a6f6ccbc1fe6169e8a3fd20e5/deltacat/aws/s3u.py#L281

rkenmi: I've since revised my changes to do post-splitting instead, since the data catalog API call itself could be expensive and it would be good to reduce the number of these calls as much as we can (we'll still have to live with more than a single stage_delta call per materialize task, if we eagerly materialize).

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -92,7 +94,10 @@ def _materialize(
         logger.info(f"Materialize result: {materialize_result}")
         return materialize_result
 
-    logger.info(f"Starting materialize task...")
+    # TODO (ricmiyam): Update this for full implementation of https://github.com/ray-project/deltacat/issues/62
+    materialize_task_id = get_runtime_context().get_task_id()
+    logger.info(f"Starting materialize task {materialize_task_id}...")
```
raghumdani, line 99: Minor: Also what could be helpful here is to print materialize bucket index.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -164,49 +172,66 @@ def _materialize(
...
-        # TODO(raghumdani): Write exactly the same number of records into each file to
-        # produce a read-optimized view of the tables.
-        if compacted_tables and \
-                total_record_count + record_count > max_records_per_output_file:
-            materialized_results.append(_materialize(compacted_tables, total_record_count))
+            # Chop off records that are already written to disk
+            pa_table = pa_table.slice(offset=records_to_fit)
+
+        record_count = len(pa_table)
+        assert record_count > 0, f"Expected non-empty compacted PyArrow table"
```
raghumdani, line 195: Would this always be true? What happens when `remainder_record_count + len(pa_table)` is equal to `max_records_per_output_file` ?

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -164,49 +172,66 @@ def _materialize(
...
+            materialized_results.append(_materialize(remainder_tables))
             # Free up written tables in memory
-            compacted_tables.clear()
-            total_record_count = 0
+            remainder_tables.clear()
+            remainder_record_count = 0
 
-        total_record_count += record_count
-        compacted_tables.append(pa_table)
+        # Add any leftovers (less than max_records_per_output_file) to the remainder list after packing
```
raghumdani, line 211: Instead of adding comments what the code next to it is doing (which is fairly easy to understand), what you may add is a doc to this method and explain `why` part i.e., why does it have to add left overs to remainder table.

deltacat/compute/compactor/steps/materialize.py
rkenmi: No, thanks for catching (this is why we need unit tests) 🙂

deltacat/compute/compactor/steps/materialize.py
rkenmi: How about a separate issue altogether for adding docstrings to each of the compaction steps?

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -164,49 +172,66 @@ def _materialize(
...
+            if remainder_record_count + len(pa_table) < max_records_per_output_file:
+                remainder_tables.append(pa_table)
+                remainder_record_count += len(pa_table)
+                continue
+
+            # 'max_records_per_output_file' will be exceeded with the
+            #  current file's records + previous remainder records.
+            records_to_fit = max_records_per_output_file - remainder_record_count
+            fitted_table = pa_table.slice(length=records_to_fit)
+            remainder_tables.append(fitted_table)
```
raghumdani, line 189: This version looks good. However, I still feel it could be further modularized to make it easy to understand. Remember, we also have to add indices to these records as and when they are getting materialized to support primary key index building in future. In the current state, it may be difficult to make those modifications. I have proposed some changes below. Let me know if that makes sense:

```python
# This class handles splitting the tables and letting us know when it's ready to materialize
class BatchedTables(List[pa.Table]):
    def __init__(self, max_records_per_table_multiple):
        self.__init__()
        self._tables_ready = []
        self._tables_leftover = []
        self._tables_leftover_len = 0
        self._tables_ready_len = 0
    def append(self, table: pa.Table):
        # split into two in the multiples of max_records and left over and append to
        # appropriate lists created above
        if (_tables_leftover_len > max_records):
            # time to split move some of them to _ready_tables

    def get_ready_tables() -> List:
        return self._tables_ready

    def clear_ready() -> List:
        self._tables_ready.clear()

    def get_remainder_tables() -> List:
        return self._tables_leftover
  
# In your materialize code

batched_tables = BatchedTables(max_records_per_file)

for table in tables:
    batched_tables.append(table)
    if (len(batched_tables.get_ready_tables()) > 0):
        _materialize(batched_tables.get_ready_tables())
        batched_tables.clear_ready_tables()

if (batched_tables.get_ready_tables()):
    _materialize(batched_tables.get_ready_tables())
    batched_tables.clear_ready_tables()

if (batched_tables.get_remainder_tables()):
    _materialize(batched_tables.get_remainder_tables())

```

deltacat/utils/pyarrow.py
```diff
@@ -18,7 +18,7 @@
 from deltacat.utils.common import ReadKwargsProvider, ContentTypeKwargsProvider
 from deltacat.utils.performance import timed_invocation
 
-from typing import Any, Callable, Dict, List, Optional, Iterable, Union
+from typing import Any, Callable, Dict, List, Optional, Iterable, Union, Generator
```
raghumdani, line 21: Is this needed now?

raghumdani (commented):
Looks good from performance aspect. Please try to address some minor comments related to readability.

deltacat/compute/compactor/steps/materialize.py
raghumdani: Makes sense.

deltacat/compute/compactor/steps/materialize.py
rkenmi: I like the general idea. I think there could be an unintended performance hit here when doing the split to move `self._tables_leftover` over to `self._ready_tables` too frequently or across too many tables?

raghumdani: I think we can mitigate that by using `double ended queue` if required as we have to maintain table ordering.

rkenmi: In the class you provided, I was more so referring to how we would need to concatenate leftover tables and split them, each time we need to shift `self._tables_leftover` over to `self._ready_tables`, since we maintain two lists of tables now. What if max records is `4000`, and each table added has 3999 records?

raghumdani: Doesn't the current implementation also require you to split pyarrow table everytime in the above case? As far as shifting the table goes, you will only have shift a particular table only once, so it will not affect performance. You can optimally shift if you use `double ended queues` instead of lists.

rkenmi: The current impl. splits once per table, and enforce the remainder to always be less than max records for the next iteration. In this multi-list (or deque) approach we will split once per table *and* also keep track of the leftovers, triggering a shift if it exceeds max records. 

For example
```
tables = [7999, 7999, 7999, ...]

bt.append(tables[0]) # leftover = [3999], ready = [4000] (one split)
bt.append(tables[1]) # leftover = [3998], ready = [4000, 4000, 4000] (one split + one shift)
bt.append(tables[2]) # leftover = [3997], ready = [4000, 4000, 4000, 4000, 4000] (one split + one shift)
...
```

rkenmi: I think the point being is that there can be other interesting ways to have this BatchedTable class, but my question is whether or not this modularization has to be P0 itself? We could have a separate issue/PR to refactor this into a class, and even have class-specific unit tests which will be easier to write vs. writing unit tests for the materialize task as a whole.

raghumdani: You may think this is not p0, but without the unit tests a small change could cause data corruption. Also, you will find it hard to unit test this code vs modularized code. That way it will be easier for someone to make changes and add new features.

deltacat/tests/utils/test_record_batch_tables.py
```diff
@@ -0,0 +1,241 @@
+import unittest
+import pyarrow as pa
+
+from deltacat.utils.pyarrow import RecordBatchTables
+
+
+def is_sorted(batched_tables: RecordBatchTables,
```
raghumdani, line 7: Minor: these methods are not reusable across different tests, so we could move them to the end of the file and start with `_` (underscore).

raghumdani (approved):
LGTM!

pdames (approved):
Looks good to merge. Thanks for the unit tests! Minor refactoring can be addressed either prior to merge or in a follow-up PR.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -92,7 +96,11 @@ def _materialize(
         logger.info(f"Materialize result: {materialize_result}")
         return materialize_result
 
-    logger.info(f"Starting materialize task...")
+    # TODO (ricmiyam): Update this for full implementation of https://github.com/ray-project/deltacat/issues/62
+    materialize_task_id = get_runtime_context().get_task_id()
```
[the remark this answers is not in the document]
rkenmi: From my compaction E2E test runs (including hash bucketing, dedupe, materialize), the use of `get_runtime_context()` didn't add any noticeable latency and the time taken for the whole workflow to complete was relatively the same. We also just merged a PR (https://github.com/ray-project/deltacat/pull/85) to inject runtime context into our logs, so I can clean up the runtime_context usages here.

PR #78, opened 2023-02-10
Fix dedupe when iterchunks() doesn't return equal length chunks across columns

raghumdani:
It has been observed that iterchunks() method of pyarrow sometimes returns unequal length chunks that leads to incorrect dedupe. This could cause record loss downstream.

### DIFF
--- a/deltacat/compute/compactor/steps/dedupe.py
+++ b/deltacat/compute/compactor/steps/dedupe.py
@@ -95,11 +95,12 @@ def _drop_duplicates_by_primary_key_hash(table: pa.Table) -> pa.Table:
     op_type_np = sc.delta_type_column_np(table)
 
     assert len(pk_hash_np) == len(op_type_np), \
-            f"Primary key digest column length ({len(pk_hash_np)}) doesn't match delta type column length ({len(op_type_np)})."
+            f"Primary key digest column length ({len(pk_hash_np)}) doesn't " \
+            f"match delta type column length ({len(op_type_np)})."
 
-    for row_idx in range(len(pk_hash_np)):
-        pk_val = pk_hash_np[row_idx]
-        op_val = op_type_np[row_idx]
+    row_idx = 0
+    pk_op_val_iter = zip(pk_hash_np, op_type_np)
+    for (pk_val, op_val) in pk_op_val_iter:
 
         # operation type is True for `UPSERT` and False for `DELETE`
         if op_val:
--- a/deltacat/compute/compactor/steps/dedupe.py
+++ b/deltacat/compute/compactor/steps/dedupe.py
@@ -98,6 +98,7 @@ def _drop_duplicates_by_primary_key_hash(table: pa.Table) -> pa.Table:
             f"Primary key digest column length ({len(pk_hash_np)}) doesn't " \
             f"match delta type column length ({len(op_type_np)})."
 
+    # TODO(raghumdani): move the dedupe to C++ using arrow methods or similar. 
     row_idx = 0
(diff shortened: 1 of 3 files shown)


deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -93,25 +90,27 @@ def _union_primary_key_indices(
...
-            else:
-                # DELETE this row
-                value_to_last_row_idx.pop(pk_val, None)
-            row_idx += 1
+
+    pk_hash_np = sc.pk_hash_column_np(table)
+    op_type_np = sc.delta_type_column_np(table)
+
+    assert len(pk_hash_np) == len(op_type_np), \
+            "Both pk hash and delta type should have equal length"
```
pdames, line 98: Maybe we can also print the observed length to assist in debugging?
```suggestion
            f"Primary key digest column length ({len(pk_hash_np)}) doesn't match delta type column length ({len(op_type_np)})."
```

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -93,25 +90,27 @@ def _union_primary_key_indices(
...
+
+        # operation type is True for `UPSERT` and False for `DELETE`
+        if op_val:
+            # UPSERT this row
+            value_to_last_row_idx[pk_val] = row_idx
+        else:
+            # DELETE this row
+            value_to_last_row_idx.pop(pk_val, None)
+
+        row_idx += 1
```
pdames, line 112: Seems like this is just overwritten on each iteration of the current loop (i.e. dead code)?

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -93,25 +90,27 @@ def _union_primary_key_indices(
...
-                value_to_last_row_idx.pop(pk_val, None)
-            row_idx += 1
+
+    pk_hash_np = sc.pk_hash_column_np(table)
+    op_type_np = sc.delta_type_column_np(table)
+
+    assert len(pk_hash_np) == len(op_type_np), \
+            "Both pk hash and delta type should have equal length"
+
+    for row_idx in range(len(pk_hash_np)):
```
pdames, line 100: Since `range` will allocate a list of integers equal to the length of pk_hash_np (e.g. maybe millions of integers per invocation?) can we just revert back to the original zipped iterator here instead to conserve some memory and CPU cycles (unless it's actually slower for some reason)?

(We should probably also have a TODO here to delegate this step to C++ eventually - it's just not an observed performance bottleneck today.)

pdames (changes requested):
Good find - just a few minor comments.

deltacat/compute/compactor/steps/dedupe.py
raghumdani: Makes sense, I didn't observe any difference in performance. It may have been overshadowed by the dict. Changing to zip to consume less memory.

pdames (approved):
LGTM

deltacat/compute/compactor/steps/dedupe.py
pdames: Good point about poor python-native dictionary performance. Maybe an easier short-term TODO to address here would also be to just switch from the current dictionary to NumPy arrays?

raghumdani: Makes sense. I have been exploring with numpy `unique` [method](https://numpy.org/doc/stable/reference/generated/numpy.unique.html) with `return_index=True` and mask of `delta_type`.

Issue #82, opened 2023-02-14
Rebatching based on delta locator


valiantljk:
For deltas from two sources: compacted table: A, and uncompacted table: B, 
Current `rebatch` will try to form an equal-sized group of deltas, where entries in each group share the same deltaLocator. This will produce a group that has A's locator but contains entries from both A and B. During materialize, the entries of B can not be found given the wrong delta locator of A. 

To support two data sources, we need to create a new `da` group, whenever the locator is changed. https://github.com/ray-project/deltacat/blob/main/deltacat/compute/compactor/model/delta_annotated.py#L82

PR #84, opened 2023-02-14
Assign new group when locator is different

valiantljk:
#82 , During rebatch, deltas are assigned to groups to form a balanced load for downstream processing. Each group has the same delta locator for all its entries. When delta locator is changed (e.g., from source A to compacted B or vice versa), this pr creates a new delta group `new_da`, and append what has been accumulated in da into final groups.

### DIFF
--- a/deltacat/compute/compactor/model/delta_annotated.py
+++ b/deltacat/compute/compactor/model/delta_annotated.py
@@ -80,6 +80,15 @@ class DeltaAnnotated(Delta):
                    f"({len(src_da_annotations)}) doesn't mach the length of "
                    f"delta manifest entries ({len(src_da_entries)}).")
             for i, src_entry in enumerate(src_da_entries):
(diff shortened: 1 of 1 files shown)


pdames (changes requested):
Main high-level comment is that I had a hard time understanding why we even find ourselves in this situation with the current code-base. It seems that there's also another change that's not shown in this PR to `io.discover_deltas` to discover data from both the source and destination tables, in preparation for branching out a simpler MVP compactor that doesn't use any primary key indices or rounds at all. Maybe it would be better to include this change in the context of that larger milestone, and/or include the working changes end-to-end that simply eliminate anything to do with iterative compaction rounds or primary key indices for now? (we can store ongoing primary key index and/or round-based compaction logic in a separate branch for now, then merge it back in later)

valiantljk:
https://github.com/ray-project/deltacat/pull/86

Issue #57, opened 2023-02-02
Lightweight profiling and straggler detector


valiantljk:
Ray itself has exposed certain metrics (memory, cpu and object store utilization) to facilitate easier debugging and monitoring. In production, though, we may need a more customized profiler that can
* report detailed memory allocation during different stage by different module, e.g., ray, pyarrow, etc
* report straggler in each step, e.g., hash bucket, dedupe and materialize
* aggregate and report any un-common issues during the entire pipeline, e.g., hash bucket takes longer for same amount of workload, etc

And these profiling and reporting should be lightweight, otherwise, we should only turn it on before onboarding or keep the minimum required logger. 

Certain tool to be considered:
* memray: https://github.com/bloomberg/memray
* ray profiling: https://docs.ray.io/en/latest/ray-contribute/profiling.html
* ray 2.3 task level break down

PR #87, opened 2023-02-20
[metrics]Add customized CloudWatch metrics support; Enable straggler detection

Zyiqin-Miranda:
This PR adds support for emitting customized CloudWatch metrics. 
Currently, it's main purpose is for runtime profiling and to detect stragglers during each step of compaction. 
Issues: part of #57

### DIFF
--- a/deltacat/compute/compactor/steps/dedupe.py
+++ b/deltacat/compute/compactor/steps/dedupe.py
@@ -6,6 +6,8 @@ import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc
 import ray
+import time
+import datetime
 from ray import cloudpickle
 from ray.types import ObjectRef
 
@@ -29,12 +31,12 @@ from deltacat.compute.compactor import SortKey, SortOrder, \
 from deltacat.compute.compactor.utils import system_columns as sc, \
     primary_key_index as pki
 from deltacat.utils.performance import timed_invocation
+from deltacat.utils.metrics import emit_timer_metrics, MetricsConfig
 
 from typing import Any, Dict, List, Optional, Tuple
 from deltacat.utils.pyarrow import ReadKwargsProviderPyArrowSchemaOverride
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
-
 
 MaterializeBucketIndex = int
 DeltaFileLocatorToRecords = Dict[DeltaFileLocator, np.ndarray]
@@ -52,7 +54,6 @@ def _union_primary_key_indices(
         round_completion_info: RoundCompletionInfo,
         hash_bucket_index: int,
         df_envelopes_list: List[List[DeltaFileEnvelope]]) -> pa.Table:
-
     logger.info(f"[Hash bucket index {hash_bucket_index}] Reading dedupe input for "
                 f"{len(df_envelopes_list)} delta file envelope lists...")
     # read compacted input parquet files first
@@ -95,8 +96,8 @@ def _drop_duplicates_by_primary_key_hash(table: pa.Table) -> pa.Table:
     op_type_np = sc.delta_type_column_np(table)
 
     assert len(pk_hash_np) == len(op_type_np), \
-            f"Primary key digest column length ({len(pk_hash_np)}) doesn't " \
-            f"match delta type column length ({len(op_type_np)})."
+        f"Primary key digest column length ({len(pk_hash_np)}) doesn't " \
+        f"match delta type column length ({len(op_type_np)})."
 
     # TODO(raghumdani): move the dedupe to C++ using arrow methods or similar. 
     row_idx = 0
@@ -122,7 +123,6 @@ def _write_new_primary_key_index(
         max_rows_per_index_file: int,
         dedupe_task_index: int,
         deduped_tables: List[Tuple[int, pa.Table]]) -> PyArrowWriteResult:
-
     logger.info(f"[Dedupe task index {dedupe_task_index}] Writing new deduped primary key index: "
                 f"{new_primary_key_index_version_locator}")
 
@@ -149,19 +149,16 @@ def delta_file_locator_to_mat_bucket_index(
     digest = df_locator.digest()
     return int.from_bytes(digest, "big") % materialize_bucket_count
 
-@ray.remote(num_returns=3)
-def dedupe(
-        compaction_artifact_s3_bucket: str,
-        round_completion_info: Optional[RoundCompletionInfo],
-        new_primary_key_index_version_locator: PrimaryKeyIndexVersionLocator,
-        object_ids: List[Any],
-        sort_keys: List[SortKey],
-        max_records_per_index_file: int,
-        num_materialize_buckets: int,
-        dedupe_task_index: int,
-        delete_old_primary_key_index: bool) -> DedupeResult:
 
-    logger.info(f"[Dedupe task {dedupe_task_index}] Starting dedupe task...")
+def _timed_dedupe(compaction_artifact_s3_bucket: str,
+                  round_completion_info: Optional[RoundCompletionInfo],
+                  new_primary_key_index_version_locator: PrimaryKeyIndexVersionLocator,
+                  object_ids: List[Any],
+                  sort_keys: List[SortKey],
+                  max_records_per_index_file: int,
+                  num_materialize_buckets: int,
+                  dedupe_task_index: int,
+                  delete_old_primary_key_index: bool):
     # TODO (pdames): mitigate risk of running out of memory here in cases of
     #  severe skew of primary key updates in deltas
     src_file_records_obj_refs = [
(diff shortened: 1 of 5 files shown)


deltacat/utils/metrics.py
```diff
@@ -0,0 +1,108 @@
...
+    return current_instance_region
+
+
+def _get_current_cluster_name() -> str:
+    with open("/home/ubuntu/ray_bootstrap_config.yaml", "r") as f:
+        cluster_config = yaml.load(f)
+    return cluster_config["cluster_name"]
+
+
+METRICS_DIMENSION_TYPE_TO_VALUE_DICT: Dict[
```
raghumdani, line 67: Minor: ideally use different modules for constants, if using the same module place constants above methods, and methods above classes.

deltacat/utils/metrics.py
```diff
@@ -0,0 +1,108 @@
...
+    return {"Name": f"ray_worker_id",
+            "Value": f"{ray.get_runtime_context().worker.core_worker.get_worker_id()}"
+            }
+
+
+def _get_current_instance_region() -> str:
+    current_instance_region = urllib.request.urlopen(
+        "http://169.254.169.254/latest/meta-data/placement/region"
+    ).read().decode("utf-8")
+    return current_instance_region
```
raghumdani, line 58: This logic is specific to EC2 which won't work on Glue. DeltaCAT should be platform agnostic (e.g., Azure, AWS etc.).

deltacat/utils/metrics.py
```diff
@@ -0,0 +1,108 @@
...
+    return metrics_name_with_type
+
+
+def _get_current_node_ip() -> str:
+    current_node_ip = urllib.request.urlopen(
+        "http://169.254.169.254/latest/meta-data/local-ipv4"
+    ).read().decode("utf-8")
+    return {"Name": f"node_ip",
+            "Value": f"{current_node_ip}"
+            }
```
raghumdani, line 39: Same here above dependency on platform. Can we use a cloudwatch metrics config or environment variables to resolve these? For example, take method `_get_current_node_ip` as argument as `metrics_config` of type 'CloudWatch'?

deltacat/utils/metrics.py
```diff
@@ -0,0 +1,108 @@
...
+    current_instance_region = urllib.request.urlopen(
+        "http://169.254.169.254/latest/meta-data/placement/region"
+    ).read().decode("utf-8")
+    return current_instance_region
+
+
+def _get_current_cluster_name() -> str:
+    with open("/home/ubuntu/ray_bootstrap_config.yaml", "r") as f:
+        cluster_config = yaml.load(f)
+    return cluster_config["cluster_name"]
```
raghumdani, line 64: Again, very specific to EC2. AWS Glue won't have this home directory.

deltacat/utils/metrics.py
```diff
@@ -0,0 +1,108 @@
...
+            "MetricName": f"{metrics_name_with_type}",
+            "Dimensions": dimensions,
+            "Timestamp": timestamp,
+            "Value": value,
+            **kwargs
+        }
+    ]
+
+
+def emit_metrics(metrics_name: str, metrics_type: Enum, value: str, dimension_types: List[Enum], timestamp, **kwargs) -> None:
```
raghumdani, line 91: Minor: define type for timestamp.

deltacat/utils/metrics.py
```diff
@@ -0,0 +1,108 @@
...
+                                             value,
+                                             dimension_types,
+                                             timestamp,
+                                             **kwargs)
+    cluster_name = _get_current_cluster_name()
+    try:
+        response = cloudwatch_client.put_metric_data(
+            Namespace=f"DEFAULT_DELTACAT_METRICS_NAMESPACE_{cluster_name}",
+            MetricData=metrics_data)
+    except ClientError as e:
```
raghumdani, line 106: Not only `ClientError`, do we want to fail compaction if emitting metrics fails? I guess not.

raghumdani (commented):
Overall, looks good. We could make some minor changes to make it more maintainable.

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -271,6 +275,16 @@ def dedupe(
             compaction_artifact_s3_bucket,
             round_completion_info.primary_key_index_version_locator,
         )
+    end_time = time.perf_counter()
+    metrics_val = end_time - start_time
+    ct = datetime.datetime.now()
+    metrics_dimension_type = [MetricsDimensionType.NODE_IP, MetricsDimensionType.RAY_TASK_ID,
+                              MetricsDimensionType.RAY_WORKER_ID]
```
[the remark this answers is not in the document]
Zyiqin-Miranda: For 1, The metrics will be published under the namespace of `cluster_name`, which in my opinion is a better identifier than the table name. Cause we compact same table daily but the cluster name should have an UUID that we can trace back to which job run it is associated with.
For 2, added task id for better visibility when debugging, feels like more info is always helpful. But I could remove it if we agree only `work_id` is needed

deltacat/utils/metrics.py
Zyiqin-Miranda: Not sure I understand the approach here, could you please elaborate more on the part `as metrics_config of type 'CloudWatch'`?

deltacat/utils/metrics.py
Zyiqin-Miranda: Yeah, should not fail the whole compaction workflow. Will catch other all exceptions here.

deltacat/compute/compactor/steps/dedupe.py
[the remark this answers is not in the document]
raghumdani: 1. Makes sense. 
2. I am fine with having it as long as it doesn't cause clutter and slow us in debugging the issue which we will know only when we have a real issue.

deltacat/utils/metrics.py
raghumdani: The idea is you could take an optional argument to `compact_partition` method named `metrics_config` with below signature:

```python
@dataclass
class MetricsConfig:
  def __init__(self, type: str, metrics_dimensions_provider: Callable, cluster_name_provider: Callable):
    self.type = type
    self.retrieve_metrics_dimensions = retrieve_metrics_dimensions
    self.cluster_name_provider = cluster_name_provider

# then in the driver we could,

def metrics_dimensions_callable():
  node_ip = _get_current_node_id()
  instance_region = _get_instance_region()
  # etc etc.

def cluster_name_provider():
  # provide cluster name

config = MetricsConfig(MetricsConfigType.CloudWatch, metrics_dimensions_callable, cluster_name_provider)

compact_partition(.., metrics_config=config, ...)

```

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -271,6 +275,16 @@ def dedupe(
             compaction_artifact_s3_bucket,
             round_completion_info.primary_key_index_version_locator,
         )
+    end_time = time.perf_counter()
```
valiantljk, line 278: May consider `timed_invocation`

deltacat/compute/compactor/steps/dedupe.py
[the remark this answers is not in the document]
Zyiqin-Miranda: Also, For `worker id` and `task_id`, double checked that they are not 1:1 mapping. One `worker id` can map to multiple `task_ids`.

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -271,6 +275,16 @@ def dedupe(
...
+    metrics_val = end_time - start_time
+    ct = datetime.datetime.now()
+    metrics_dimension_type = [MetricsDimensionType.NODE_IP, MetricsDimensionType.RAY_TASK_ID,
+                              MetricsDimensionType.RAY_WORKER_ID]
+    emit_metrics(metrics_name="dedupe",
+                 metrics_type=MetricsType.TIMER,
+                 value=metrics_val,
+                 dimension_types=metrics_dimension_type,
+                 timestamp=ct)
```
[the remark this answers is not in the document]
valiantljk: good idea!

deltacat/utils/metrics.py
valiantljk: correct me if I'm wrong. Cluster name might be same across jobs if we want to speed up cluster provision with cached instances.

deltacat/utils/metrics.py
valiantljk: It looks like a major blocker for now to be platform-agnostic.

deltacat/utils/metrics.py
valiantljk: Do we really need node ip? for debugging purpose, probably node id is sufficient? What we want is probably just an identifier to locate the actual worker logs, e.g., worker id; Please share ur thoughts.

valiantljk:
LGTM. Any thoughts on post-analysis instead of runtime detection? In another words, if we have cloudwatch log internally persisted. We may retrieve the log and run a post-analysis script to figure out the straggler.

deltacat/utils/metrics.py
Zyiqin-Miranda: Here the node id is not included. My original thought is that Ray uses node ip instead of node id  in their dashboard and logs, so using same node ip as identifier could make debugging easier for us?

deltacat/utils/metrics.py
Zyiqin-Miranda: Yeah, if going forward we need to use cached instances, the cluster name will not be a good identifier here. We used to design the cluster name as an unique mapping to each job run. 
At the end of the day, the orchestrator launches the job should know the unique identifier for us to detect which cluster and job instructions this cluster is based on. So how about I add a variable here as `job_run_id` and let orchestrator fill in this environment variable?

Zyiqin-Miranda:
> LGTM. Any thoughts on post-analysis instead of runtime detection? In annother words, if we have cloudwatch log internally persisted. We may retrieve the log and run a post-analysis script to figure out the straggler.

Post analysis sounds like a good idea. We can add another task to derive the script. I think the implementation will be very coupled to how we persist all logs. 
Right now it's streaming all nodes' logs to one CloudWatch log stream, which is quite easier to detect straggler. 
However, if we enable multiple job runs on same cluster, the log structure needs to be redesigned a bit. Also, how Glue persist all logs are quite different from our current approach, feels like we need some careful thought here.

deltacat/utils/metrics.py
Zyiqin-Miranda: > 
Given it a second thought, if we wrap the MetricsConfig in driver, the node id can not be derived correctly because the driver do not know each individual node's ip beforehand?

raghumdani: I am not sure I understood. Let's sync up offline.

Issue #97, opened 2023-03-01
Backfill compaction from source partition only and no previous round completion fails


rkenmi:
Reproduction script:

```python
        # compact all partitions
        for i, expected_delta_pa in enumerate(expected_compacted_deltas_pa):
            # compact each partition twice to ensure idempotence
            prev_delta = None
            for _ in range(2):
                actual_delta = compact_partition(
                    dest_stream.locator,
                    src_partitions[i].locator,
                    src_partition_deltas[i].stream_position,
                    part_col_vals[i],
                    root_compactor_path,
                )
                actual_delta_df = _compacted_delta_to_pandas(
                    actual_delta,
                    deltacat_storage,
                )
                # ensure the compacted delta contents match expectations
                expected_delta_df = expected_delta_pa.to_pandas()
                _assert_df_contents_equal(actual_delta_df, expected_delta_df)
                # ensure no new delta is created when compaction is run again
                if not prev_delta:
                    prev_delta = actual_delta
                else:
                    assert prev_delta == actual_delta, \
                           f"Previous delta does not equal current delta: " \
                           f"{prev_delta} != {actual_delta}"
```

PR #99, opened 2023-03-01
Fix compactor code path with only source partition to compact from and no previous rounds

rkenmi:
Issues: https://github.com/ray-project/deltacat/issues/97


raghumdani (approved):
LGTM!

PR #102, opened 2023-03-10
add property info

valiantljk:
add kwargs in `compact_partition`, allow users to pass in any additional catalog-specific metadata


pfaraone:
The linting step is failing - can you run the pre-commit linters?

raghumdani (approved):
Looks good.

PR #103, opened 2023-03-10
Make memray install optional and support custom logging config

pdames:
This PR (1) resolves an issue where `memray` cannot be installed on unsupported environments (e.g. WIndows, MacOS) by moving the `memray` requirement into a new conditional requirement in `dev-requirements.txt` and (2) adds support for custom application/system logging levels and log file paths read from system environment variables.

Also includes some minor refactoring like a rename of `compacted_partition_locator` to `destination_partition_locator` where appropriate to help resolve ambiguity when multiple compacted partitions may either serve as file sources or destinations during execution of a compaction round.


deltacat/compute/compactor/steps/materialize.py
```diff
@@ -36,6 +36,9 @@
 )
 from deltacat.utils.metrics import emit_timer_metrics, MetricsConfig
 
+if importlib.util.find_spec("memray"):
+    import memray
+
```
raghumdani, line 41: Minor: we may be able to avoid repeating this if statement by creating a separate module. 

```python
# profiling_util.py

import importlib

memray = None
if (importlib.util.find_spec("memray")):
    import memray as mem
    memray = mem
```

and then in materialize or any other file:

```python
from profiling_util import memray
```

dev-requirements.txt
```diff
@@ -2,6 +2,7 @@
 black == 22.12.0
 flake8 == 5.0.4
 isort == 5.10.1
+memray == 1.6.0; platform_system != "Windows" and sys_platform != "darwin" and platform_machine != "arm64"
```
raghumdani, line 5: on `r6g.large` instance when I tested, `sys_platform` was `linux`, and `platform_machine` was `aarch64` (not `arm64`). I assume you tested this on AWS Glue, so you may also want to add?

```suggestion
memray == 1.6.0; platform_system != "Windows" and sys_platform != "darwin" and platform_machine != "arm64" and platform_machine != "aarch64"
```

raghumdani (approved):
Looks good overall. Couple of minor comments.

requirements.txt
```diff
@@ -1,7 +1,6 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
 boto3 == 1.20.24
-memray == 1.6.0
 numpy == 1.21.5
 pandas == 1.3.5
 pyarrow == 10.0.1
```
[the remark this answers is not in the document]
pdames: Good callout. I think I can add ray >= 2.0.0 here to ensure that this both runs on EC2 aarch64 machines and on Glue.

setup.py
```diff
@@ -44,7 +44,6 @@ def find_version(*paths):
         "s3fs == 2022.2.0",
         "tenacity == 8.1.0",
         "typing-extensions == 4.4.0",
-        "memray == 1.6.0",
     ],
```
rkenmi, line 47: It might be nice to also provide this optionally if we are supporting installation through setuptools: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

rkenmi (approved):

Zyiqin-Miranda:
LGTM!

Zyiqin-Miranda (approved):

setup.py
pdames: Thanks for the link. I'm OK keeping memray in `dev-requirements.txt` for simplicity for now. It might also be a bit weird if we wind up making production deploys of DeltaCAT that sometimes do and sometimes don't have profiling enabled based on the underlying architecture they're deployed to so, either way, I think we should continue to separate it out as a non-prod dependency for now (unless we can optionally include profiling for all supported production architectures, and without any undesirable side-effects).

Issue #72, opened 2023-02-07
Hash bucket index data is never materialized from pki if none of the records in delta fall into that bucket


raghumdani:
During the dedupe, we union data from primary key index for an hash bucket index. If the new delta records does not fall into an hash bucket index, even if the records are present in pk index, they are not materialized into a new compacted shard.

PR #105, opened 2023-03-11
Compaction with no primary key index

valiantljk:
Reverts ray-project/deltacat#86, to fix unclear commit message


raghumdani:
#72

PR #106, opened 2023-03-13
Fix key type

valiantljk:
`PartitionLocator` is not hashable when using with dict as key. Revert it to string type. 
Remove un-used function parameters for compaction with no pki

### DIFF
--- a/deltacat/compute/compactor/model/round_completion_info.py
+++ b/deltacat/compute/compactor/model/round_completion_info.py
@@ -6,6 +6,22 @@ from deltacat.compute.compactor.model.pyarrow_write_result import PyArrowWriteRe
 from typing import Any, Dict, Optional, Union
 
 
+class HighWaterMark(dict):
+    """
+    Inherit from dict to make it easy for serialization/deserialization.
+    Keep both partition locator and high watermark as a tuple to be persisted in the rcf
+    """
+
+    def append(self, partition_locator: PartitionLocator, high_watermark: int):
+        self[partition_locator.canonical_string()] = (partition_locator, high_watermark)
+
+    def get(self, partition_locator: PartitionLocator) -> int:
+        val = self.get(partition_locator.canonical_string())
+        if val is not None:
+            return val[1]
+        return 0
+
+
 class RoundCompletionInfo(dict):
     """
     In case of compacting deltas from multi-sources, the high_watermark dict records the high watermark for each source
@@ -15,7 +31,7 @@ class RoundCompletionInfo(dict):
 
     @staticmethod
     def of(
-        high_watermark: Union[Dict[str, int], int],
+        high_watermark: Union[HighWaterMark, int],
         compacted_delta_locator: DeltaLocator,
         compacted_pyarrow_write_result: PyArrowWriteResult,
         sort_keys_bit_width: int,
@@ -31,7 +47,7 @@ class RoundCompletionInfo(dict):
         return rci
 
     @property
-    def high_watermark(self) -> Union[Dict[str, int], int]:
+    def high_watermark(self) -> Union[HighWaterMark, int]:
         return self["highWatermark"]
 
     @property
--- a/deltacat/compute/compactor/model/round_completion_info.py
+++ b/deltacat/compute/compactor/model/round_completion_info.py
@@ -15,7 +15,7 @@ class RoundCompletionInfo(dict):
 
     @staticmethod
     def of(
-        high_watermark: Union[Dict[PartitionLocator, int], int],
+        high_watermark: Union[Dict[str, int], int],
         compacted_delta_locator: DeltaLocator,
         compacted_pyarrow_write_result: PyArrowWriteResult,
         sort_keys_bit_width: int,
@@ -31,7 +31,7 @@ class RoundCompletionInfo(dict):
         return rci
 
     @property
-    def high_watermark(self) -> Union[Dict[PartitionLocator, int], int]:
+    def high_watermark(self) -> Union[Dict[str, int], int]:
         return self["highWatermark"]
 
     @property
--- a/deltacat/compute/compactor/model/round_completion_info.py
+++ b/deltacat/compute/compactor/model/round_completion_info.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 
 from deltacat.storage import DeltaLocator, PartitionLocator
 from deltacat.compute.compactor.model.pyarrow_write_result import PyArrowWriteResult
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Optional
 
 
 class HighWatermark(dict):
@@ -13,7 +13,10 @@ class HighWatermark(dict):
     """
 
     def set(self, partition_locator: PartitionLocator, delta_stream_position: int):
-        self[partition_locator.canonical_string()] = (partition_locator, delta_stream_position)
+        self[partition_locator.canonical_string()] = (
+            partition_locator,
+            delta_stream_position,
+        )
 
     def get(self, partition_locator: PartitionLocator) -> int:
         val = self.get(partition_locator.canonical_string())
@@ -31,7 +34,7 @@ class RoundCompletionInfo(dict):
 
     @staticmethod
     def of(
-        high_watermark: Union[HighWatermark, int],
+        high_watermark: HighWatermark,
         compacted_delta_locator: DeltaLocator,
         compacted_pyarrow_write_result: PyArrowWriteResult,
         sort_keys_bit_width: int,
--- a/deltacat/compute/compactor/model/round_completion_info.py
+++ b/deltacat/compute/compactor/model/round_completion_info.py
@@ -6,14 +6,14 @@ from deltacat.compute.compactor.model.pyarrow_write_result import PyArrowWriteRe
 from typing import Any, Dict, Optional, Union
 
 
-class HighWaterMark(dict):
+class HighWatermark(dict):
     """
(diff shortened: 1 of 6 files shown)


deltacat/compute/compactor/model/round_completion_info.py
```diff
@@ -31,7 +31,7 @@ def of(
         return rci
 
     @property
-    def high_watermark(self) -> Union[Dict[PartitionLocator, int], int]:
```
raghumdani, line 34: can we make PartitionLocator hashable instead?

raghumdani (commented):
Please check if we can make `PartitionLocator` hashable. Ideally, I don't see a reason why it shouldn't be hashable.

Zyiqin-Miranda (commented):
Do we also need to change line [419](https://github.com/ray-project/deltacat/blob/338ccf04fa540f869bd9a8ed50703cbaff20f279/deltacat/compute/compactor/compaction_session.py#L419) since `last_stream_position_compacted` returned by `limit_input_deltas` is a dict now?

valiantljk:
> Do we also need to change line [419](https://github.com/ray-project/deltacat/blob/338ccf04fa540f869bd9a8ed50703cbaff20f279/deltacat/compute/compactor/compaction_session.py#L419) since `last_stream_position_compacted` returned by `limit_input_deltas` is a dict now?
should be ok, it's returned as a dict and then written to rcf as dict as well.

deltacat/compute/compactor/model/round_completion_info.py
pdames: Agreed - all locators are intended to be hashable, but looks like we just didn't override `__hash__` previously. 

This is one of the use cases for the `Locator` `digest()` method. All of our models that inherit from locator should actually override `__eq__` and `__hash__` to use the `digest` of the locator.

pdames: On second though, `digest()` wouldn't be ideal because it is (1) expensive due to an underlying reliance on generating a SHA1 hash and (2) only delivers a point-in-time hash of the mutable locator. I think we should capture a longer term improvement to our models to either convert them all to immutable locators (i.e. inherit from `frozendict` instead of `dict`), or at least make them convertable to immutable locators. 

For the short-term, I think we can capture the same intent here of telegraphing the type of key being passed in, and recording additional information related to this key, by making a new `HighWatermark` class that keeps a `Dict[str, Tuple[int, PartitionLocator]]` under the hood where the key is the canonical string and the value is a pair of `(high_watermark_stream_position, partition_locator_for_canonical_string)`.

One way to do this is by giving `HighWatermark` a method like `append(stream_position: int, partition_locator: PartitionLocator)`.

valiantljk: Thanks for the ideas. Just updated this pr; high watermark is wrapped into a class, using its canonical_string as key.

deltacat/compute/compactor/utils/io.py
```diff
@@ -10,15 +10,18 @@
...
 from deltacat import logs
 from deltacat.compute.compactor import DeltaAnnotated
-from collections import defaultdict
 from typing import Dict, List, Optional, Tuple, Union
+from deltacat.compute.compactor import HighWaterMark
+
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
+
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
```
raghumdani, line 19: Minor: Duplicate logger

deltacat/compute/compactor/model/round_completion_info.py
```diff
@@ -31,7 +47,7 @@ def of(
...
+    def high_watermark(self) -> Union[HighWaterMark, int]:
```
raghumdani, line 50: The return doesn't need to be `Union` then.

deltacat/compute/compactor/model/round_completion_info.py
valiantljk: This however, will still call the `digest()`. I think it's fine for now, the max number of calls throughout the compaction is around `O(deltas)`; In case of incremental compaction, it should be just a few calls.

raghumdani (commented):
Looks good. Couple of minor comments.

deltacat/compute/compactor/model/round_completion_info.py
```diff
@@ -6,6 +6,22 @@
 from typing import Any, Dict, Optional, Union
 
 
+class HighWaterMark(dict):
```
pdames, line 9: Minor: Watermark is one word.

```suggestion
class HighWatermark(dict):
```

deltacat/compute/compactor/model/round_completion_info.py
```diff
@@ -15,7 +31,7 @@ class RoundCompletionInfo(dict):
 
     @staticmethod
     def of(
-        high_watermark: Union[Dict[PartitionLocator, int], int],
+        high_watermark: Union[HighWaterMark, int],
```
pdames, line 34:
```suggestion
        high_watermark: Union[HighWatermark, int],
```

deltacat/compute/compactor/model/round_completion_info.py
pdames, line 50:
```suggestion
    def high_watermark(self) -> HighWatermark:
```

deltacat/compute/compactor/utils/io.py
```diff
@@ -10,15 +10,18 @@
 )
 from deltacat import logs
 from deltacat.compute.compactor import DeltaAnnotated
-from collections import defaultdict
 from typing import Dict, List, Optional, Tuple, Union
+from deltacat.compute.compactor import HighWaterMark
```
pdames, line 14:
```suggestion
from deltacat.compute.compactor import HighWatermark
```

deltacat/compute/compactor/utils/io.py
```diff
@@ -10,15 +10,18 @@
...
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
+
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 
 def discover_deltas(
     source_partition_locator: PartitionLocator,
-    high_watermark: Union[dict, int],
+    high_watermark: Union[HighWaterMark, int],
```
pdames, line 24:
```suggestion
    high_watermark: Union[HighWatermark, int],
```

deltacat/compute/compactor/utils/io.py
```diff
@@ -132,7 +135,7 @@ def limit_input_deltas(
     user_hash_bucket_chunk_size: int,
     input_deltas_stats: Dict[int, DeltaStats],
     deltacat_storage=unimplemented_deltacat_storage,
-) -> Tuple[List[DeltaAnnotated], int, dict]:
+) -> Tuple[List[DeltaAnnotated], int, HighWaterMark]:
```
pdames, line 138:
```suggestion
) -> Tuple[List[DeltaAnnotated], int, HighWatermark]:
```

deltacat/compute/compactor/utils/io.py
```diff
@@ -155,8 +158,8 @@ def limit_input_deltas(
     delta_bytes = 0
     delta_bytes_pyarrow = 0
     delta_manifest_entries = 0
-    latest_stream_position = defaultdict(
-        lambda: -1
+    latest_stream_position = (
+        HighWaterMark()
     )  # tracks the latest stream position for each partition locator
```
pdames, line 163:
```suggestion
    # tracks the latest stream position for each partition locator
    high_watermark = HighWatermark() 
```

deltacat/compute/compactor/utils/io.py
```diff
@@ -187,8 +190,9 @@ def limit_input_deltas(
             delta_bytes += entry.meta.content_length
             if not delta_stats:
                 delta_bytes_pyarrow = delta_bytes * PYARROW_INFLATION_MULTIPLIER
-        latest_stream_position[delta.locator.partition_locator] = max(
-            position, latest_stream_position[delta.locator.partition_locator]
+        latest_stream_position.append(
```
pdames, line 193:
```suggestion
        high_watermark.set(
```

deltacat/compute/compactor/model/round_completion_info.py
```diff
@@ -6,6 +6,22 @@
...
 
 
+class HighWaterMark(dict):
+    """
+    Inherit from dict to make it easy for serialization/deserialization.
+    Keep both partition locator and high watermark as a tuple to be persisted in the rcf
+    """
+
+    def append(self, partition_locator: PartitionLocator, high_watermark: int):
+        self[partition_locator.canonical_string()] = (partition_locator, high_watermark)
```
pdames, line 16: Although I recommended `append` initially, looking at this again I think `set` is a better natural fit here alongside `get`. Also, since the class is already named `HighWatermark` it may be more descriptive to call the integer here `delta_stream_position`?

```suggestion
    def set(self, partition_locator: PartitionLocator, delta_stream_position: int):
        self[partition_locator.canonical_string()] = (partition_locator, delta_stream_position)
```

pdames (changes requested):
Nice! Looks good to merge with a few minor changes.

deltacat/compute/compactor/model/round_completion_info.py
valiantljk: I assume other users can still a simple integer as the high watermark

raghumdani (approved):
Thanks for the fix! Please merge after running a full compaction.

pdames (approved):

PR #111, opened 2023-03-22
adding kwargs to all invocations of _discover_deltas

raghumdani:
All invocations of `_discover_deltas` will require the `equivalent_table_types` arg to be passed to disable resolving an appropriate table type. Earlier only one of the `_discover_deltas` had kwargs. Hence, the invocation that did not have this kwargs failed.


pfaraone (approved):
LGTM

PR #112, opened 2023-03-23
fixing high watermark deserialization and TypeError

raghumdani:
There are two fixes in this PR: 

a. TypeError when operator < is used between NoneType objects.
b. Deserialization of `HighWatermark` when reading round completion files.


valiantljk:
LGTM, thx for fix.

valiantljk (approved):

Issue #124, opened 2023-04-10
Support passing kwargs to the underlying reader/writer in pyarrow via compact_partition


raghumdani:
#### Problem Statement

The kwargs to the underlying reader is hard coded right now. To improve customization, we can allow clients to pass additional kwargs to the underlying content readers i.e., `parquet.read_table` etc.

#### Solution

Add new arg `read_kwargs_provider` and `write_kwargs_provider` that returns appropriate kwargs depending on the content type of the file.

PR #125, opened 2023-04-10
support read_kwargs_provider arg to compact_partition

raghumdani:
Motivation #124

### DIFF
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -280,6 +280,7 @@ def _execute_compaction_round(
         uniform_deltas,
         hash_bucket_count,
         last_stream_position_compacted,
+        require_multiple_rounds,
     ) = io.limit_input_deltas(
         input_deltas,
         cluster_resources,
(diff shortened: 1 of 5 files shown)


deltacat/compute/compactor/utils/io.py
```diff
@@ -282,4 +284,4 @@ def limit_input_deltas(
     logger.info(f"Hash bucket count: {hash_bucket_count}")
     logger.info(f"Input uniform delta count: {len(rebatched_da_list)}")
 
-    return rebatched_da_list, hash_bucket_count, high_watermark
+    return rebatched_da_list, hash_bucket_count, high_watermark, require_multiple_rounds
```
rkenmi, line 287: Minor: Could we also update the function return type signature?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -163,7 +168,9 @@ def _execute_compaction_round(
     rebase_source_partition_high_watermark: Optional[int],
     enable_profiler: Optional[bool],
     metrics_config: Optional[MetricsConfig],
-    list_deltas_kwargs=Optional[Dict[str, Any]],
+    list_deltas_kwargs: Optional[Dict[str, Any]],
+    read_kwargs_provider: Optional[ReadKwargsProvider],
```
rkenmi, line 172: Thanks for adding. Have you thought about allowing clients to pass in various ReadKwargsProviders for specific steps (hash bucketing, materialize)?

rkenmi (approved):

valiantljk (approved):

valiantljk:
LGTM, thanks for the update

deltacat/compute/compactor/compaction_session.py
raghumdani: Yeah, thought about that. Assuming, we have to read the pk the same way (enforcing schema) we read during materialize, I kept it as one arg for now. If we find a need we can add if required in future.

PR #126, opened 2023-04-12
Add assertions to fast fail compaction in case of a metadata/execution issue

rkenmi:
This change adds logging for record counts and validates `# of hash bucket records - # of deduped records = # of materialized records`.

### DIFF
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -440,14 +440,18 @@ def _execute_compaction_round(
     mat_results = sorted(mat_results, key=lambda m: m.task_index)
     deltas = [m.delta for m in mat_results]
     merged_delta = Delta.merge_deltas(deltas)
+    record_info_msg = (
+        f"Hash bucket records: {total_hb_record_count},"
+        f" Deduped records: {total_dd_record_count}, "
+        f" Materialized records: {merged_delta.meta.record_count}"
+    )
+    logger.info(record_info_msg)
     assert (
         total_hb_record_count - total_dd_record_count == merged_delta.meta.record_count
     ), (
         f"Number of hash bucket records minus the number of deduped records"
         f" does not match number of materialized records.\n"
-        f" Hash bucket records: {total_hb_record_count},"
-        f" Deduped records: {total_dd_record_count}, "
-        f" Materialized records: {merged_delta.meta.record_count}"
+        f" {record_info_msg}"
     )
     compacted_delta = deltacat_storage.commit_delta(
         merged_delta, properties=kwargs.get("properties", {})
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -10,6 +10,8 @@ from deltacat.compute.compactor import (
     RoundCompletionInfo,
     SortKey,
 )
+from deltacat.compute.compactor.model.dedupe_result import DedupeResult
+from deltacat.compute.compactor.model.hash_bucket_result import HashBucketResult
 from deltacat.compute.stats.models.delta_stats import DeltaStats
 from deltacat.storage import (
     Delta,
@@ -273,6 +275,7 @@ def _execute_compaction_round(
         uniform_deltas,
         hash_bucket_count,
         last_stream_position_compacted,
+        require_multiple_rounds,
     ) = io.limit_input_deltas(
         input_deltas,
         cluster_resources,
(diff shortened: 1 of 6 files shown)


valiantljk (approved):
LGTM, any test done before merge?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -430,6 +440,15 @@ def _execute_compaction_round(
     mat_results = sorted(mat_results, key=lambda m: m.task_index)
     deltas = [m.delta for m in mat_results]
     merged_delta = Delta.merge_deltas(deltas)
+    assert (
+        total_hb_record_count - total_dd_record_count == merged_delta.meta.record_count
+    ), (
```
raghumdani, line 445: Can we also audit this info? It's worth looking at this even when assertion doesn't fail to understand what's going on.

rkenmi: Sure, by auditing are we talking about simply logging here or a dump of this info into some data store?

rkenmi:
> LGTM, any test done before merge?

Compaction was tested on two tables, one with no partition keys and 26TB of data (using 280 worker nodes / 8960 vCPU), and a smaller table with partition keys.

deltacat/compute/compactor/steps/hash_bucket.py
```diff
@@ -203,7 +207,7 @@ def _timed_hash_bucket(
             num_buckets,
```
raghumdani, line 207: can we also ensure the grouping has retained the same number of records? I think instead of the above assertion (where we have change return types from all the methods), this would be more useful. What do you think?

deltacat/compute/compactor/model/dedupe_result.py
```diff
@@ -0,0 +1,8 @@
+from dataclasses import dataclass
+from typing import Dict, Tuple
+
+
+@dataclass
+class DedupeResult:
+    mat_bucket_idx_to_obj_id: Dict[int, Tuple]
+    deduped_record_count: int
```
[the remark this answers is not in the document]
rkenmi: Adding as an enhancement issue #128

deltacat/compute/compactor/compaction_session.py
rkenmi: Deferring this to a separate PR/issue #127

deltacat/compute/compactor/model/hash_bucket_result.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+
+import numpy as np
+
+
+@dataclass
+class HashBucketResult:
+    hash_bucket_group_to_obj_id: np.ndarray
+    hb_record_count: int
```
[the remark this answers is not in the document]
rkenmi: If we're talking about just updating `hb_record_count` to a NumPy primitive int, the improvement here could be negligible. We could also just include it as a part of #128.

deltacat/compute/compactor/steps/hash_bucket.py
rkenmi: The number of records aggregated up to the driver are still used to assert the expected number of materialized records in the parent workflow (after subtracting deduped records), so I think this would be sufficient for now.

deltacat/compute/compactor/model/hash_bucket_result.py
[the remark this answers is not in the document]
raghumdani: Maybe its worth benchmarking. 8000 * 8000 int is nearly 512 MB (considering 8000 hb each with 8000 envelopes), using numpy will allow zero copy transfer without (de)serialization. I am also skeptical about returning `@dataclass` here as it may deteriorate performance. From the Ray documentation, they optimize storing python collections by converting to arrow memory format but I didn't see similar for arbitrary classes. At the end of the day, it may depend on how cloudpickle works. 

Quoting from [this](https://docs.ray.io/en/latest/ray-core/objects/serialization.html#serialization-notes):

> Whenever possible, use numpy arrays or Python collections of numpy arrays for maximum performance.

deltacat/compute/compactor/compaction_session.py
raghumdani: sure. Can we atleast log it irrespective of assertion result? Here, it's getting logged only if the assertion fails.

raghumdani (commented):
Looks good. Please address minor comments.

deltacat/compute/compactor/model/hash_bucket_result.py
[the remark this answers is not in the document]
rkenmi: Yeah it could be worth benchmarking. FWIW though, Ray Datasets itself uses `@dataclass` annotations for `BlockMetadata`: https://github.com/ray-project/ray/blob/releases/2.3.1/python/ray/data/block.py#L211

raghumdani (approved):

PR #129, opened 2023-04-13
Update hash bucket and dedupe result dataclasses to tuple type

rkenmi:
Update the hash bucket and dedupe result from dataclasses to tuple to mitigate some performance issues found during compaction on certain tables.


raghumdani:
Can you update the description on why this change was needed and any performance improvements?

raghumdani (approved):
LGTM

PR #130, opened 2023-04-28
Fix the stream position of the delta

raghumdani:
1. The stream position delta will be the last stream position of the data it compacts to ensure it is appropriately sorted in future rounds.
2. Write the round completion file after partition commit to avoid data loss.


valiantljk:
In case of rebase, what will be the high watermark persisted in rcf?

raghumdani:
> In case of rebase, what will be the high watermark persisted in rcf?

Yes, I do need to fix this. Will update soon

valiantljk (approved):
LGTM

Zyiqin-Miranda (approved):

PR #132, opened 2023-06-07
Repartition

valiantljk:
* Added new compute step 'repartition', in deltacat, to support different repartition strategy, e.g., hash, range, column
* Currently implemented range repartition
* Added new standalone repartition session, such that data layout optimization via repartition can be decoupled from compaction

Use case: Hot/cold partitioning
* To split the delta by date into hot and cold sub-files, use the range repartition API to reshuffle the data
* Round completion file will be generated, so that following-up compaction can pick up and use the latest re-organized delta

Testing
* Tested with one production table, used 100 r5.8xlarge workers, completed the repartition in 160 seconds.


deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+    annotated_delta: DeltaAnnotated,
+    repartition_type: str,
+    repartition_args: dict,
+    destination_partition: Partition,
+    enable_profiler: bool,
+    read_kwargs_provider: Optional[ReadKwargsProvider],
+    deltacat_storage=unimplemented_deltacat_storage,
+) -> RePartitionResult:
+
+    if repartition_type == "range":
```
Zyiqin-Miranda, line 39: Could we make repartition_type as an enum to clearly show what kind of repartition types are supported?

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # create a remote options provider to round-robin tasks across all nodes or allocated bundles
+    logger.info(f"Setting round robin scheduling with node id:{node_resource_keys}")
+    round_robin_opt_provider = functools.partial(
+        round_robin_options_provider,
+        resource_keys=node_resource_keys,
+        pg_config=pg_config.opts if pg_config else None,
+    )
+
+    (deltas, _,) = io.discover_deltas(
```
raghumdani, line 75: Don't we only need deltas from one table given that we are repartitioning the data within that table only?

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+        destination_partition_locator,
+        rebase_source_partition_locator,
+        rebase_source_partition_high_watermark,
+        deltacat_storage,
+        **list_deltas_kwargs,
+    )
+
+    uniform_deltas = []
+    for delta in deltas:
+        # limit the input deltas to fit on this cluster and convert them to
```
raghumdani, line 88: Is this comment correct? We are not limiting input deltas to fit on this cluster.

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+        # annotated deltas of equivalent size for easy parallel distribution
+        uniform_deltas_part = DeltaAnnotated.rebatch(
+            [DeltaAnnotated.of(delta)],
+            200 * 2**20,  # 200 MiB compressed data per task,
```
raghumdani, line 92: Any reason why 200 MB per task? Since each task is using 1 cpu, an r5.8xlarge can provider 8GB per 1 cpu. Moreover, this can be configurable.

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+    repar_tasks_pending = invoke_parallel(
+        items=uniform_deltas,
+        ray_task=repar.repartition,
+        max_parallelism=max_parallelism,
+        options_provider=round_robin_opt_provider,
+        repartition_type=repartition_type,
+        repartition_args=repartition_args,
+        destination_partition=partition,
+        enable_profiler=False,
+        metrics_config=None,
```
raghumdani, line 122: `metrics_config` is always `None`?

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+        bit_width_of_sort_keys,
+        None,
+    )
+    rcf_source_partition_locator = source_partition_locator
+    round_completion_file_s3_url = None
+    round_completion_file_s3_url = rcf.write_round_completion_file(
+        compaction_artifact_s3_bucket,
+        rcf_source_partition_locator,
+        new_round_completion_info,
+    )
```
raghumdani, line 167: This seems to overwrite the round completion file where a round is not a compaction round at all.

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+    new_compacted_delta_locator = DeltaLocator.of(
+        new_compacted_partition_locator,
+        compacted_delta.stream_position,
+    )
+    bit_width_of_sort_keys = SortKey.validate_sort_keys(
+        source_partition_locator,
+        sort_keys,
+        deltacat_storage,
+    )
+    new_round_completion_info = RoundCompletionInfo.of(
```
Zyiqin-Miranda, line 154: Do we need to write round completion info for repartitioning? Will it get confused to be actual compaction is done on this partition and stream position?

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+            total_record_count = 0
+            col_name_int64 = f"{column}_int64"
+            for table in tables:
+                total_record_count += len(table)
+                table_new = table.add_column(
+                    0,
+                    pa.field(col_name_int64, pa.int64()),
+                    pc.cast(table[column], pa.int64()),
+                )
+                for filter_value in partition_ranges:
```
raghumdani, line 66: Is this ignoring records that doesn't belong to any range? That would result into data loss.

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+                    partitioned_tables[partition_ranges.index(filter_value)].append(
+                        table_new.filter(
+                            (pc.field(col_name_int64) <= pc.scalar(filter_value))
+                        )
+                    )
+                    if filter_value == partition_ranges[-1]:
```
raghumdani, line 72: It is randomly distributing the records within each range. If the range is too big, stream position can always be an indicator for grouping files. For example, initial file ordered also must be preserved as it was grouped by stream position earlier.

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+            range_deltas: List[Delta] = []
+            for _, value in partitioned_tables.items():
+                if len(value) > 0:
+                    range_table = pa.concat_tables(value)
+                    if len(range_table) > 0:
+                        range_table_length += len(range_table)
+                        range_delta = deltacat_storage.stage_delta(
+                            range_table,
+                            destination_partition,
+                            content_type=ContentType.PARQUET,
```
raghumdani, line 89: Do we need to hard-code this? `compaction_driver` is generic enough to write any supported content type.

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+    read_kwargs_provider: Optional[ReadKwargsProvider],
+    deltacat_storage=unimplemented_deltacat_storage,
+) -> RePartitionResult:
+
+    if repartition_type == "range":
+        # A delta that is partitioned by range is partitioned in such a way
+        # that each partition contains rows for which the partitioning expression value lies within a given range for a specified column.
+        # For example, if the partitioning column is "date", the partition ranges are "2020-01-01" to "2020-01-10" and "2020-01-11" to "2020-01-20".
+        # The partitioned delta will have two partitions, one for each range.
+        column: str = repartition_args["column"]
```
raghumdani, line 44: What about tables that do not have any such column which can be used for partitioning?

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+                            (pc.field(col_name_int64) <= pc.scalar(filter_value))
+                        )
+                    )
+                    if filter_value == partition_ranges[-1]:
+                        partitioned_tables[partition_ranges.index(filter_value)].append(
+                            table_new.filter(
+                                (pc.field(col_name_int64) > pc.scalar(filter_value))
+                            )
+                        )
+            # TODO(rootliu) set optimal or max number of records per file to defer the performance degradation due to too many small files
```
raghumdani, line 78: This can be configurable and optimality decision can be delegated to caller.

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+                range_table_length == total_record_count
+            ), f"Repartitioned table should have the same number of records {range_table_length} as the original table {total_record_count}"
+            return RePartitionResult(
+                range_deltas=range_deltas,
+            )
+    else:
+        # Other repartition types, e.g., hash, key, list, column, etc are not supported yet
+        raise NotImplementedError(
+            f"Repartition type {repartition_type} is not supported."
+        )
```
raghumdani, line 103: We definitely need a doc on what each repartition means and why they are required.

raghumdani (commented):
On a high level, we have been writing unit tests for all the modules we are developing. Since this is a completely new functionality, can we add unit tests i.e., definition of done?

deltacat/compute/compactor/utils/repartition_session.py
valiantljk: also, repartition is similar to a rebase compaction assuming no new deltas

raghumdani: I am not sure, that idea does not resonate very well with what repartition does. It will not `compact` at all, so it cannot be thought of as a compaction.

deltacat/compute/compactor/utils/repartition_session.py
```diff
@@ -0,0 +1,168 @@
...
+    DeltaLocator,
+    PartitionLocator,
+    interface as unimplemented_deltacat_storage,
+)
+
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
+
+
+def repartition(
+    source_partition_locator: PartitionLocator,
```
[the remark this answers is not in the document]
raghumdani: why not write a different `discover_delta`?

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,132 @@
...
+        column: str = repartition_args["column"]
+        partition_ranges: List = repartition_args["ranges"]
+        task_id = get_current_ray_task_id()
+        worker_id = get_current_ray_worker_id()
+        with memray.Tracker(
+            f"repartition_{worker_id}_{task_id}.bin"
+        ) if enable_profiler else nullcontext():
+            tables: List[pa.Table] = deltacat_storage.download_delta(
+                annotated_delta,
+                storage_type=StorageType.LOCAL,
```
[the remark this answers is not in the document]
raghumdani: Ok, can we add a TODO here then?

deltacat/compute/compactor/utils/repartition_session.py
raghumdani: I get your thought process, but it's currently not run as part of a normal compaction flow.

deltacat/compute/compactor/steps/repartition.py
raghumdani: Yeah, it depends on the ranges that are passed as well. I see you have assert statement at the end which should mitigate this concern. But its better if the caller didn't have to know the min/max values of the columns so that range could represent all the records.

deltacat/compute/compactor/steps/repartition.py
raghumdani: Makes sense. I guess we will end up producing lot of small files as a result. The skipping untouched files also will ignore if the records aren't touched in a new delta. This could defeat the purpose of compaction altogether.

deltacat/compute/compactor/steps/repartition.py
raghumdani: So, we have to schedule repartitioning job separately depending on whether the column is present. That should work, but one caveat is that we may still not be able to utilize hot cold partitioning for all tables.

Zyiqin-Miranda (approved):
LGTM!

deltacat/compute/compactor/steps/repartition.py
Zyiqin-Miranda: I think that's why we need a compaction right after hot/cold repartitioning. Initial rebase compaction should normalize the file size for us

deltacat/compute/compactor/steps/repartition.py
valiantljk: correct

deltacat/compute/compactor/utils/repartition_session.py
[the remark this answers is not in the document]
valiantljk: or refactor the existing discover_delta. Will mark it as a TODO

deltacat/compute/compactor/utils/repartition_session.py
valiantljk: It does 'compact', right? it compact cold data together and hot data together; In rebase, it compact small files into larger files, by setting records per files;

And with further optimization in next release, repartition can be a better version of rebase, as we can figure out how to compact cold data together into less number of cold files.

deltacat/compute/compactor/steps/repartition.py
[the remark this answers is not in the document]
valiantljk: https://github.com/ray-project/deltacat/issues/133

deltacat/compute/compactor/utils/repartition_session.py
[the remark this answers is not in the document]
valiantljk: https://github.com/ray-project/deltacat/issues/133

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,162 @@
...
+                total_record_count += len(table)
+                table_new = table.add_column(
+                    0,
+                    pa.field(col_name_int64, pa.int64()),
+                    pc.cast(table[column], pa.int64()),
+                )
+                # handle the partition for values less than or equal to the smallest value
+                partitioned_tables_list[0].append(
+                    table_new.filter(
+                        pc.field(col_name_int64) <= pc.scalar(partition_ranges[0])
```
raghumdani, line 85: This approach iterates over the array M times where M is number of buckets. Isn't it faster to sort the array (in N logN) and iterate through it once to split the array using pyarrow [slice](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.slice)?

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,162 @@
...
+            )
+    else:
+        raise NotImplementedError(
+            f"Repartition type {repartition_type} is not supported."
+        )
+
+
+@ray.remote
+def repartition(
+    annotated_delta: DeltaAnnotated,
```
raghumdani, line 136: Does driver now create annotated delta? Why did you remove the `repartition_session`?

deltacat/compute/compactor/steps/repartition.py
raghumdani: I think this needs to be part of the first release itself. We won't get enough time to work on it after 6/30 as this approach would end up creating lot of small files defeating the purpose of compaction.

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,162 @@
...
+                            (pc.field(col_name_int64) > pc.scalar(lower_limit))
+                            & (pc.field(col_name_int64) <= pc.scalar(upper_limit))
+                        )
+                    )
+                # handle the partition for values greater than the largest value
+                partitioned_tables_list[-1].append(
+                    table_new.filter(
+                        pc.field(col_name_int64) > pc.scalar(partition_ranges[-1])
+                    )
+                )
```
raghumdani, line 104: Can these statements be avoided by add an -Inf and +Inf in the list of partition ranges?

deltacat/compute/compactor/utils/repartition_session.py
raghumdani: The Iceberg domain word is [copy-on-write](https://aws.amazon.com/blogs/big-data/use-apache-iceberg-in-a-data-lake-to-support-incremental-data-processing/#Copy-on-write%20vs.%20read-on-merge), where we duplicate the files that were updated upon writes. One critical requirement is that it works on new logs while re-partitioning works on current state of data in the table though I somewhat agree goals are same.

deltacat/compute/compactor/steps/repartition.py
raghumdani: n cold files and n hot files in case 1 seems incorrect. You are basically passing `max_records_per_entry` as None which will only create 1 file per range. So we may end up with only two files at the end. Can you confirm?

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,162 @@
...
+            for partition_tables in partitioned_tables_list:
+                if len(partition_tables) > 0:
+                    partition_table = pa.concat_tables(partition_tables)
+                    if len(partition_table) > 0:
+                        partition_table_length += len(partition_table)
+                        partition_delta = deltacat_storage.stage_delta(
+                            partition_table,
+                            destination_partition,
+                            content_type=repartitioned_file_content_type,
+                        )
```
raghumdani, line 119: The `max_records_per_entry` arg is being passed as None. Can you check if this is creating only 1 file per range?

deltacat/compute/compactor/model/repartition_result.py
```diff
@@ -0,0 +1,6 @@
+from typing import NamedTuple, List
```
raghumdani, line 1: If you agree this is a separate compute job in DeltaCAT, should we create a separate module e.g., `compute/repartition`?

raghumdani (commented):
Mostly looks good. Please check high level comments.

deltacat/compute/compactor/steps/repartition.py
valiantljk: you are right that one file per range, so in our case, given one date, e.g., ranges=[2023-01-01], each file will be split into two files, i.e., records <= 2023-01-01, and records>2023-01-01, so it'll be 2 files per original file

deltacat/compute/compactor/steps/repartition.py
valiantljk: internally, I think all tables have this column; for open source and other catalogs, we may recommend customers to append such column

deltacat/compute/compactor/steps/repartition.py
```diff
@@ -0,0 +1,162 @@
...
+            tables: List[pa.Table] = deltacat_storage.download_delta(
+                annotated_delta,
+                storage_type=StorageType.LOCAL,
+                file_reader_kwargs_provider=read_kwargs_provider,
+            )
+            # check if the column exists in the table
+            if not all(column in table.column_names for table in tables):
+                raise ValueError(f"Column {column} does not exist in the table")
+            # given a range [x, y, z], we need to split the table into 4 files, i.e., (-inf, x], (x, y], (y, z], (z, inf)
+            partitioned_tables_list = [[] for _ in range(len(partition_ranges) + 1)]
```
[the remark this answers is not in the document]
valiantljk: added partition_ranges.sort()

deltacat/compute/compactor/steps/repartition.py
valiantljk: This actually solved the small file size issue. thx

deltacat/compute/compactor/model/repartition_result.py
valiantljk: good idea. let's stay with the current structure as it's still high customized to the compaction use case. In future, especially on glue, we may expand to more generic API.

valiantljk:
> On a high level, we have been writing unit tests for all the modules we are developing. Since this is a completely new functionality, can we add unit tests i.e., definition of done?

https://github.com/ray-project/deltacat/issues/134

deltacat/compute/compactor/repartition_session.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # create a remote options provider to round-robin tasks across all nodes or allocated bundles
+    logger.info(f"Setting round robin scheduling with node id:{node_resource_keys}")
+    round_robin_opt_provider = functools.partial(
+        round_robin_options_provider,
+        resource_keys=node_resource_keys,
+        pg_config=pg_config.opts if pg_config else None,
+    )
+
+    (deltas, _,) = io.discover_deltas(
```
raghumdani, line 84: I see you have created a TODO (#133 ) to refactor `discover_deltas`, can we add a comment in the code as well?

raghumdani (commented):
Overall looks good. Please address minor comments.

deltacat/compute/compactor/repartition_session.py
```diff
@@ -0,0 +1,176 @@
...
+def repartition(
+    source_partition_locator: PartitionLocator,
+    destination_partition_locator: PartitionLocator,
+    repartition_args: Any,
+    compaction_artifact_s3_bucket: str,
+    last_stream_position_to_compact: int,
+    repartition_type: RepartitionType = RepartitionType.RANGE,
+    rebase_source_partition_locator: Optional[PartitionLocator] = None,
+    rebase_source_partition_high_watermark: Optional[int] = None,
+    sort_keys: List[SortKey] = None,
```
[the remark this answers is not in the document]
valiantljk: yes: https://github.com/ray-project/deltacat/issues/133

deltacat/storage/model/types.py
```diff
@@ -1,15 +1,23 @@
...
 
 LocalTable = Union[pa.Table, pd.DataFrame, np.ndarray]
 LocalDataset = List[LocalTable]
-DistributedDataset = Dataset[Union[ArrowRow, np.ndarray, Any]]
+# Starting Ray 2.5.0, Dataset follows a strict mode (https://docs.ray.io/en/latest/data/faq.html#migrating-to-strict-mode),
+# and generic annotation is removed. So add a version checker to determine whether to use the old or new definition.
+ray_version = pkg_resources.parse_version(pkg_resources.get_distribution("ray").version)
+change_version = pkg_resources.parse_version("2.5.0")
+if ray_version < change_version:
+    DistributedDataset = Dataset[Union[ArrowRow, np.ndarray, Any]]
```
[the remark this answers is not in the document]
valiantljk: Yeah, according to Patrick, ` it was primarily used for IDE and user hints about some concrete types expected be contained in the Dataset at the time it was introduced.` This is no difference with Dataset[Any], just explicit type expectation for dev. 

Also, I think we may completely remove this version check and replace DistributedDataset with Dataset later, i.e.,
```
DistributedDataset= Dataset
```
 if we find out it doesn't break anything backwards. Our current codepath in production doesn't use the DistributedDataset.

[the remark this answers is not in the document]
raghumdani: Then can we use just Dataset instead?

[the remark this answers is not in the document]
valiantljk: it looks like we can, but we want to keep it until we fully test the backward compatibility, which is out of the scope of this pr.

deltacat/compute/compactor/model/repartition_result.py
valiantljk: added a TODO

deltacat/compute/compactor/steps/repartition.py
valiantljk: added a TODO

deltacat/compute/compactor/repartition_session.py
```diff
@@ -0,0 +1,176 @@
...
+    repar_tasks_pending = invoke_parallel(
+        items=uniform_deltas,
+        ray_task=repar.repartition,
+        max_parallelism=max_parallelism,
+        options_provider=round_robin_opt_provider,
+        repartition_type=repartition_type,
+        repartition_args=repartition_args,
+        max_records_per_output_file=records_per_repartitioned_file,
+        destination_partition=partition,
+        enable_profiler=False,
```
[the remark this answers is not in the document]
valiantljk: fixed

deltacat/storage/model/types.py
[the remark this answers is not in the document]
valiantljk: https://github.com/ray-project/deltacat/issues/135

deltacat/compute/compactor/repartition_session.py
```diff
@@ -0,0 +1,178 @@
...
+        None,
+        bit_width_of_sort_keys,
+        None,
+    )
+    rcf_source_partition_locator = source_partition_locator
+    round_completion_file_s3_url = None
+    round_completion_file_s3_url = rcf.write_round_completion_file(
+        compaction_artifact_s3_bucket,
+        rcf_source_partition_locator,
+        new_round_completion_info,
```
raghumdani, line 176: Lastly, you might want to run incremental after repartition using the same round completion file to make sure it works correctly.

raghumdani (approved):
LGTM!

PR #137, opened 2023-06-13
[skip untouched files]Enable skipping untouched files during materialize

Zyiqin-Miranda:
This PR adds support for:
1. Skipping untouched files during materialize. 
All manifest entries that do not contain records to be updated are considered untouched. Those manifest entries are copied by reference. 
2. Expose the ratio of untouched file count divided by total manifest entry files count as a logger line for now.
Will follow up with PR to emit the ratio as a metrics.

### DIFF
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -445,6 +445,25 @@ def _execute_compaction_round(
     )
     logger.info(f"Getting {len(mat_tasks_pending)} materialize result(s)...")
     mat_results = ray.get(mat_tasks_pending)
+    total_count_of_src_dfl_not_touched = sum(
+        m.count_of_src_dfl_not_touched for m in mat_results
+    )
+    total_length_src_dfl = sum(m.count_of_src_dfl for m in mat_results)
+    logger.info(
+        f"Got total of {total_count_of_src_dfl_not_touched} manifest files not touched."
+    )
+    logger.info(
+        f"Got total of {total_length_src_dfl} manifest files during compaction."
+    )
+    manifest_entry_copied_by_reference_ratio = (
+        (round(total_count_of_src_dfl_not_touched / total_length_src_dfl, 4) * 100)
+        if total_length_src_dfl != 0
+        else None
+    )
+    logger.info(
+        f"{manifest_entry_copied_by_reference_ratio} percent of manifest files are copied by reference during materialize."
+    )
+
     logger.info(f"Got {len(mat_results)} materialize result(s).")
 
     mat_results = sorted(mat_results, key=lambda m: m.task_index)
@@ -491,13 +510,13 @@ def _execute_compaction_round(
         PyArrowWriteResult.union([m.pyarrow_write_result for m in mat_results]),
         bit_width_of_sort_keys,
         last_rebase_source_partition_locator,
+        manifest_entry_copied_by_reference_ratio,
     )
     rcf_source_partition_locator = (
         rebase_source_partition_locator
         if rebase_source_partition_locator
         else source_partition_locator
     )
-
     logger.info(
         f"partition-{source_partition_locator.partition_values},"
         f"compacted at: {last_stream_position_compacted},"
--- a/deltacat/compute/compactor/steps/materialize.py
+++ b/deltacat/compute/compactor/steps/materialize.py
@@ -1,10 +1,11 @@
 import importlib
 import logging
 import time
+from uuid import uuid4
 from collections import defaultdict
 from contextlib import nullcontext
 from itertools import chain, repeat
-from typing import List, Optional, Tuple, Dict, Any
+from typing import List, Optional, Tuple, Dict, Any, Union
 import pyarrow as pa
 import ray
 from ray import cloudpickle
@@ -18,7 +19,18 @@ from deltacat.compute.compactor.steps.dedupe import (
     DedupeTaskIndexWithObjectId,
     DeltaFileLocatorToRecords,
 )
-from deltacat.storage import Delta, DeltaLocator, Partition, PartitionLocator
+from deltacat.storage import (
+    Delta,
+    DeltaLocator,
+    DeltaType,
+    Partition,
+    PartitionLocator,
+    Manifest,
+    ManifestEntry,
+    LocalDataset,
+    LocalTable,
+    DistributedDataset,
+)
 from deltacat.storage import interface as unimplemented_deltacat_storage
 from deltacat.utils.common import ReadKwargsProvider
 from deltacat.types.media import DELIMITED_TEXT_CONTENT_TYPES, ContentType
@@ -56,12 +68,44 @@ def materialize(
     read_kwargs_provider: Optional[ReadKwargsProvider] = None,
     s3_table_writer_kwargs: Optional[Dict[str, Any]] = None,
     deltacat_storage=unimplemented_deltacat_storage,
-) -> MaterializeResult:
+):
+    def _stage_delta_implementation(
+        data: Union[LocalTable, LocalDataset, DistributedDataset, Manifest],
+        partition: Partition,
+        stage_delta_from_existing_manifest: Optional[bool],
+    ) -> Delta:
+        if stage_delta_from_existing_manifest:
+            delta = Delta.of(
+                locator=DeltaLocator.of(partition.locator),
+                delta_type=DeltaType.UPSERT,
+                meta=manifest.meta,
+                manifest=data,
+                previous_stream_position=partition.stream_position,
+                properties={},
+            )
+            return delta
+
+    def _stage_delta_from_manifest_entry_reference_list(
+        manifest_entry_list_reference: List[ManifestEntry],
+        partition: Partition,
+        delta_type: DeltaType = DeltaType.UPSERT,
+    ) -> Delta:
+        assert (
+            delta_type == DeltaType.UPSERT
+        ), "Stage delta with existing manifest entries only supports UPSERT delta type!"
+        manifest = Manifest.of(entries=manifest_entry_list_reference, uuid=str(uuid4()))
+        delta = _stage_delta_implementation(
+            data=manifest,
+            partition=partition,
+            delta_type=delta_type,
+            stage_delta_from_existing_manifest=True,
+        )
+        return delta
+
     # TODO (rkenmi): Add docstrings for the steps in the compaction workflow
     #  https://github.com/ray-project/deltacat/issues/79
     def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
         compacted_table = pa.concat_tables(compacted_tables)
-
         if compacted_file_content_type in DELIMITED_TEXT_CONTENT_TYPES:
             # TODO (rkenmi): Investigate if we still need to convert this table to pandas DataFrame
             # TODO (pdames): compare performance to pandas-native materialize path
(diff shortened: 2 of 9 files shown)


deltacat/compute/compactor/compaction_session.py
```diff
@@ -445,6 +445,25 @@ def _execute_compaction_round(
...
+        f"Got total of {total_count_of_src_dfl_not_touched} manifest files not touched."
+    )
+    logger.info(
+        f"Got total of {total_length_src_dfl} manifest files during compaction."
+    )
+    untouched_manifest_file_percentage = (
+        round(total_count_of_src_dfl_not_touched / total_length_src_dfl, 3) * 100
+    )
+    logger.info(
+        f"{untouched_manifest_file_percentage} percent of manifest files are not touched during materialize."
```
valiantljk, line 464: Can we persist the skipping file ratio to the round completion file? We thought about using ddb, but it seems unless we return through compaction_session to compaction driver, or we figure out how to retrieve from cloudwatch in compaction driver, the easiest way might be through the rcf.

valiantljk (commented):
Overall, LGTM, just a minor comment on how to persist the ratio, so we can easily reuse it in subsequent runs.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -56,12 +67,25 @@ def materialize(
     read_kwargs_provider: Optional[ReadKwargsProvider] = None,
     s3_table_writer_kwargs: Optional[Dict[str, Any]] = None,
     deltacat_storage=unimplemented_deltacat_storage,
-) -> MaterializeResult:
+):
+    def _construct_delta_from_manifest_entry_reference_list(
```
raghumdani, line 71: I see this code is similar to `stage_delta` implementation of the storage interface. Would it be make sense to make this part of `stage_delta` API itself? i.e., staging delta using existing manifests? 

Moreover, some catalogs may not support referencing the existing files during compaction. Hence, this must be left to the storage layer implementation. For example, can Iceberg or Delta Lake support referencing existing S3 files?

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -56,12 +67,25 @@ def materialize(
...
+        manifest_entry_list_reference: List[ManifestEntry],
+    ) -> Delta:
+        manifest = Manifest.of(entries=manifest_entry_list_reference, uuid=str(uuid4()))
+        delta = Delta.of(
+            locator=DeltaLocator.of(partition.locator, None),
```
raghumdani, line 76: Do we have to specify `None` explicitly when arg is optional? I think default is missing in the signature which may be forcing us to pass it.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
                 f" to download delta locator {delta_locator} with entry ID {src_file_idx_np.item()}"
                 f" is: {download_delta_manifest_entry_time}s"
             )
-            mask_pylist = list(repeat(False, len(pa_table)))
+            record_numbers_tpl_copy = copy.deepcopy(record_numbers_tpl)
```
raghumdani, line 217: Can we not create a copy? It would use up double the memory than what we had estimated for.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
+                    f"Untouched manifest file found, "
+                    f"record numbers length: {record_numbers_length} "
+                    f"same as downloaded table length: {len(pa_table)}"
+                )
+                count_of_src_dfl_not_touched += 1
+                untouched_src_manifest_entry = manifest.entries[src_file_idx_np.item()]
+                manifest_entry_list_reference.append(untouched_src_manifest_entry)
+            else:
+                # TODO(raghumdani): reference the same file URIs while writing the files
+                # instead of copying the data over and creating new files.
```
raghumdani, line 235: You can remove this TODO now.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
             )
-            mask_pylist = list(repeat(False, len(pa_table)))
+            record_numbers_tpl_copy = copy.deepcopy(record_numbers_tpl)
+            record_numbers_length_list = chain.from_iterable(record_numbers_tpl_copy)
+            record_numbers_length = sum(1 for item in record_numbers_length_list)
             record_numbers = chain.from_iterable(record_numbers_tpl)
-            # TODO(raghumdani): reference the same file URIs while writing the files
-            # instead of copying the data over and creating new files.
-            for record_number in record_numbers:
-                mask_pylist[record_number] = True
```
raghumdani, line 193: I think this is where you could figure out the length.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
-            mask = pa.array(mask_pylist)
-            pa_table = pa_table.filter(mask)
-            record_batch_tables.append(pa_table)
-            if record_batch_tables.has_batches():
-                batched_tables = record_batch_tables.evict()
-                materialized_results.append(_materialize(batched_tables))
+            if (
+                record_numbers_length == len(pa_table)
+                and src_file_partition_locator != source_partition_locator
```
raghumdani, line 223: I would instead ensure that `src_file_partition_locator` is always equal to `round_completion_info.compacted_delta_locator.partition_locator` as it is by design.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
+                    manifest_entry_list_reference
+                )
+            )
+            logger.info(
+                f"Got delta with {len(referenced_manifest_delta.manifest.entries)} referenced manifest entries"
+            )
+
         merged_delta = Delta.merge_deltas([mr.delta for mr in materialized_results])
+        if referenced_manifest_delta:
+            merged_delta = Delta.merge_deltas([merged_delta, referenced_manifest_delta])
```
raghumdani, line 264: This block can be simplified in terms of readability.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
+            if (
+                record_numbers_length == len(pa_table)
+                and src_file_partition_locator != source_partition_locator
+            ):
+                logger.info(
+                    f"Untouched manifest file found, "
+                    f"record numbers length: {record_numbers_length} "
+                    f"same as downloaded table length: {len(pa_table)}"
+                )
+                count_of_src_dfl_not_touched += 1
```
raghumdani, line 230: Do you need to separately keep track of this variable? Wouldn't it be the `len(manifest_entry_list_reference)`? Ideally, we should avoid redundant variables. They cause clutter and impact readability.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,23 +214,55 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
-                materialized_results.append(_materialize(batched_tables))
+            if (
+                record_numbers_length == len(pa_table)
+                and src_file_partition_locator != source_partition_locator
+            ):
+                logger.info(
+                    f"Untouched manifest file found, "
+                    f"record numbers length: {record_numbers_length} "
+                    f"same as downloaded table length: {len(pa_table)}"
+                )
```
raghumdani, line 229: We may be unnecessarily logging each file as there can be thousands of files in a single delta. However, it can be a debug log if you think it will be useful. Also, consider how it can impact performance.

raghumdani (commented):
Mostly looks good. Please check high level comments and few minor comments.

deltacat/compute/compactor/steps/materialize.py
Zyiqin-Miranda: Yeah, I was thinking about moving `stage_delta` to implementation of the storage interface as well. Keep it here in initial version to show the implementation. Will move it to internal implementation package

deltacat/compute/compactor/steps/materialize.py
Zyiqin-Miranda: My understanding is that `src_file_partition_locator` will equal to `source_partition_locator` if this file is coming from new deltas from source table?

raghumdani: Yes, unless `src_partition_locator` is equal to destination, we shouldn't be caching deltas as they are completely different streams.

deltacat/compute/compactor/steps/materialize.py
raghumdani: I see it as tech debt and not iterative changes. The latter approach would work for all catalogs while the former is specific to Amazon internal catalog.

valiantljk:
LGTM

valiantljk (approved):

valiantljk:
Please create issues for todos, and linked in the pr.

Zyiqin-Miranda:
Related issue tracking TODO for integration tests: https://github.com/ray-project/deltacat/issues/144

Zyiqin-Miranda:
Related issue tracking moving stage delta implementation to internal: https://github.com/ray-project/deltacat/issues/145

deltacat/compute/compactor/steps/materialize.py
raghumdani: Please remove this TODO.

deltacat/compute/compactor/steps/materialize.py
Zyiqin-Miranda: Could you explain more on what you mean by src_file_partition_locator always equal to round_completion_info.compacted_delta_locator.partition_locator? There will be two types of delta: one from previous compacted table, one from new uncompacted table. The partition locator is different here right?

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -185,39 +234,80 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
                 mask_pylist[record_number] = True
-            mask = pa.array(mask_pylist)
-            pa_table = pa_table.filter(mask)
-            record_batch_tables.append(pa_table)
-            if record_batch_tables.has_batches():
-                batched_tables = record_batch_tables.evict()
-                materialized_results.append(_materialize(batched_tables))
+            if (
+                record_numbers_length == len(pa_table)
+                and src_file_partition_locator != source_partition_locator
```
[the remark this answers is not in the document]
Zyiqin-Miranda: Changed. And removed TODO.

deltacat/compute/compactor/steps/materialize.py
raghumdani: TODO #145 created.

raghumdani (approved):
Looks good to me.

Zyiqin-Miranda:
@raghumdani, any more comment on this PR?

raghumdani:
> @raghumdani, any more comment on this PR?

Looks good to me! Thanks

Issue #138, opened 2023-06-16
Set appropriate hash bucket count and distribute deltas uniformly across cluster


raghumdani:
This change will ensure that optimal hash bucket count is set according to the size of cluster based on POC results. Using too small hash bucket count leads to OOM and too much leads to latency increase due to shuffling involved. Hence, an ideal hash bucket count is needed to ensure the compaction runs optimally as well as reliably.

PR #139, opened 2023-06-16
Setting hash bucket count based on POC results

raghumdani:
[Description] This change ensures the hash bucket count is optimal, which is derived based on the earlier POC results. A ratio between pyarrow memory bytes and the optimal hash bucket count is introduced.

[Motivation] #138

#### Test Results

```
test_sanity (tests.compactor.utils.test_io.TestFitInputDeltas) ... ok
test_when_cpu_resources_is_not_passed (tests.compactor.utils.test_io.TestFitInputDeltas) ... ok
test_when_hash_bucket_count_overridden (tests.compactor.utils.test_io.TestFitInputDeltas) ... ok
test_when_no_input_deltas (tests.compactor.utils.test_io.TestFitInputDeltas) ... ok
test_when_not_enough_memory_splits_manifest_entries (tests.compactor.utils.test_io.TestFitInputDeltas) ... ok
```


deltacat/compute/compactor/utils/io.py
```diff
@@ -285,3 +241,125 @@ def limit_input_deltas(
...
+        total_files += len(manifest_entries)
+
+        high_watermark.set(
+            delta.locator.partition_locator,
+            max(position, high_watermark.get(delta.locator.partition_locator)),
+        )
+        delta_annotated = DeltaAnnotated.of(delta)
+        annotated_input_da_list.append(delta_annotated)
+
+    def estimate_size(content_length):
```
valiantljk, line 296: can we add some explanation on why we estimate the size like this?

valiantljk (approved):
LGTM

deltacat/compute/compactor/utils/io.py
```diff
@@ -285,3 +241,125 @@ def limit_input_deltas(
...
+
+    # Assuming each CPU consumes equal amount of memory
+    min_delta_bytes = total_memory / worker_cpus
+    rebatched_da_list = DeltaAnnotated.rebatch(
+        annotated_deltas=annotated_input_da_list,
+        min_delta_bytes=min_delta_bytes,
+        estimation_function=estimate_size,
+    )
+
+    # Recommended hash buckets based on the POC results
```
rkenmi, line 307: "POC results" seems a bit vague to an external user, maybe we should clarify what types of benchmarks were used to derive the memory to hash bucket count ratio?

rkenmi (approved):

Issue #141, opened 2023-06-20
Log the memory consumption to calculate correctness of worker estimation


raghumdani:
This change will allow us to validate the correctness of the workers we estimated to run a compaction job reliably.

raghumdani:
The `totalMemoryGBSeconds` and `usedMemoryGBSeconds` are added to audit for on-demand scaling based compactor. 

This is addressed as part of epic: #150

PR #142, opened 2023-06-20
Logging memory consumed to validate worker estimation correctness

raghumdani:
[Description] This change will allow us to validate the correctness of the workers we estimated to run a compaction job reliably.


raghumdani:
This will resolve #141

deltacat/utils/profiling.py
```diff
@@ -0,0 +1,39 @@
+import ray
+from dataclasses import dataclass
+
+
+@dataclass
+class ClusterResourcesUsage:
```
pdames, line 6:
```suggestion
class ClusterResources:
```

Maybe this can be more concisely stated as `ClusterUtilization` here?

Also, if we change the API to take dictionaries of `cluster_resources` and `available_resources` from Ray, then this class becomes both more complete and maintainable IMO if we just make a 3rd dictionary of `used_resources` which takes the difference of all values with the same key between `cluster_resources` and `available_resources`. The values held in these dictionaries are already well-documented, so I think adding individual members for each of them isn't totally necessary, but creating dataclass members for commonly used values like you've done here could still at least help with IDE autocompletion.

deltacat/utils/profiling.py
```diff
@@ -0,0 +1,39 @@
...
+def get_current_cluster_resources_usage():
+
+    cluster_resources = ray.cluster_resources()
+    available_resources = ray.available_resources()
+
+    return ClusterResourcesUsage(
+        total_cpu=cluster_resources["CPU"],
+        used_cpu=(cluster_resources["CPU"] - available_resources["CPU"]),
+        total_memory_bytes=cluster_resources["memory"],
+        used_memory_bytes=(cluster_resources["memory"] - available_resources["memory"]),
```
pdames, line 33: This puts the responsibility on every client constructing the class to compute used resources appropriately. Seems like we should just take the dictionaries returned from `ray.cluster_resources()` and `ray.available_resources()` as input then populate all computed "used resources" values internally.

deltacat/utils/profiling.py
```diff
@@ -0,0 +1,39 @@
...
+    ):
+        self.total_memory_bytes = total_memory_bytes
+        self.used_memory_bytes = used_memory_bytes
+        self.total_cpu = total_cpu
+        self.used_cpu = used_cpu
+        self.total_object_store_memory_bytes = total_object_store_memory_bytes
+        self.used_object_store_memory_bytes = used_object_store_memory_bytes
+
+
+def get_current_cluster_resources_usage():
```
pdames, line 24:
```suggestion
def get_cluster_utilization():
```

Same comment as above for class naming. Also, if the primary purpose of this function is to construct the corresponding class, then should we just make it a static factory method attached to the `ClusterUtilization` class?

deltacat/utils/profiling.py
```diff
@@ -0,0 +1,39 @@
+import ray
```
pdames, line 1: Since we don't have anything related to profiling in this module yet, maybe we should just rename it to something more specific like `resources.py`?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -334,6 +335,14 @@ def _execute_compaction_round(
             "Multiple rounds are not supported. Please increase the cluster size and run again."
         )
 
+    cluster_resources_usage = get_current_cluster_resources_usage()
```
pdames, line 338: For this common case, may want to consider another utility function that wraps `invoke_parallel`, includes the call to `ray.get()`, and has an argument like `log_utilization_diff` to add before/after log-lines that print out all dictionaries inside of the `ClusterUtilization` class before and after any requested parallel invocation completes.

pdames (changes requested):

deltacat/utils/profiling.py
raghumdani: Yeah, that makes sense.

deltacat/compute/compactor/compaction_session.py
raghumdani: created `log_current_cluster_utilization` method.

deltacat/utils/resources.py
```diff
@@ -0,0 +1,51 @@
...
+        cluster_resources = ray.cluster_resources()
+        available_resources = ray.available_resources()
+
+        return ClusterUtilization(
+            cluster_resources=cluster_resources, available_resources=available_resources
+        )
+
+
+def log_current_cluster_utlization(log_identifier: str):
+    cluster_utilization = ClusterUtilization.get_current_cluster_utilization()
```
pdames, line 45: For sanity and insights into long-term trends, I'd also like to keep durable tabs on how much time this is taking to collect across successive versions of Ray and DeltaCAT:
```
cluster_utilization, latency = timed_invocation(ClusterUtilization.get_current_cluster_utilization)
logger.info(f"Cluster Utilization Collection Latency: {latency}s")
```

pdames (changes requested):

deltacat/utils/resources.py
```diff
@@ -0,0 +1,51 @@
...
+    def get_current_cluster_utilization():
+        cluster_resources = ray.cluster_resources()
+        available_resources = ray.available_resources()
+
+        return ClusterUtilization(
+            cluster_resources=cluster_resources, available_resources=available_resources
+        )
+
+
+def log_current_cluster_utlization(log_identifier: str):
```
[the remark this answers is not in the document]
raghumdani: I thought of doing it earlier, it contains lot of unnecessary values which causes cognitive load during debugging.

deltacat/utils/resources.py
raghumdani: I believe this is fetching cached data from GCS, so it should not involve lot of latency. Do you think there is any value in tracking this?

pdames: It's better to validate than assume when doing additional operations like this to collect telemetry that isn't necessary for E2E functionality in production environments. Ideally, these types of telemetry operations should be disabled automatically in production if their latency exceeds a given threshold, but this is a basic starting point to ensure that your assumption is correct today, and remains correct going forward.

raghumdani: Okay, that makes sense. One scenario I can think of is, when cluster is overloaded and GCS isn't easily accessible. I will log the latency and later (next PR) capture overall telemetry latency in the audit. FYI, this call takes `0.5857944488525391` milliseconds.

pdames (approved):
LGTM

Issue #133, opened 2023-06-08
Refactor the existing discover_delta to handle different cases


valiantljk:
Current discover_delta assumes either rebase partition locator or round completion file exists; In case of repartition, neither of them will exist, so we need to refactor the discover_delta to handle all different cases.

raghumdani:
repartition does not consider rebase. Hence, closing this.

PR #143, opened 2023-06-21
remove dependency on rebase_source_partition

valiantljk:
Repartition only needs to know the source and destination. This pr removes the dependency on rebase_source_partition and simplifies the delta discovery by directly scanning the compacted table, either ray or spark.
```
If repartition_during_rebase:
  use spark variant
else:
  use ray variant
```
Related issue: https://github.com/ray-project/deltacat/issues/133

### DIFF
--- a/deltacat/compute/compactor/repartition_session.py
+++ b/deltacat/compute/compactor/repartition_session.py
@@ -166,7 +166,9 @@ def repartition(
         bit_width_of_sort_keys,
         None,
     )
-    return rcf.write_repartition_completion_file(
-        repartition_completion_file_s3_url,
+    return rcf.write_round_completion_file(
+        None,
+        None,
         repartition_completion_info,
+        repartition_completion_file_s3_url,
     )
--- a/deltacat/compute/compactor/repartition_session.py
+++ b/deltacat/compute/compactor/repartition_session.py
@@ -35,7 +35,7 @@ from deltacat.utils.metrics import MetricsConfig
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 
-# TODO:(rootliu) move this repartition function to a separate module under compute
+# TODO: move this repartition function to a separate module under compute
 def repartition(
     source_partition_locator: PartitionLocator,
     destination_partition_locator: PartitionLocator,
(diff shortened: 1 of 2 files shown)


deltacat/compute/compactor/repartition_session.py
```diff
@@ -35,16 +35,15 @@
...
 def repartition(
     source_partition_locator: PartitionLocator,
     destination_partition_locator: PartitionLocator,
     repartition_args: Any,
     compaction_artifact_s3_bucket: str,
     last_stream_position_to_compact: int,
     repartition_type: RepartitionType = RepartitionType.RANGE,
-    rebase_source_partition_locator: Optional[PartitionLocator] = None,
-    rebase_source_partition_high_watermark: Optional[int] = None,
+    repartition_during_rebase: bool = True,
```
raghumdani, line 46: Do we need this arg at all? I think we are adding this because of discover_deltas signature. If it's rebase source != destination, otherwise source == destination. Right?

raghumdani (changes requested):
Mostly looks good. Please check minor comment.

deltacat/compute/compactor/repartition_session.py
valiantljk: I think source != destination for both rebase and incremental;
rebase: source=spark, destination=ray
incremental:source=source, destination=ray

raghumdani: What does incremental has to do with repartitioning? We are repartitioning all the deltas in the source and writing the output to destination. That is a simple contract.

valiantljk: makes sense.

valiantljk: simplified the arg and logic.

deltacat/compute/compactor/utils/round_completion_file.py
```diff
@@ -55,3 +52,24 @@ def write_round_completion_file(
     )
     logger.info(f"round completion file written to: {round_completion_file_s3_url}")
     return round_completion_file_s3_url
+
+
+# write rcf given s3 url
+def write_repartition_completion_file(
+    repartition_completion_file_s3_url: str,
```
raghumdani, line 59: It is not necessary that this method to be used only for repartition usecase. Any round completion file could be uploaded to s3 using this method.

deltacat/compute/compactor/repartition_session.py
```diff
@@ -161,18 +159,14 @@ def repartition(
...
     )
-    rcf_source_partition_locator = source_partition_locator
-    round_completion_file_s3_url = None
-    round_completion_file_s3_url = rcf.write_round_completion_file(
-        compaction_artifact_s3_bucket,
-        rcf_source_partition_locator,
-        new_round_completion_info,
+    return rcf.write_repartition_completion_file(
+        repartition_completion_file_s3_url,
+        repartition_completion_info,
```
raghumdani, line 171: We can only write if the rcf url is passed.

raghumdani (commented):
LGTM. Please address minor comment.

raghumdani (approved):
LGTM

PR #146, opened 2023-06-27
Capturing all the performance metrics in an audit

raghumdani:
[Description] It is important to capture audit automatically and analyze the datapoints from significant number of runs to recognize a pattern and come up with a formula to predict resource consumption. This would improve reliability and efficiency.

[Motivation] #141 

Example Audit Info:

```
"compactionAudit": {
    "deltacatVersion": "0.1.18b1",
    "clusterCpu": 3658,
    "deltaDiscoveryTimeInSeconds": 2.710052251815796,
    "inputFileCount": 31750,
    "inputSizeBytes": 5234658392405,
    "totalClusterMemoryBytes": 42087227851116,
    "hashBucketCount": 385,
    "hashBucketTimeInSeconds": 111.62457704544067,
    "totalClusterObjectStoreMemoryBytes": 23050000000000,
    "objectStoreMemoryUsedBytesByHashBucketing": 1719833215762,
    "inputRecords": 61183765994,
    "peakMemoryUsedBytesByHashBucketing": 1595277312,
    "dedupeTimeInSeconds": 351.89735102653503,
    "objectStoreMemoryUsedBytesByDedupe": 492114348197,
    "recordsDeduped": 0,
    "peakMemoryUsedBytesByDedupe": 40259301376,
    "materializeBuckets": 3658,
    "materializeTimeInSeconds": 297.6463842391968,
    "peakMemoryUsedBytesByMaterialize": 40259301376,
    "peakMemoryUsedBytes": 40259301376,
    "telemetryTimeInSeconds": 0.029021817000057126,
    "compactionTimeInSeconds": 768.3967623710632,
    "outputFileCount": 16766,
    "outputSizeBytes": 5003900801250,
    "outputSizePyarrowBytes": 16070617399347
  }
```

### DIFF
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -4,6 +4,8 @@ import functools
 import logging
 import ray
 import time
+import json
+from deltacat.aws import s3u as s3_utils
 import deltacat
 from deltacat import logs
 import pyarrow as pa
@@ -43,8 +45,7 @@ from deltacat.utils.metrics import MetricsConfig
 from deltacat.compute.compactor.model.compaction_session_audit_info import (
     CompactionSessionAuditInfo,
 )
-from deltacat.utils.resources import ClusterUtilization
-from deltacat.utils.performance import timed_invocation
+
 
 if importlib.util.find_spec("memray"):
     import memray
@@ -195,10 +196,22 @@ def _execute_compaction_round(
     **kwargs,
 ) -> Tuple[Optional[Partition], Optional[RoundCompletionInfo], Optional[str]]:
 
-    compaction_audit = CompactionSessionAuditInfo()
-    compaction_audit.set_deltacat_version(deltacat.__version__)
+    rcf_source_partition_locator = (
+        rebase_source_partition_locator
+        if rebase_source_partition_locator
+        else source_partition_locator
+    )
 
-    compaction_start = time.time()
+    base_audit_url = rcf_source_partition_locator.path(
+        f"{compaction_artifact_s3_bucket}/compaction-audit"
+    )
+    audit_url = f"{base_audit_url}.json"
+
+    logger.info(f"Compaction will be written to {audit_url}")
+
+    compaction_audit = CompactionSessionAuditInfo(deltacat.__version__, audit_url)
+
+    compaction_start = time.monotonic()
 
     if not primary_keys:
         # TODO (pdames): run simple rebatch to reduce all deltas into 1 delta
@@ -242,7 +255,7 @@ def _execute_compaction_round(
             f"{node_resource_keys}"
         )
 
-    compaction_audit.set_cluster_cpu(cluster_cpus)
+    compaction_audit.set_cluster_cpu_max(cluster_cpus)
     # create a remote options provider to round-robin tasks across all nodes or allocated bundles
     logger.info(f"Setting round robin scheduling with node id:{node_resource_keys}")
     round_robin_opt_provider = functools.partial(
@@ -281,7 +294,7 @@ def _execute_compaction_round(
         round_completion_info.high_watermark if round_completion_info else None
     )
 
-    delta_discovery_start = time.time()
+    delta_discovery_start = time.monotonic()
     (
         input_deltas,
         previous_last_stream_position_compacted_on_destination_table,
@@ -296,10 +309,12 @@ def _execute_compaction_round(
         **list_deltas_kwargs,
     )
 
-    delta_discovery_end = time.time()
+    delta_discovery_end = time.monotonic()
     compaction_audit.set_delta_discovery_time_in_seconds(
         delta_discovery_end - delta_discovery_start
     )
+
+    s3_utils.upload(compaction_audit.audit_url, str(json.dumps(compaction_audit)))
 
     if not input_deltas:
         logger.info("No input deltas found to compact.")
@@ -356,7 +371,7 @@ def _execute_compaction_round(
             "Multiple rounds are not supported. Please increase the cluster size and run again."
         )
 
-    hb_start = time.time()
+    hb_start = time.monotonic()
 
     hb_tasks_pending = invoke_parallel(
         items=uniform_deltas,
@@ -373,26 +388,24 @@ def _execute_compaction_round(
         read_kwargs_provider=read_kwargs_provider,
         deltacat_storage=deltacat_storage,
     )
+
+    hb_invoke_end = time.monotonic()
+
     logger.info(f"Getting {len(hb_tasks_pending)} hash bucket results...")
     hb_results: List[HashBucketResult] = ray.get(hb_tasks_pending)
     logger.info(f"Got {len(hb_results)} hash bucket results.")
+    hb_end = time.monotonic()
+    hb_results_retrieved_at = time.time()
 
-    hb_end = time.time()
-    compaction_audit.set_hash_bucket_time_in_seconds(hb_end - hb_start)
-
-    cluster_utilization_after_hb, cluster_util_after_hb_latency = timed_invocation(
-        ClusterUtilization.get_current_cluster_utilization
+    telemetry_time_hb = compaction_audit.save_step_stats(
+        CompactionSessionAuditInfo.HASH_BUCKET_STEP_NAME,
+        hb_results,
+        hb_results_retrieved_at,
+        hb_invoke_end - hb_start,
+        hb_end - hb_start,
     )
 
-    compaction_audit.set_total_cluster_object_store_memory_bytes(
-        cluster_utilization_after_hb.total_object_store_memory_bytes
-    )
-    compaction_audit.set_total_cluster_memory_bytes(
-        cluster_utilization_after_hb.total_memory_bytes
-    )
-    compaction_audit.set_object_store_memory_used_bytes_by_hash_bucketing(
-        cluster_utilization_after_hb.used_object_store_memory_bytes
-    )
+    s3_utils.upload(compaction_audit.audit_url, str(json.dumps(compaction_audit)))
 
     all_hash_group_idx_to_obj_id = defaultdict(list)
     for hb_result in hb_results:
@@ -407,18 +420,8 @@ def _execute_compaction_round(
     logger.info(
         f"Got {total_hb_record_count} hash bucket records from hash bucketing step..."
     )
-    peak_memory_during_hb = max(
-        [hb_result.peak_memory_usage_bytes for hb_result in hb_results]
-    )
-
-    telemetry_time_hb = sum(
-        [hb_result.telemetry_time_in_seconds for hb_result in hb_results]
-    )
 
     compaction_audit.set_input_records(total_hb_record_count.item())
-    compaction_audit.set_peak_memory_used_bytes_by_hash_bucketing(
-        peak_memory_during_hb.item()
-    )
 
     # TODO (pdames): when resources are freed during the last round of hash
     #  bucketing, start running dedupe tasks that read existing dedupe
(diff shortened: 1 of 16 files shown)


deltacat/compute/compactor/compaction_session.py
```diff
@@ -406,8 +462,28 @@ def _execute_compaction_round(
...
+
+    cluster_utilization_after_dd, cluster_util_after_dd_latency = timed_invocation(
+        ClusterUtilization.get_current_cluster_utilization
+    )
+
+    compaction_audit.set_object_store_memory_used_bytes_by_dedupe(
+        cluster_utilization_after_dd.used_object_store_memory_bytes
+        - cluster_utilization_after_hb.used_object_store_memory_bytes
+    )
+
```
pdames, line 477: Since we're dedicating an increasingly large number of lines to collecting telemetry info, it may be helpful to put larger blocks like this (plus the below lines around peak memory usage and telemetry collection time) inside of a dedicated method like `compaction_audit.save_step_stats("dedupe", dedupe_end - dedupe_start, ddr, cluster_util_after_dd_latency, telemetry_time_dd)`

deltacat/compute/compactor/compaction_session.py
```diff
@@ -432,6 +510,9 @@ def _execute_compaction_round(
 
     # parallel step 3:
     # materialize records to keep by index
+
+    materialize_start = time.time()
```
pdames, line 514:
```suggestion
    materialize_start = time.monotonic()
```

deltacat/compute/compactor/compaction_session.py
```diff
@@ -454,14 +535,54 @@ def _execute_compaction_round(
         deltacat_storage=deltacat_storage,
     )
     logger.info(f"Getting {len(mat_tasks_pending)} materialize result(s)...")
-    mat_results = ray.get(mat_tasks_pending)
+    mat_results: List[MaterializeResult] = ray.get(mat_tasks_pending)
     logger.info(f"Got {len(mat_results)} materialize result(s).")
 
-    log_current_cluster_utilization(log_identifier="post_materialize")
+    materialize_end = time.time()
```
pdames, line 541:
```suggestion
    materialize_end = time.monotonic()
```

deltacat/compute/compactor/compaction_session.py
```diff
@@ -497,12 +622,22 @@ def _execute_compaction_round(
...
     )
+
+    pyarrow_write_result = PyArrowWriteResult.union(
+        [m.pyarrow_write_result for m in mat_results]
+    )
+
+    compaction_audit.set_output_file_count(pyarrow_write_result.files)
+    compaction_audit.set_output_size_bytes(pyarrow_write_result.file_bytes)
+    compaction_audit.set_output_size_pyarrow_bytes(pyarrow_write_result.pyarrow_bytes)
+
```
pdames, line 633: Maybe also roll this into a single method like `compaction_audit.save_write_result(pyarrow_write_result)`.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -454,14 +535,54 @@ def _execute_compaction_round(
...
+    compaction_audit.set_telemetry_time_in_seconds(
+        cluster_util_after_hb_latency
+        + cluster_util_after_dd_latency
+        + telemetry_time_dd
+        + telemetry_time_hb
+        + telemetry_time_materialize
+        + cluster_util_after_mat_latency
+    )
+
```
pdames, line 585: As long as we're saving or at least capturing a sum of all of the telemetry collection latencies in the above recommended change to create a higher-level method to save step stats, then we can also roll this block + total compaction time + pyarrow write result persistence into a higher level method like `compaction_audit.save_round_completion_stats(total_obj_store_mem_used, pyarrow_write_result, compaction_end - compaction-start)`

I'm also assuming here that we'll take the max of all peak memory usage and the sum of all telemetry latencies previously collected inside of this method.

pdames (changes requested):
As a high-level comment, there are still a couple important pieces of timing information that I think we're still missing like: (1) The time Ray requires to schedule all requested tasks for a given BSP step (e.g. the time required to complete a call like `hb_tasks_pending = invoke_parallel(...)`), (2) The difference between the time that the last task for a BSP step completes vs. the time that collecting all results on the driver completes (e.g. the difference between the max epoch timestamp of the last line of code executing across all `hb.hash_bucket` tasks and the time that a call like `hb_results: List[HashBucketResult] = ray.get(hb_tasks_pending)` completes.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -188,6 +195,11 @@ def _execute_compaction_round(
     **kwargs,
 ) -> Tuple[Optional[Partition], Optional[RoundCompletionInfo], Optional[str]]:
 
+    compaction_audit = CompactionSessionAuditInfo()
+    compaction_audit.set_deltacat_version(deltacat.__version__)
+
+    compaction_start = time.time()
```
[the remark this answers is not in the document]
raghumdani: From the docs, what I can understand is `monotonic` gives better precision compared to `time` (which in most platforms gives 1 second precision). In this case, it makes sense to use it as we are only persisting difference between two times.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -366,6 +407,18 @@ def _execute_compaction_round(
...
+    peak_memory_during_hb = max(
+        [hb_result.peak_memory_usage_bytes for hb_result in hb_results]
+    )
+
+    telemetry_time_hb = sum(
+        [hb_result.telemetry_time_in_seconds for hb_result in hb_results]
+    )
+
+    compaction_audit.set_input_records(total_hb_record_count.item())
+    compaction_audit.set_peak_memory_used_bytes_by_hash_bucketing(
```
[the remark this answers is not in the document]
raghumdani: Renamed to `set_peak_memory_used_bytes_per_hash_bucket_task`. Since this goes to data warehouse, a single column for each task is easy to analyze instead of a dictionary.

deltacat/compute/compactor/compaction_session.py
raghumdani: Done. We can create somewhat generic methods like this. I created individual setters to not couple to the compaction session implementation.

raghumdani:
> As a high-level comment, there are still a couple important pieces of timing information that I think we're missing like: (1) The time Ray requires to schedule all requested tasks for a given BSP step (e.g. the time required to complete a call like `hb_tasks_pending = invoke_parallel(...)`), (2) The difference between the time that the last task for a BSP step completes vs. the time that collecting all results on the driver completes (e.g. the difference between the max epoch timestamp of the last line of code executing across all `hb.hash_bucket` tasks and the time that a call like `hb_results: List[HashBucketResult] = ray.get(hb_tasks_pending)` completes.

Captured these additional metrics in an audit.

deltacat/compute/compactor/compaction_session.py
```diff
@@ -403,11 +462,28 @@ def _execute_compaction_round(
         enable_profiler=enable_profiler,
         metrics_config=metrics_config,
     )
+
+    dedupe_invoke_end = time.monotonic()
     logger.info(f"Getting {len(dd_tasks_pending)} dedupe results...")
     dd_results: List[DedupeResult] = ray.get(dd_tasks_pending)
     logger.info(f"Got {len(dd_results)} dedupe results.")
+
+    dedupe_results_retrieved_at = time.time()
```
pdames, line 471: Minor: Can we leave a comment here about why we use time.time() in this case vs. other timing mechanisms? I can imagine someone inadvertently breaking this in the future and it being easily missed during the code review...

deltacat/compute/compactor/model/compaction_session_audit_info.py
```diff
@@ -0,0 +1,658 @@
...
+        task_results_retrieved_at: float,
+        invoke_time_in_seconds: float,
+        task_time_in_seconds: float,
+    ) -> float:
+        """
+        Saves the stats by calling individual setters and returns the cluster telemetry time.
+        """
+
+        last_hb_task_completed_at = max(
+            hb_result.task_completed_at for hb_result in task_results
```
pdames, line 558:
```suggestion
        last_task_completed_at = max(
            result.task_completed_at for result in task_results
```
Since this is now used for more than just hash bucketing...

deltacat/compute/compactor/model/materialize_result.py
```diff
@@ -13,15 +14,19 @@ def of(
         delta: Delta,
         task_index: int,
         pyarrow_write_result: PyArrowWriteResult,
-        count_of_src_dfl_not_touched: Optional[int] = 0,
-        count_of_src_dfl: Optional[int] = 0,
+        referenced_pyarrow_write_result: Optional[PyArrowWriteResult] = None,
+        peak_memory_usage_bytes: Optional[np.double] = None,
+        telemetry_time_in_seconds: Optional[np.double] = None,
+        task_completed_at: Optional[np.double] = None,
```
pdames, line 20: Since these 3 members are now common to all results, lets push them down into a `StepResult` base class that all materialize/hash-bucket/dedupe results extend.

deltacat/compute/compactor/model/compaction_session_audit_info.py
```diff
@@ -0,0 +1,658 @@
...
+        self.set_total_object_store_memory_used_bytes(
+            cluster_utilization_after_task.used_object_store_memory_bytes
+        )
+
+        self[
+            f"{step_name}PostObjectStoreMemoryUsedBytes"
+        ] = cluster_utilization_after_task.used_object_store_memory_bytes
+
+        peak_task_memory = max(
+            hb_result.peak_memory_usage_bytes for hb_result in task_results
```
pdames, line 587:
```suggestion
            result.peak_memory_usage_bytes for result in task_results
```

deltacat/compute/compactor/model/compaction_session_audit_info.py
```diff
@@ -0,0 +1,658 @@
...
+        )
+
+        telemetry_time = sum(
+            hb_result.telemetry_time_in_seconds for hb_result in task_results
```
pdames, line 591:
```suggestion
            result.telemetry_time_in_seconds for result in task_results
```

deltacat/compute/compactor/model/compaction_session_audit_info.py
```diff
@@ -0,0 +1,658 @@
...
+        """
+        Saves the stats by calling individual setters and returns the cluster telemetry time.
+        """
+
+        last_hb_task_completed_at = max(
+            hb_result.task_completed_at for hb_result in task_results
+        )
+
+        self[f"{step_name}ResultWaitTimeInSeconds"] = (
+            task_results_retrieved_at - last_hb_task_completed_at.item()
```
pdames, line 562:
```suggestion
            task_results_retrieved_at - last_task_completed_at.item()
```

pdames (changes requested):
Looks pretty good! Just a few minor comments.

deltacat/compute/compactor/model/compaction_session_audit_info.py
raghumdani: fixed.

deltacat/compute/compactor/model/materialize_result.py
raghumdani: I wanted to do that but I was skeptical about the performance deterioration as we had observed that when we started returning dataclass. I believe composition is better than inheritance for reusing variables. Hence, clubbed them into a single object.

raghumdani: The job now is taking 10 mins more to complete. I believe it's due to additional overhead due to nested tuple. Hence, I will keep it as is.

pdames (approved):

pdames (approved):
Latest commit looks good - thanks!

Issue #148, opened 2023-07-06
Inject dynamic object store implementations to share objects across cluster


raghumdani:
We would ideally need to use a particular object store like Memcached to avoid setting object store memory upfront and S3 object store if the files are too big (>1GB) to being with. Hence, it will be beneficial have this as an interface and user decides the object store implementation he/she needs.

PR #149, opened 2023-07-06
Refactoring the compact_partition to support different object store implementations

raghumdani:
#148 

- Tested all implementations on internal 15TB table
- Unit test pending for each object store impl. 

update: unit tests added

### DIFF
--- a/deltacat/compute/compactor/steps/dedupe.py
+++ b/deltacat/compute/compactor/steps/dedupe.py
@@ -1,5 +1,6 @@
 import importlib
 import logging
+from typing import Optional
 import time
 from collections import defaultdict
 from contextlib import nullcontext
@@ -8,7 +9,6 @@ import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc
 import ray
-from ray import cloudpickle
 
 from deltacat import logs
 from deltacat.compute.compactor import (
@@ -25,6 +25,7 @@ from deltacat.utils.ray_utils.runtime import (
 )
 from deltacat.utils.performance import timed_invocation
 from deltacat.utils.metrics import emit_timer_metrics, MetricsConfig
+from deltacat.io.object_store import IObjectStore
 from deltacat.utils.resources import get_current_node_peak_memory_usage_in_bytes
 
 if importlib.util.find_spec("memray"):
@@ -106,6 +107,7 @@ def _timed_dedupe(
     num_materialize_buckets: int,
     dedupe_task_index: int,
     enable_profiler: bool,
+    object_store: Optional[IObjectStore],
 ):
     task_id = get_current_ray_task_id()
     worker_id = get_current_ray_worker_id()
@@ -114,15 +116,12 @@ def _timed_dedupe(
     ) if enable_profiler else nullcontext():
         # TODO (pdames): mitigate risk of running out of memory here in cases of
         #  severe skew of primary key updates in deltas
-        src_file_records_obj_refs = [
-            cloudpickle.loads(obj_id_pkl) for obj_id_pkl in object_ids
-        ]
         logger.info(
             f"[Dedupe task {dedupe_task_index}] Getting delta file envelope "
-            f"groups for {len(src_file_records_obj_refs)} object refs..."
+            f"groups for {len(object_ids)} object refs..."
         )
 
-        delta_file_envelope_groups_list = ray.get(src_file_records_obj_refs)
+        delta_file_envelope_groups_list = object_store.get_many(object_ids)
         hb_index_to_delta_file_envelopes_list = defaultdict(list)
         for delta_file_envelope_groups in delta_file_envelope_groups_list:
             for hb_idx, dfes in enumerate(delta_file_envelope_groups):
@@ -201,7 +200,6 @@ def _timed_dedupe(
                 src_file_id_to_row_indices[src_dfl].append(row_idx_col[row_idx])
 
         logger.info(f"Finished all dedupe rounds...")
-        mat_bucket_to_src_file_record_count = defaultdict(dict)
         mat_bucket_to_src_file_records: Dict[
             MaterializeBucketIndex, DeltaFileLocatorToRecords
         ] = defaultdict(dict)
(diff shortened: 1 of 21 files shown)


deltacat/io/redis_object_store.py
```diff
@@ -0,0 +1,114 @@
...
+            cache_result = client.mget(uids)
+            assert len(cache_result) == len(
+                uids
+            ), "Not all values were returned from cache"
+
+            total_bytes = 0
+
+            deserialize_start = time.monotonic()
+            for serialized in cache_result:
+                deserialized = cloudpickle.loads(serialized)
```
valiantljk, line 77: This assumes that the object is cloudpickled first, which may not be true

valiantljk (commented):
Great work!
Major concern: other than plasma object store, how does other 'object store' solutions handle object recovery in case of failure? And ray has the locality-aware scheduling, so that tasks can be placed close to object, I assume this object ref resolution only works on plasma object store; Thirdly, when the task and object are not physically close, ray will trigger the inter-node object transfer, how does the new proposed object store path handle this?

deltacat/compute/compactor/steps/dedupe.py
```diff
@@ -114,15 +116,12 @@ def _timed_dedupe(
...
-            cloudpickle.loads(obj_id_pkl) for obj_id_pkl in object_ids
-        ]
         logger.info(
             f"[Dedupe task {dedupe_task_index}] Getting delta file envelope "
-            f"groups for {len(src_file_records_obj_refs)} object refs..."
+            f"groups for {len(object_ids)} object refs..."
         )
 
-        delta_file_envelope_groups_list = ray.get(src_file_records_obj_refs)
+        delta_file_envelope_groups_list = object_store.get_many(object_ids)
```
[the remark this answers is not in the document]
raghumdani: no. There is a default set in `compaction_session.py` and it doesn't change current behavior.

deltacat/io/object_store.py
```diff
@@ -0,0 +1,51 @@
...
+        of object references corresponding to each object in the input.
+        """
+
+    def get(self, ref: Any, *args, **kwargs) -> object:
+        """
+        Get a single object from an object store.
+        """
+        return self.get_many([ref])[0]
+
+    def get_many(self, refs: List[Any], *args, **kwargs) -> List[object]:
```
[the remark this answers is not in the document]
raghumdani: it's mainly to bring all object stores under a common interface. Implementation wise, we will use `ray.get` to retrieve all objects.

deltacat/io/redis_object_store.py
raghumdani: you can only get what you put. So, from the `put` and `put_many` implementation of `RedisObjectStore`, it should be always true.

raghumdani:
> other than plasma object store, how does other 'object store' solutions handle object recovery in case of failure?

Can you explain the failure scenario? Behavior wise, it should be same. 

> And ray has the locality-aware scheduling, so that tasks can be placed close to object, I assume this object ref resolution only works on plasma object store

Memcached and Redis both put the object into the node they were called in. Did you mean any other feature we won't make use of?

> when the task and object are not physically close, ray will trigger the inter-node object transfer, how does the new proposed object store path handle this?

each object store is handling it differently, for example 1) S3 is global, 2) same file system must be attached to all instances, 3) memcached and redis will bring in object from it's owner node, 4) ray handles owner resolution for us.

deltacat/io/redis_object_store.py
valiantljk: what if later we use this object store interface to put un-pickled object?

valiantljk:
> > other than plasma object store, how does other 'object store' solutions handle object recovery in case of failure?
> 
> Can you explain the failure scenario? Behavior wise, it should be same.
> 
features like 'object reconstruction' seems to be missing in the proposed new interface. Failures can be any kind, e.g., node failure, owner failure, etc: https://docs.ray.io/en/latest/ray-core/fault_tolerance/objects.html
> > And ray has the locality-aware scheduling, so that tasks can be placed close to object, I assume this object ref resolution only works on plasma object store
> 
> Memcached and Redis both put the object into the node they were called in. Did you mean any other feature we won't make use of?
Yeah, when we pass the ref to tasks, it'll be auto-resolved, we don't need to call 'get' explicitly, and object transfer will be triggered underneath. Also, what if memory is full and disk spilling is needed?
> 
> > when the task and object are not physically close, ray will trigger the inter-node object transfer, how does the new proposed object store path handle this?
> 
> each object store is handling it differently, for example 1) S3 is global, 2) same file system must be attached to all instances, 3) memcached and redis will bring in object from it's owner node, 4) ray handles owner resolution for us.

I realize this question is same with the object ref resolution, so if we have this new object store interface, we'll assume future dev will always explicitly call get/get_many to retrieve the objects. This seems to be against some of the Ray object store design?

deltacat/io/redis_object_store.py
raghumdani: That will be part of the implementation whether to pickle or not. The `object_store.get` will return what was `object_store.put`

raghumdani:
> features like 'object reconstruction' seems to be missing in the proposed new interface. Failures can be any kind, e.g., node failure, owner failure, etc: https://docs.ray.io/en/latest/ray-core/fault_tolerance/objects.html

object reconstruction is a ray feature that works on task returns in the ray plasma store when a worker dies. However, if we call ray.put explicitly we are responsible for reconstructing the object anyway. What is the behavior or feature are you looking for?

> Yeah, when we pass the ref to tasks, it'll be auto-resolved, we don't need to call 'get' explicitly, and object transfer will be triggered underneath. Also, what if memory is full and disk spilling is needed?

This is a ray feature, not an object store feature. Moreover, for our use-case we are circumventing this by pickling the object ref which is an anti-pattern.

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -96,10 +78,13 @@ def _stage_delta_from_manifest_entry_reference_list(
             delta_type == DeltaType.UPSERT
         ), "Stage delta with existing manifest entries only supports UPSERT delta type!"
         manifest = Manifest.of(entries=manifest_entry_list_reference, uuid=str(uuid4()))
-        delta = _stage_delta_implementation(
-            data=manifest,
-            partition=partition,
-            stage_delta_from_existing_manifest=True,
+        delta = Delta.of(
```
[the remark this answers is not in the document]
raghumdani: Yes, that was intentional as it's removes the redundant method. We can anyway modify that when we implement `stage_delta`.

valiantljk:
LGTM. Please note down any side effect, e.g., any difference between the new object store and the default plasma store.

valiantljk (approved):

Zyiqin-Miranda:
LGTM

PR #151, opened 2023-07-06
Adding a data model to represent compact_partition function parameters

pfaraone:
### Notes
- This PR adds the CompactPartitionParams class, which represents the compact_partition (deltacat/compute/compactor/compaction_session.py) parameters. The class has properties for all of the parameters that are currently used by the compaction job, and it also has a serialize() method that can be used to serialize the class to a JSON-formatted string.


deltacat/tests/compactor/test_compact_partition_params.py
```diff
@@ -0,0 +1,204 @@
+import json
+
+import pytest
+
+
+@pytest.fixture
+def valid_compact_partition_params():
```
raghumdani, line 7: We have been using unittest and an unit test class to test all the methods. It will provide a consistent test report.

raghumdani (commented):
Looks good. Mostly minor comments. Please squash and merge.

deltacat/compute/compactor/model/compact_partition_params.py
```diff
@@ -0,0 +1,153 @@
+from __future__ import annotations
+
+import copy
+import json
+from typing import Any, Dict, List, Optional
+
+from deltacat.types.media import ContentType
+
+
+class CompactPartitionParams(dict):
```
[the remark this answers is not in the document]
raghumdani: As an alternative (for backward compatibility) you can create a separate method.

raghumdani (approved):
LGTM

deltacat/compute/compactor/compaction_session.py
```diff
@@ -640,3 +643,14 @@ def _execute_compaction_round(
...
+
+def compact_partition_from_request(
+    compact_partition_params: CompactPartitionParams,
+) -> Optional[str]:
+    """
+    Wrapper for compact_partition that allows for the compact_partition parameters to be
+    passed in as a custom dictionary-like CompactPartitionParams object.
+    :param compact_partition_params:
+    """
+    return compact_partition(**compact_partition_params)
```
rkenmi, line 656: I think we would also need to pass in the non-keyword arguments here to match the function signature of compact_partition.

PR #153, opened 2023-07-17
[skip untouched files]Disable copy by reference during backfill and rebase only compaction; Fix referenced manifest PyarrowWriteResult

Zyiqin-Miranda:
This PR adds support for 1. Disable manifest entry copy by reference for special compaction use cases 2. Correct PyarrowWriteResult when using referenced manifest entry metadata.

Testing:
1. Tested disable copy by reference by passing a rebase_source_partition_locator.
2. Tested copy by reference in normal incremental compaction case.

### DIFF
--- a/deltacat/compute/compactor/steps/materialize.py
+++ b/deltacat/compute/compactor/steps/materialize.py
@@ -64,6 +64,7 @@ def materialize(
     dedupe_task_idx_and_obj_id_tuples: List[DedupeTaskIndexWithObjectId],
     max_records_per_output_file: int,
     compacted_file_content_type: ContentType,
+    enable_manifest_entry_copy_by_reference: bool,
     enable_profiler: bool,
     metrics_config: MetricsConfig,
     schema: Optional[pa.Schema] = None,
@@ -231,7 +232,9 @@ def materialize(
                 record_numbers_length += 1
                 mask_pylist[record_number] = True
             if (
-                record_numbers_length == src_file_record_count
+                round_completion_info
+                and enable_manifest_entry_copy_by_reference
+                and record_numbers_length == src_file_record_count
                 and src_file_partition_locator
                 == round_completion_info.compacted_delta_locator.partition_locator
             ):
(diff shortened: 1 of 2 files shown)


valiantljk:
LGTM

valiantljk (approved):

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -230,26 +231,33 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
+                        1,
+                        untouched_src_manifest_entry.meta.source_content_length,
+                        untouched_src_manifest_entry.meta.content_length,
+                        src_file_record_count,
+                    )
+                    referenced_pyarrow_write_results.append(
+                        referenced_pyarrow_write_result
+                    )
+                    can_copy_by_reference = True
+            if not can_copy_by_reference:
```
raghumdani, line 260: Can it just be the else part? Why do we need `can_copy_by_reference` var?

raghumdani (approved):
LGTM

deltacat/compute/compactor/steps/materialize.py
Zyiqin-Miranda: There are two nested if statement here, 
if condition A:
   if condition B:
One of which being False we will need to run Else part.
`can_copy_by_reference` var set to True only if both if statement is True, so we guarantee any False of two if statement will need to download the manifest.

raghumdani:
```
if (               enable_manifest_entry_copy_by_reference and round_completion_info and
                    record_numbers_length == src_file_record_count
                    and src_file_partition_locator
                    == round_completion_info.compacted_delta_locator.partition_locator
                ):
else:
      # no skipping
```

Isn't the above and the your code equivalent?

Zyiqin-Miranda: This will break when this is no round_completion_info right?

Zyiqin-Miranda: @raghumdani, updated

PR #158, opened 2023-07-26
Allow s3_client_kwargs to be passed into repartition

rkenmi:
Allows S3 keyword arguments to be passed into repartition session for use cases such as AWS credential overrides.

### DIFF
--- a/deltacat/compute/compactor/compaction_session.py
+++ b/deltacat/compute/compactor/compaction_session.py
@@ -115,7 +115,7 @@ def compact_partition(
     read_kwargs_provider: Optional[ReadKwargsProvider] = None,
     s3_table_writer_kwargs: Optional[Dict[str, Any]] = None,
     object_store: Optional[IObjectStore] = RayPlasmaObjectStore(),
-    s3_client_kwargs: Optional[Dict[str, Any]] = {},
+    s3_client_kwargs: Optional[Dict[str, Any]] = None,
     deltacat_storage=unimplemented_deltacat_storage,
     **kwargs,
 ) -> Optional[str]:
@@ -283,6 +283,9 @@ def _execute_compaction_round(
     # "safe" number of parallel tasks that our autoscaling cluster could handle
     max_parallelism = int(cluster_cpus)
     logger.info(f"Max parallelism: {max_parallelism}")
+
+    if s3_client_kwargs is None:
+        s3_client_kwargs = {}
 
     # read the results from any previously completed compaction round
     round_completion_info = None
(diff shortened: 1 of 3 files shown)


deltacat/compute/compactor/repartition_session.py
```diff
@@ -53,6 +53,7 @@ def repartition(
     pg_config: Optional[PlacementGroupConfig] = None,
     list_deltas_kwargs: Optional[Dict[str, Any]] = None,
     read_kwargs_provider: Optional[ReadKwargsProvider] = None,
+    s3_client_kwargs: Optional[Dict[str, Any]] = None,
```
raghumdani, line 56:
```suggestion
    s3_client_kwargs: Optional[Dict[str, Any]] = {},
```

deltacat/compute/compactor/repartition_session.py
```diff
@@ -166,9 +167,13 @@ def repartition(
         bit_width_of_sort_keys,
         None,
     )
+    if s3_client_kwargs is None:
+        s3_client_kwargs = {}
+
```
raghumdani, line 172:
```suggestion
```

raghumdani (commented):
Couple of minor comments. LGTM

deltacat/compute/compactor/repartition_session.py
pfaraone: Mutable default arguments are not ideal - see https://docs.python-guide.org/writing/gotchas/. We can accomplish optional argument behavior with this:

```
...
s3_client_kwargs: Optional[Dict[str, Any]] = None):
if s3_client_kwargs is None:
        s3_client_kwargs = {}
```

rkenmi: I chose the default `None` approach since this would cause `s3_client_kwargs` to be mutable on subsequent calls. [ref](https://stackoverflow.com/questions/366422/how-can-i-avoid-issues-caused-by-pythons-early-bound-default-parameters-e-g-m)

raghumdani: Good point. Can you change this behavior in `compaction_session` as well?

raghumdani (approved):
LGTM, please also change the default to None `compaction_session`.

pfaraone (approved):
Could you publish a pre-release release here for 0.1.18.beta10 after merging? -See https://github.com/ray-project/deltacat/releases

valiantljk (approved):
LGTM

deltacat/compute/compactor/compaction_session.py
```diff
@@ -284,6 +284,9 @@ def _execute_compaction_round(
     max_parallelism = int(cluster_cpus)
     logger.info(f"Max parallelism: {max_parallelism}")
 
+    if s3_client_kwargs is None:
```
raghumdani, line 287: The defaulting must be done in compact_partition method itself.

raghumdani (commented):
Please address minor comment.

PR #159, opened 2023-07-26
Move s3_client_kwargs default setter to parent scope

rkenmi:
Setting default value in parent scope to avoid unpacking None type values.


pfaraone (approved):
LGTM

valiantljk (approved):
LGTM

PR #160, opened 2023-07-28
Add pytest benchmarking for Parquet reads

jaychia:
# Summary

This PR adds in `pytest-benchmark` harnesses for benchmarking DeltaCat's Parquet reads across different possible solutions.

The various axes being benchmarked in this PR:

1. `read_fn` is the function used to read the Parquet file (DeltaCat vs PyArrow vs Daft Table)
2. We use publicly-available Parquet files hosted in AWS S3, where different files have different data distributions/profiles
3. We also benchmark single-column-reads vs all-column-reads

## Running Benchmarks

```
pytest deltacat/benchmarking/benchmark_parquet_reads.py --benchmark-only --benchmark-group-by=group,param:name
```

The above will output a benchmarking report, comparing the performance of various functions which perform a Parquet file read into an Arrow table.

### DIFF
--- a/deltacat/benchmarking/README.md
+++ b/deltacat/benchmarking/README.md
@@ -2,8 +2,8 @@
 
 ## Setup
 
-1. Install DeltaCat's `requirements.txt` and `dev-requirements.txt`
-2. Install the appropriate versions of additional packages that are being benchmarked (e.g. `pip install getdaft`)
+1. Install DeltaCat's `benchmark-requirements.txt`
+2. Ensure the appropriate versions of additional packages that are being benchmarked are installed (e.g. `pip install getdaft==X.X.XX`)
 
 ## Adding Benchmarks
 
--- a/deltacat/benchmarking/README.md
+++ b/deltacat/benchmarking/README.md
@@ -1,0 +1,18 @@
+# Benchmarking
+
+## Setup
+
+1. Install DeltaCat's `requirements.txt` and `dev-requirements.txt`
+2. Install the appropriate versions of additional packages that are being benchmarked (e.g. `pip install getdaft`)
+
+## Running Benchmarks
+
+### Parquet Reads
+
+We recommend running these benchmarks in a production AWS environment with the appropriate setup for high bandwidth access to AWS S3.
+
+```bash
+pytest deltacat/benchmarking/benchmark_parquet_reads.py --benchmark-only --benchmark-group-by=group,param:path
+```
+
+Grab a coffee, it will be a few minutes!
--- a/deltacat/benchmarking/README.md
+++ b/deltacat/benchmarking/README.md
@@ -12,7 +12,7 @@
 We recommend running these benchmarks in a production AWS environment with the appropriate setup for high bandwidth access to AWS S3.
 
 ```bash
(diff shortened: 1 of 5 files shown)


raghumdani (approved):
Looks good to me. Thanks for the benchmarking script.

valiantljk (approved):
LGTM. Thanks for the contribution.

deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,18 @@
+# Benchmarking
+
+## Setup
+
+1. Install DeltaCat's `requirements.txt` and `dev-requirements.txt`
+2. Install the appropriate versions of additional packages that are being benchmarked (e.g. `pip install getdaft`)
+
```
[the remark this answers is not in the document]
jaychia: Thanks and will do!

deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,18 @@
+# Benchmarking
+
+## Setup
+
+1. Install DeltaCat's `requirements.txt` and `dev-requirements.txt`
```
pdames, line 5: What do you think about introducing a new `benchmark-requirements.txt` that includes `dev-requirements.txt` and daft, then just changing this line to tell the reader to just install `benchmark-requirements.txt`?

deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,18 @@
...
+2. Install the appropriate versions of additional packages that are being benchmarked (e.g. `pip install getdaft`)
+
+## Running Benchmarks
+
+### Parquet Reads
+
+We recommend running these benchmarks in a production AWS environment with the appropriate setup for high bandwidth access to AWS S3.
```
pdames, line 12:
```suggestion
We recommend running these benchmarks in an AWS environment configured for high bandwidth access to AWS S3. For example, on an EC2 instance with enhanced networking support: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html.
```

deltacat/benchmarking/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+from deltacat.utils.pyarrow import s3_file_to_table
+from deltacat.types.media import (
+    ContentEncoding,
+    ContentType,
+)
+
+
+def pyarrow_read(path: str, columns: list[str] | None = None) -> pa.Table:
+    assert path.startswith("s3://")
```
pdames, line 19: Minor (same for other asserts):
```suggestion
    assert path.startswith("s3://"), f"Expected file path to start with 's3://', but got {path}."
```

pdames (approved):
Looks good overall! Just a few minor suggestions. Feel free to address them now or in a follow-up PR.

dev-requirements.txt
```diff
@@ -6,3 +6,4 @@ memray == 1.6.0; platform_system != "Windows" and sys_platform != "darwin" and p
 pre-commit == 2.20.0
 pytest == 7.2.0
 pytest-cov == 4.0.0
+pytest-benchmark == 4.0.0
```
pdames, line 9:
```suggestion
pytest-benchmark == 4.0.0
pytest-cov == 4.0.0
```
I think putting these in lexicographic order should fix the linter...

PR #161, opened 2023-07-28
keep null row and remove dw_int64 column

valiantljk:
Fixed Issue 1: a new dw_last_update_int64 column is added for filtering, but need to be dropped 
Fixed issue 2: dw_last_update could have null value, avoid dropping it during filtering

### DIFF
--- a/deltacat/compute/compactor/steps/repartition.py
+++ b/deltacat/compute/compactor/steps/repartition.py
@@ -126,9 +126,12 @@ def repartition_range(
     partition_deltas: List[Delta] = []
     for partition_tables in partitioned_tables_list:
         if len(partition_tables) > 0:
+            col_name_int64_id = partition_tables[0].schema.get_field_index(
+                col_name_int64
+            )
             partition_table: pa.Table = pa.concat_tables(
                 partition_tables
-            ).remove_column(0)
+            ).remove_column(col_name_int64_id)
             assert col_name_int64 not in partition_table.schema.names
             if len(partition_table) > 0:
                 partition_table_length += len(partition_table)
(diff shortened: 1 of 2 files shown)


deltacat/compute/compactor/steps/repartition.py
```diff
@@ -117,12 +118,18 @@ def repartition_range(
...
+
     partition_table_length = 0
     # After re-grouping the tables by specified ranges, for each group, we need concat and stage the tables
     partition_deltas: List[Delta] = []
     for partition_tables in partitioned_tables_list:
         if len(partition_tables) > 0:
-            partition_table: pa.Table = pa.concat_tables(partition_tables)
+            partition_table: pa.Table = pa.concat_tables(
+                partition_tables
+            ).remove_column(0)
```
raghumdani, line 131: should we use `drop` method with actual column name created above? Just to make sure we don't drop any actual column from table?

raghumdani (commented):
Thanks for the fix.

deltacat/compute/compactor/steps/repartition.py
valiantljk: good idea

valiantljk: turned out drop_columns/drop is not supported in our current version 10.0.1; I'll just use remove_column but by first locating the index with the column name;

raghumdani: Then how does this line work? https://github.com/ray-project/deltacat/blob/037b584d7b7330dca6be94adaf279ff01593fee8/deltacat/compute/compactor/steps/hash_bucket.py#L59

valiantljk: updated.

raghumdani (approved):
looks good. Please check why the checks are failing.

Issue #162, opened 2023-07-31
Support reading Parquet meta using ParquetFile class in pyarrow


raghumdani:
Right now we read the entire table using `read_parquet` method of pyarrow. However, we lose the parquet metadata in the process and this will eagerly load all the data. In this commit we will use `ParquetFile` (https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html)

Epic #150

PR #163, opened 2023-07-31
Adding support for reading table into ParquetFile

raghumdani:
[Motivation] #162

Testing
-----

```python
delta_result = DS.list_deltas(namespace=namespace, 
...                         table_name=table_name, 
...                         partition_values=ordered_partition_values, 
...                         table_version=table_version, 
...                         include_manifest=True,
...                         equivalent_table_types=[])

delta = delta_result.read_page()[0]

table = DS.download_delta_manifest_entry(delta, 0, TableType.PYARROW_PARQUET)
# <pyarrow.parquet.core.ParquetFile object at 0xffffa168c100>
```

### DIFF
--- a/deltacat/aws/s3u.py
+++ b/deltacat/aws/s3u.py
@@ -3,6 +3,8 @@ import multiprocessing
 from functools import partial
 from typing import Any, Callable, Dict, Generator, List, Optional, Union
 from uuid import uuid4
+from botocore.config import Config
+from deltacat.aws.constants import BOTO_MAX_RETRIES
 
 import pyarrow as pa
 import ray
@@ -385,14 +387,16 @@ def download_manifest_entry(
     content_encoding: Optional[ContentEncoding] = None,
 ) -> LocalTable:
 
+    conf = Config(retries={"max_attempts": BOTO_MAX_RETRIES, "mode": "adaptive"})
     s3_client_kwargs = (
         {
             "aws_access_key_id": token_holder["accessKeyId"],
             "aws_secret_access_key": token_holder["secretAccessKey"],
             "aws_session_token": token_holder["sessionToken"],
+            "config": conf,
         }
         if token_holder
-        else {}
+        else {"config": conf}
     )
     if not content_type:
         content_type = manifest_entry.meta.content_type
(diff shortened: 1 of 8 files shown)


raghumdani:
The repartition tests are failing due to a usage of `float("inf")` for comparison in `Expression`. @valiantljk is working on fixings these tests.

Zyiqin-Miranda:
Looks good overall! Just one quick question:
Could you explain why we don't have to support the `input_file_init` for different `content_encoding` as [here](https://github.com/ray-project/deltacat/blob/main/deltacat/utils/pyarrow.py#L259)?

raghumdani:
@Zyiqin-Miranda Good question. The creation of ParquetFile instance does not lead to file read. Content encoding is used to decode the file bytes (i.e., after the file has been downloaded) but in our file has not yet been downloaded. The assumption is that pyarrow library should be capable of decoding appropriate encodings.

Zyiqin-Miranda:
> @Zyiqin-Miranda Good question. The creation of ParquetFile instance does not lead to file read. Content encoding is used to decode the file bytes (i.e., after the file has been downloaded) but in our case file has not yet been downloaded. The assumption is that pyarrow library should be capable of decoding appropriate encodings.

Thanks for the explanation. So to actually read the table, we need to use something like `parquet_file.read()` and assume that the file will be decoded appropriately?

raghumdani:
> > @Zyiqin-Miranda Good question. The creation of ParquetFile instance does not lead to file read. Content encoding is used to decode the file bytes (i.e., after the file has been downloaded) but in our case file has not yet been downloaded. The assumption is that pyarrow library should be capable of decoding appropriate encodings.
> 
> Thanks for the explanation. So to actually read the table, we need to use something like `parquet_file.read()` and assume that the file will be decoded appropriately?

Yes.

Zyiqin-Miranda (approved):

raghumdani:
Screen Shot 2023-07-31 at 7 37 04 PM

Here is the performance comparison between using `pyarrow.fs.S3FileSystem` vs `s3fs.S3FileSystem`. Overall, the performance is almost the same (looking at median) however s3fs provides capability to easily pass the `s3_client_kwargs` which might be handy in future while `pyarrow.fs.S3FileSystem` does not.

Issue #165, opened 2023-07-31
Cleanup rehash and other dead primary key index code


raghumdani:
Hash partitioning leverages several primary key index module methods like grouping records via primary key hash etc. Hence, we need to ensure the module conforms to new requirements and this task is to remove the dead functions in the module.

raghumdani:
Tracking epic: #150

PR #166, opened 2023-07-31
Cleaning up rehashing logic as it is a dead code as of now.

raghumdani:
[Motivation] #165


deltacat/__init__.py
```diff
@@ -43,7 +43,7 @@
 
 deltacat.logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-__version__ = "0.1.18b12"
```
pfaraone, line 46: Version number HEAD is 0.1.18b12

pfaraone: LGTM apart from the version number

raghumdani: Fixed.

pfaraone (approved):
LGTM

Issue #167, opened 2023-08-01
Support pyarrow 12.0.1 for repartitioning 


raghumdani:
Currently repartitioning has two issues:

1. The stream position set in the delta is current epoch which leads to race condition and correctness issue. 
2. The filter expression in repartitioning uses float("inf") that is unsupported in pyarrow 12.0.1. 

Hence, as part of this change, we will fix both the bugs. 

Epic: #150

PR #168, opened 2023-08-01
Fix stream position and support latest pyarrow

raghumdani:
[Motivation] #167


valiantljk (approved):
LGTM, thanks for the fix.

Issue #174, opened 2023-08-04
Add local deltacat storage that mimics the Catalog behavior


raghumdani:
## Use-cases

- Write functional tests for the DeltaCAT compute module seamlessly without connecting to the internal catalog.
- Provide examples of how the DeltaCAT storage module works via it's local implementation. 
- Ensure migrating to a `compactor_v2` is a smooth experience. 

#150

PR #175, opened 2023-08-04
Adding local deltacat storage module

raghumdani:
#174


deltacat/storage/model/table_version.py
```diff
@@ -30,6 +32,7 @@ def of(
         table_version.description = description
         table_version.properties = properties
         table_version.content_types = content_types
+        table_version.sort_keys = sort_keys
```
Zyiqin-Miranda, line 35: Could you explain what's the intention of adding sort_keys to table version?

raghumdani: Table version has primary keys as the attribute, but not the sort key. It is not captured in the catalog object anywhere but used in compute module.

rkenmi (approved):

Zyiqin-Miranda (approved):

PR #180, opened 2023-08-08
Trigger a "Publish Python distributions to PyPI" GH Actions workflow on additional event types

pfaraone:
Editing and creating a release should now trigger the Publish Python distributions to PyPI workflow. Previously, a new release *and* tag had to be created to trigger the workflow

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,7 +1,7 @@
 name: Publish Python distributions to PyPI
 on:
   release:
-    types: [published] # triggered whenever a new GitHub release is published
+    types: [published, created, edited] # triggered whenever a new GitHub release is published
 jobs:
   build-n-publish:
     name: Build and publish Python distributions to PyPI


.github/workflows/publish-to-pypi.yml
```diff
@@ -1,7 +1,7 @@
 name: Publish Python distributions to PyPI
 on:
   release:
-    types: [published] # triggered whenever a new GitHub release is published
+    types: [published, created, edited] # triggered whenever a new GitHub release is published
```
raghumdani, line 4: What is the motivation for this change?

pfaraone: In the past, we've had to re-create releases/tags if they didn't have a corresponding deltacat version change (see [0.1.18.beta12-2](https://github.com/ray-project/deltacat/releases/tag/0.1.18.beta12-2) and [0.1.18.beta12-1](https://github.com/ray-project/deltacat/releases/tag/0.1.18.beta12 for example) to trigger the workflow properly. The `published` event/activity type is only triggered if the tag and release title are unique. This change will allow us to reuse tags and releases for deltacat versions, even if they don't have a corresponding change to the deltacat version number. We can simply update the tags and releases with the new deltacat version number, and the workflow will still be triggered.

raghumdani (approved):
LGTM

PR #184, opened 2023-08-12
Use list_partition_deltas to discover deltas in compaction

rkenmi:
Updates the delta discovery lookup during compaction by fetching deltas by a given shard ID.


deltacat/compute/compactor/utils/io.py
```diff
@@ -343,16 +343,8 @@ def _discover_deltas(
     deltacat_storage=unimplemented_deltacat_storage,
     **kwargs,
 ) -> List[Delta]:
-    stream_locator = source_partition_locator.stream_locator
-    namespace = stream_locator.namespace
-    table_name = stream_locator.table_name
```
raghumdani, line 348: do you also want to change the interface of `list_partition_deltas`?

raghumdani (commented):
Looks good overall. Can you also modify the module `tests.local_deltacat_storage` accordingly?

raghumdani (approved):
LGTM

Issue #186, opened 2023-08-14
Test compaction session v1 and v2 using local deltacat storage


raghumdani:
#### Solution

We have a local deltacat implementation which can be used to test the entire compute module without relying on any internal resources. This will simplify the migration to compactor v2 and also ensures the quality of new changes to the deltacat.

raghumdani:
This is also part of #150

raghumdani:
Compactor v2 is now part of functional test suite written for V1.

Issue #185, opened 2023-08-14
Implement compaction session v2


raghumdani:
### Problem Statement

The current compaction session uses fixed size cluster and runs hash bucket, dedupe and merge v1. We have a v2 implementation of hash bucket and merge. We also need a compaction session that uses the v2 algo along with the partial file downloads when needed. 

### Solution

- We implement `PartialFileDownloadParams` to indicate the manifest entry must be downloaded partially. 
- We implement task options provider to go away from fixed size cluster although this PR can still work with fixed size cluster.
- More refactoring...

Epic: #150

Issue #181, opened 2023-08-08
Implement merge v2


raghumdani:
#150

raghumdani:
This is now merged.

PR #187, opened 2023-08-14
Compaction session implementation for algo v2

raghumdani:
#185 

Testing
------
- [x] Implement compaction session v2
- [x] Implement the task resource options provider
- [x] Use `pg_config` to run on a fixed size cluster. 
- [ ] #181 
- [ ] Integration tests pending on #186

### DIFF
--- a/deltacat/compute/compactor_v2/utils/io.py
+++ b/deltacat/compute/compactor_v2/utils/io.py
@@ -1,0 +1,178 @@
+import logging
+import functools
+from deltacat.constants import PYARROW_INFLATION_MULTIPLIER
+from deltacat.types.media import ContentType
+from deltacat.storage import (
+    PartitionLocator,
+    Delta,
+    interface as unimplemented_deltacat_storage,
+)
+from deltacat import logs
+from deltacat.compute.compactor.utils import io as io_v1
+from deltacat.compute.compactor import DeltaAnnotated
+from typing import Dict, List, Optional, Any
+from deltacat.compute.compactor_v2.constants import (
+    MIN_FILES_IN_BATCH,
+    MIN_DELTA_BYTES_IN_BATCH,
+)
+from deltacat.compute.compactor.model.compaction_session_audit_info import (
+    CompactionSessionAuditInfo,
+)
+from deltacat.compute.compactor_v2.utils.task_options import (
+    estimate_manifest_entry_size_bytes,
+)
+from deltacat.types.media import TableType
+from deltacat.types.partial_download import PartialParquetParameters
+
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
+
+
+def append_content_type_params(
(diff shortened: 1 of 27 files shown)


deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -0,0 +1,224 @@
...
+                total_pk_size += entry_size
+            else:
+                total_pk_size += pk_size
+
+    # total size + pk size + pk hash column + hash bucket index column
+    # Refer to hash_bucket step for more details.
+    total_memory = size_bytes + total_pk_size + num_rows * 20 + num_rows * 4
+
+    # Consider 20% buffer
+    total_memory = total_memory * 1.2
```
Zyiqin-Miranda, line 140: Could we make the 1.2 buffer as a constant to be more clear and easy to change?

deltacat/compute/compactor_v2/utils/io.py
```diff
@@ -0,0 +1,178 @@
...
+    logger.info(f"Input delta files to compact: {delta_manifest_entries}")
+
+    if not input_da_list:
+        raise RuntimeError("No input deltas to compact!")
+
+    size_estimation_function = functools.partial(
+        func=estimate_manifest_entry_size_bytes, previous_inflation=previous_inflation
+    )
+
+    rebatched_da_list = DeltaAnnotated.rebatch(
```
Zyiqin-Miranda, line 162: So in v2, our default will split the annotated deltas into row group first then do an original rebatch? So in the case of small files, the `row_groups_to_download` will always be 1 so that we will not split on a small file?

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -0,0 +1,224 @@
...
+
+                    if pk_size is None:
+                        pk_size_bytes += current_entry_size
+                    else:
+                        pk_size_bytes += pk_size
+
+    # total data downloaded + primary key hash column + primary key column + dict size for merge
+    total_memory = data_size + pk_size + num_rows * 20
+
+    total_memory = total_memory * 1.2
```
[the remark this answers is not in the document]
raghumdani: 20% is that buffer, hence multiplied it with 1.2.

deltacat/compute/compactor_v2/utils/task_options.py
raghumdani: done.

deltacat/compute/compactor_v2/utils/io.py
raghumdani: Yeah, the lowest split granularity we can go is at the row group level.

Zyiqin-Miranda:
LGTM!

Zyiqin-Miranda (approved):

PR #199, opened 2023-08-29
Configure STS to use us-east-1 regional endpoint

yankevn:
On the driver side, we have set the `AWS_STS_REGIONAL_ENDPOINTS` environment variable to "regional". Per the [documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html), this will automatically infer the correct STS regional endpoint based on the configured AWS region. While we do set `AWS_REGION` in the CDK, also setting the region explicitly will ensure we always use a regional endpoint instead of the global STS endpoint. 

As a reminder, we are interested in only using STS regional endpoints as it will allow us to have better latency and request higher TPS limits.

Testing
- Tested locally with the `AWS_STS_REGIONAL_ENDPOINTS` environment variable set to "regional" and can confirm the correct regional endpoint is being hit depending on the region passed into `boto3.client()`.


deltacat/compute/metastats/stats.py
```diff
@@ -94,7 +94,11 @@ def start_stats_collection(
 
 
 def _get_account_id() -> str:
-    client = client_cache("sts", None)
+    client = client_cache(
+        "sts",
+        region_name="us-east-1",
+        endpoint_url="https://sts.us-east-1.amazonaws.com",
```
raghumdani, line 100: What would happen if the environment is not setup in us-east-1? The region and endpoint must be picked from environment variable.

yankevn: The most recent change will first check the environment variable before setting the region and endpoint url. In the event that a node is not in us-east-1 and does not have the environment variable set, [they will still be able to use the us-east-1 endpoint](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html), although just at higher latency.

deltacat/aws/constants.py
```diff
@@ -1,6 +1,10 @@
 from typing import List
 
-from deltacat.utils.common import env_integer
+from deltacat.utils.common import env_integer, env_string
 
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 15)
 TIMEOUT_ERROR_CODES: List[str] = ["ReadTimeoutError", "ConnectTimeoutError"]
+AWS_REGION = env_string("AWS_REGION", "us-east-1")
+STS_REGIONAL_ENDPOINT = env_string(
```
raghumdani, line 8: This need not be env variable as it can be derived from the `AWS_REGION`.

raghumdani: Use this `AWS_STS_REGIONAL_ENDPOINTS` in config.

yankevn:
@raghumdani description has been updated 👍

raghumdani (approved):
LGTM

PR #200, opened 2023-08-30
Fix bug where s3_table_writer_kwargs is not passed into repartition

yankevn:
We have a bug where our repartition function is not passing in any writer kwargs to stage_delta, whereas our materialize function is.


raghumdani:
Looks good. How did you test this?

raghumdani (approved):
LGTM! I will be testing this.

PR #202, opened 2023-08-31
Workaround for pydantic 2.0 incompatibility, new AWS glue job runner, and new makefile targets.

pdames:
This pull request:
1. Refactors `requirements.txt`, `setup.py`, and `dev-requirements.txt` to get `make test-integration` working again after PyIceberg's migration to use Pydantic v2 (see https://github.com/apache/iceberg/pull/7782 and https://github.com/ray-project/ray/issues/37019 for details).
2. Introduces new `make` targets to separate `build` artifact creation/maintenance from `venv` creation/maintenance. Also introduces appropriate dependencies between targets so that you don't have to know the correct order to run targets in anymore.
3. Introduces `glue_runner.py` to provide rapid feedback for cyclical distributed development workflows that follow steps like: (a) run `$ python glue_runner.py deltacat/examples/hello_world.py --deploy-local-deltacat` (b) {change deltacat/examples/hello_world.py and any other DeltaCAT files} (c) return to step [a]
4. Has a lot of additional lines of code and files changed due to also including a rebase of the `iceberg` branch on top of deltacat `main` branch. =(
We may want to break these out into separate PRs but, for now, the actual changed files to pay attention to are `requirements.txt`, `dev-requirements.txt`, `setup.py`, `glue_runner.py`, `Makefile`, `s3-build-and-deploy.sh`, `deltacat/examples/hello_world.py`, `deltacat/examples/logging.py`.

Tests run:
1. `$ make test-integration` (passing with latest Iceberg `master` commit and `ray[data] == 2.6`)
2. `$ python glue_runner.py deltacat/examples/hello_world.py`
3. `$ python glue_runner.py deltacat/examples/logging.py`

Glue runner tests were run with various permutations of supported arguments against local and PyPi builds of DeltaCAT.

Follow-up Issues to create:
1. **Makefile Developer Documentation**: Update `README-development.md` to include instructions on how to best use the new `make` targets and resolve common problems (e.g. `make clean` -> `make build` | `make test-integration` solves most dirty workspace problems).
2. **Revert workarounds for PyIceberg+Ray+Pydantic v2 Incompatibilities**: Once Ray 2.8 is released (or at least a nightly wheel is published that resolves the Pydantic v2 conflict) we need to (1) remove `pyiceberg`, `pydantic`, and `ray[data]` overrides from `dev-requirements.txt` and (2) update the corresponding lines for each of these dependencies in `requirements.txt` and `setup.py` to settle on either a single version or set of backwards-compatible versions.
3. **DeltaCAT+Iceberg+Glue Read & Transform Example**: Introduce a self-contained AWS Glue example script to setup an Iceberg catalog, read an Iceberg table, and run some simple distributed transforms on its data.
5. **AWS Glue Runner User Guide**: Introduce dedicated documentation guiding users through various AWS Glue Runner usage patterns.
6. **DeltaCAT+Daft+Iceberg+Glue Read & Transform Support + Example**: Refactor DeltaCAT's `DistributedDataset` to refer to either a Ray Dataset or a Daft Dataframe, add the ability to read Iceberg tables into Daft dataframes natively (no intermediate conversion through a Ray Dataset), and run some simple distributed transform on its data within an AWS Glue example script.
7. **CI/CD Workflows**: Introduce CI/CD workflows that run distributed integration tests to protect against regressions between DeltaCAT+RayData+AWS Glue+Daft+Iceberg integration points.
8. **Support running latest PyIceberg Commit on AWS Glue for Ray**: Investigate best ways to install latest PyIceberg from Apache Iceberg `master` branch inside of an AWS Glue job run. One possible method may involve using a command like `pip install "https://github.com/apache/iceberg/archive/refs/heads/master.zip#subdirectory=python&egg=pyiceberg[s3fs]"` once python is included in the `master.zip` artifact or extracting this zip within an Glue Job then configuring the Ray Runtime Environment to use the `python` subdirectory as the `working_dir`.
9. **Merge Iceberg Branch into Main Branch**: After we take care of PydanticV2 compatibility problems, we should consider merging out work here back into the `main` branch of DeltaCAT to prevent ongoing merges/rebases between the two (with merges effectively required once other developer dependencies are taken on the `iceberg` branch).

Some things I'm looking for feedback on:
1. AWS Glue Ray runtime environment installation for DeltaCAT takes longer than expected... a simple end-to-end job run that just prints a few lines to the console and exits now takes ~1.5-2 minutes here. Any suggestions, ideas to reduce this are welcome.
2. A command like `aws logs tail /aws-glue/ray/jobs/script-log --follow` prints a lot of duplicate log lines on my machine (2021 Apple M1 Pro with AWS CLI version: aws-cli/2.13.3 Python/3.11.4 Darwin/22.5.0 source/arm64) - curious if others see the same thing.
3. One-time IAM configuration for running AWS Glue jobs can be a bit tricky/finicky to configure correctly... in my case, I had to add full CloudWatch permissions to resolve a persistent internal server error in my AWS account here, so again curious if others see similar problems.


pdames:
@jaychia @samster25 FYI! Feel free to review & try out the new AWS Glue Runner here. :)
@chappidim @jjacobj84 Please review and let me know what you think about the Glue integration. Specifically, looking for feedback about how to get the log lines printed by `logging_worker` printed out to the `/aws-glue/ray/jobs/ray-worker-out-logs` log group inside of `deltacat/examples/logging.py` since this hasn't been working so far (only plain `print` statements show up in the worker standard out logs).

jjacobj84:
> @jaychia @samster25 FYI! Feel free to review & try out the new AWS Glue Runner here. :) @chappidim @jjacobj84 Please review and let me know what you think about the Glue integration. Specifically, looking for feedback about how to get the log lines printed by `logging_worker` printed out to the `/aws-glue/ray/jobs/ray-worker-out-logs` log group inside of `deltacat/examples/logging.py` since this hasn't been working so far (only plain `print` statements show up in the worker standard out logs).

@pdames can you point me to a JobRun ID mainly to understand 1) higher install time of delta cat 2) repeated log line printed issue. Now wrt log lines from `deltacat/examples/logging.py` missing in `/aws-glue/ray/jobs/ray-worker-out-logs`  is because the CW configuration that we have in Glue Ray for log group `ray-worker-out-logs` reads from log files at  `/tmp/ray/session_latest/logs/worker-**.out`. The logger used by deltacat by default writes to `/tmp/deltacat/var/output/logs/`. Would it be possible to configure the logger to write to `/tmp/ray/session_latest/logs/worker-**.out`

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+        JobName=glue_job_name,
+    )
+    job_runs = response["JobRuns"]
+    last_job_run_args = (
+        job_runs[0]["Arguments"] if job_runs and "Arguments" in job_runs[0] else {}
+    )
+    print(f"Last Job Run Args for {glue_job_name}: {last_job_run_args}\n\n")
+
+    cli_args = args.example_script_args
+    last_job_run_packages = last_job_run_args.get("--PACKAGE")
```
chappidim, line 225: How is this different from pip-install? Wondering if we are double installing? 1/pip-install 2/runtime_env using `PACKAGE` (based on fixture.py example).

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+            Name=glue_job_name,
+            **glue_job_config,
+        )
+    print(f"New Glue Job Config: {glue_job_config}")
+
+    print(f"Getting Last Job Run Arguments for {glue_job_name}")
+    response = glue_client.get_job_runs(
+        JobName=glue_job_name,
+    )
+    job_runs = response["JobRuns"]
```
jjacobj84, line 218: would validate response for `JobRuns` key before trying to access. It could be quite possible the first iteration of a new Job does not have any Job runs.

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+    if dc_url:
+        new_package = f"deltacat @ {dc_url}"
+    elif dc_version:
+        new_package = f"deltacat=={dc_version}"
+    elif not (last_job_run_packages and last_job_run_pip):
+        # install the latest version of deltacat
+        new_package = "deltacat"
+    if new_package:
+        glue_start_job_run_args["--pip-install"] = new_package
+        glue_start_job_run_args["--PACKAGE"] = new_package
```
jjacobj84, line 252: Can you help me understand why we would need `--pip-install` as well as `--PACKAGE`. I believe the later is a Job parameter but curious to know do we try to install dependencies in the script again.

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+        returncode = proc.poll()
+        if returncode != 0:
+            raise RuntimeError(f"Command {cmd} Failed. Exit Code: {returncode}")
+
+    s3_client = boto3.client("s3")
+
+    # upload the example runner script to S3
+    bucket_name = f"deltacat-glue-scripts-{stage}"
+    print(f"Ensuring Glue Job Script Bucket exists: {bucket_name}")
+    s3_client.create_bucket(Bucket=bucket_name)
```
raghumdani, line 159: `LocationConstraint` must be added to create buckets in regions other than `us-east-1`. Ref: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/create-bucket.html

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+    type=int,
+)
+parser.add_argument(
+    "-n",
+    "--glue-job-name",
+    help="Name of the Glue Job to create/update (default: deltacat-example-runner-$USER).",
+    type=str,
+)
+args_mutex = parser.add_mutually_exclusive_group()
+args_mutex.add_argument(
```
raghumdani, line 87: Does this script also handle deploying the custom implementation of deltacat storage interface? Because, without that all APIs yield `NotImplementedError` correct?

s3-build-and-deploy.sh
```diff
@@ -0,0 +1,79 @@
...
+      info "copying $file to $UNIQUE_WHEEL"
+      cp "$file" "$UNIQUE_WHEEL"
+    else
+      barf "ERROR: multiple deltacat wheels found: ${ls dist/deltacat-*.whl}"
+      exit 1
+    fi
+done
+# upload the new deltacat wheel to s3 and create a presigned URL for it
+info "uploading $UNIQUE_WHEEL to $S3_PACKAGE_BUCKET_URL"
+aws s3 cp "$UNIQUE_WHEEL" "$S3_PACKAGE_BUCKET_URL"
```
raghumdani, line 71: This script assumes that user's environment has aws cli installed but doesn't look like Makefile has it. Can you confirm?

dev-requirements.txt
```diff
@@ -8,3 +8,6 @@ pre-commit == 2.20.0
 pytest == 7.2.0
 pytest-cov == 4.0.0
 requests-mock == 1.11.0
+pyiceberg @ git+https://github.com/apache/iceberg#subdirectory=python
+pydantic == 2.0
+ray[data] ~= 2.6
```
raghumdani, line 13: `ray[data]` has conflicting versions in `dev-requirements` and `requirements.txt`. Was it intended?

deltacat/examples/common/fixtures.py
```diff
@@ -0,0 +1,54 @@
...
+            "DELTACAT_SYS_DEBUG_LOG_BASE_FILE_NAME": DELTACAT_SYS_DEBUG_LOG_BASE_FILE_NAME,
+        }
+
+        # setup runtime environment from system environment variables:
+        #   1. STAGE: The execution environment stage (e.g. alpha, beta, dev, etc.)
+        #   2. PACKAGE: Package to Pip Install. (e.g. signed S3 URL for DeltaCAT)
+        runtime_environment = {
+            "env_vars": worker_env_vars,
+            "pip": package,
+        }
```
raghumdani, line 53: Minor: Can these be part of glue job itself? Do we need to create a separate runtime env within the job?

raghumdani (commented):
Looks good overall. Is the runtime 1.5 - 2mins with pip installing deltacat version from PyPI or from custom S3 URL?

pdames:
> > @jaychia @samster25 FYI! Feel free to review & try out the new AWS Glue Runner here. :) @chappidim @jjacobj84 Please review and let me know what you think about the Glue integration. Specifically, looking for feedback about how to get the log lines printed by `logging_worker` printed out to the `/aws-glue/ray/jobs/ray-worker-out-logs` log group inside of `deltacat/examples/logging.py` since this hasn't been working so far (only plain `print` statements show up in the worker standard out logs).
> 
> @pdames can you point me to a JobRun ID mainly to understand 1) higher install time of delta cat 2) repeated log line printed issue. Now wrt log lines from `deltacat/examples/logging.py` missing in `/aws-glue/ray/jobs/ray-worker-out-logs` is because the CW configuration that we have in Glue Ray for log group `ray-worker-out-logs` reads from log files at `/tmp/ray/session_latest/logs/worker-**.out`. The logger used by deltacat by default writes to `/tmp/deltacat/var/output/logs/`. Would it be possible to configure the logger to write to `/tmp/ray/session_latest/logs/worker-**.out`

@jjacobj84 Thanks for the info about the worker log file path - I've confirmed that both the driver and workers will pick up logs written to this path in the `/aws-glue/ray/jobs/ray-worker-out-logs` CloudWatch log group. However, I noticed that I also need to initialize a dedicated logger inside of each worker process for this to work correctly, which remains a minor con.

Job runs are still taking quite a while to install the runtime environment, here are the logs from my last job run (including embedded job run ID) which shows the time taken to complete the runtime environment setup as ~1m43s:
```
2023-09-01T04:10:17.155000+00:00 deltacat-example-runner-pdames-jr_82256a2a4fde372dc76572026451ae0166cf488587e35859405c20b1bcc0cbd3 2023-09-01 04:10:12,374 - __main__ - INFO - Setting up ray runtime environment to install custom lib
2023-09-01T04:11:55.429000+00:00 deltacat-example-runner-pdames-jr_82256a2a4fde372dc76572026451ae0166cf488587e35859405c20b1bcc0cbd3 2023-09-01 04:11:55,102 - __main__ - INFO - Successfully set up ray runtime environment installing custom lib
```

deltacat/examples/common/fixtures.py
pdames: We can now get rid of `pip` (see conversations above), but don't know of any good way to get rid of the environment variables, which are currently used to configure our loggers appropriately during job run execution. @jjacobj84 @chappidim - please correct me if there's a better way to pass environment variables out to all workers in the cluster.

glue_runner.py
pdames: Good catch - will fix this to remove the assumption that job runs exist.

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+        "glue",
+        region,
+    )
+    try:
+        response = glue_client.get_job(JobName=glue_job_name)
+        print(f"Found Existing Glue Job `{glue_job_name}`: {response}")
+        print(f"Updating Existing Glue Job `{glue_job_name}`")
+        if not glue_iam_role:
+            glue_job_config["Role"] = response["Job"]["Role"]
+        glue_client.update_job(
```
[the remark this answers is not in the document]
pdames: Was kind of thinking the same about this code block - I'll double-check the response here and add some extra validation.

pdames:
> Looks good overall. Is the runtime 1.5 - 2mins with pip installing deltacat version from PyPI or from custom S3 URL?

Doesn't seem to make much difference (they're both slow). The log lines I gave to @jjacobj84 were from a custom S3 URL, and these are from an install from PyPi (0.1.18b14) which took 2m45s:

```
2023-09-01T05:17:24.428000+00:00 deltacat-example-runner-pdames-jr_2cff5bf30b7e0489a03714fcfca0de3c4135707c061b31d7df1d460b6376c76c 2023-09-01 05:17:19,604 - __main__ - INFO - Setting up ray runtime environment to install custom lib
2023-09-01T05:20:09.840000+00:00 deltacat-example-runner-pdames-jr_2cff5bf30b7e0489a03714fcfca0de3c4135707c061b31d7df1d460b6376c76c 2023-09-01 05:20:09,348 - __main__ - INFO - Successfully set up ray runtime environment installing custom lib
```

deltacat/examples/common/fixtures.py
raghumdani: AFAIK any non-reserved job parameters become env vars. @jjacobj84 @chappidim Please confirm.

pdames:
Alright, I think all existing questions/comments/issues are resolved by the latest push. @raghumdani @jjacobj84 @chappidim  Any additional questions or concerns before merging back to the `iceberg` branch?

raghumdani (approved):
Looks good!

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+        bucket_name,
+        run_script_path,
+    )
+    print(f"Uploaded Glue Job Script to: {s3_glue_script_file_path}")
+    glue_max_workers = args.glue_max_workers
+    glue_iam_role = args.glue_iam_role
+    glue_job_config = {
+        "Role": glue_iam_role,
+        "GlueVersion": "4.0",
+        "ExecutionProperty": {"MaxConcurrentRuns": 10},
```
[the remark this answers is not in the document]
pdames: Changed this to 2 max concurrent runs to help mitigate the risk of exceeding DPU limits for now.

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+                break
+            line = line.strip()
+            if line.find("DELTACAT_PRESIGNED_URL=") != -1:
+                dc_url = line.split("=", 1)[1]
+        returncode = proc.poll()
+        if returncode != 0:
+            raise RuntimeError(f"Command {cmd} Failed. Exit Code: {returncode}")
+
+    s3_client = boto3.client("s3")
+
```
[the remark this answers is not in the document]
pdames: Good catch - added inline here and other locations!

glue_runner.py
pdames: This was an old workaround to ensure that the runtime environment was installed on all cluster nodes (including the head node) that was needed in the earlier days of the AWS Glue for Ray Preview release, but is no longer needed today. Removed (doesn't seem to impact long 1-3 minute runtime environment setup times much though, unfortunately).

glue_runner.py
```diff
@@ -0,0 +1,276 @@
...
+    last_job_run_packages = last_job_run_args.get("--PACKAGE")
+    last_job_run_pip = last_job_run_args.get("--pip-install")
+    glue_start_job_run_args = {
+        "--PACKAGE": last_job_run_packages,
+        "--pip-install": last_job_run_pip,
+        "--STAGE": stage,
+        # set deltacat logging environment variables to be picked up by Glue
+        "--DELTACAT_APP_LOG_LEVEL": "DEBUG",
+        "--DELTACAT_SYS_LOG_LEVEL": "DEBUG",
+        "--DELTACAT_APP_LOG_DIR": "/home/glue-ray-engine/var/log/",
```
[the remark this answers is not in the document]
pdames: Done! All logs from the driver and workers are now consolidate in Ray worker logs, and the one bit of weirdness here still is that each worker task logger needs to be initialized within the worker process it's logging for (can't reuse a globally scoped logger across the driver and workers on AWS Glue for Ray, but there doesn't seem to be an equivalent problem in Ray on EC2).

glue_runner.py
pdames: The separate `--PACKAGE` arg is no longer needed, and has been removed (see related discussion with @chappidim above).

glue_runner.py
pdames: Good catch - added!

glue_runner.py
pdames: This script will handle deploying implementations as they're made available in subsequent PRs, but that's correct - all storage APIs yield `NotImplementedError` with this PR merged today.

s3-build-and-deploy.sh
pdames: The AWS CLI needs to be installed as a separate development environment setup step. The process here will be documented (i.e. pulled up from the `glue_runner.py` script into `README-development.md`) in a follow-up PR to update developer documentation.

dev-requirements.txt
pdames: Yes, this is intentional for now (see follow-up issue noted in description to revert this behavior later once Ray creates an official release that is compatible with Pydantic 2.x). The different dependency versions stated in `dev-requirements`  intentionally override `requirements` so that `make test-integration` passes by using `dev-requirements` while AWS Glue Job Runs continue to use the older stable versions given in `requirements` (but this may change in a subsequent PR as we start introducing integration tests for Iceberg on AWS Glue).

deltacat/examples/common/fixtures.py
pdames: They do become environment variables, but these environment variables only appear to be propagated to the head node by default (you need to manually propagate them to worker nodes like this, unless there's a separate argument to set or this behavior is considered a bug on the AWS Glue side).

raghumdani: Got it, seems like a bug. Don't see any reason for this to be intentional.

Issue #205, opened 2023-09-01
Gracefully handle empty CSV


raghumdani:
Today, empty CSV can cause the read to fail causing continuous retries during reads. Hence, it would be ideal to gracefully handle this and create an empty pyarrow table instead of failing during the reads. An empty CSV is a valid scenario in our internal catalog and the compactor should write back empty files if the compactor produces no records to indicate completeness.

PR #206, opened 2023-09-01
Gracefully handle empty CSV

raghumdani:
#205

### Testing

Unit tests added.


deltacat/tests/utils/data/non_empty_valid.csv
```diff
@@ -0,0 +1,3 @@
+Y	1754-08-30T22:43:41
+""	2022-08-10T22:43:21
+N	2022-08-10T22:43:21
```
yankevn, line 3: Should we also test `null` (truly empty values), timestamps that are far in the future, and timestamps with decimal values?

deltacat/utils/pyarrow.py
```diff
@@ -33,15 +33,74 @@
...
+    for column_name in columns:
+        index = schema.get_field_index(column_name)
+
+        if index != -1:
+            target_field = schema.field(index)
+            target_schema_fields.append(target_field)
+
+    target_schema = pa.schema(target_schema_fields, metadata=schema.metadata)
+
+    return target_schema
```
yankevn, line 57: So what exactly is this function doing? It might be good to add a docstring. Why would we want to remove columns from the schema?

deltacat/tests/utils/data/non_empty_valid.csv
raghumdani: Formatter used by DeltaCAT is cleaning it up. Hence, didn't include it, but it pyarrow related functionality.

deltacat/utils/pyarrow.py
raghumdani: Creates a new schema that adheres to the given column name ordering and types.

deltacat/tests/utils/data/non_empty_valid.csv
yankevn: Is that just for nulls? Is that only if the first field/second field is null, or it doesn't matter and the formatter always complains if any field is null? Could we still test timestamps like `9999-08-10T22:43:21.123456`

raghumdani: Sure, added year 9999.

deltacat/utils/pyarrow.py
```diff
@@ -33,15 +33,74 @@
...
+
+
+def pyarrow_read_csv(*args, **kwargs) -> pa.Table:
+    try:
+        kwargs = sanitize_kwargs_by_supported_kwargs(
+            ["read_options", "parse_options", "convert_options", "memory_pool"], kwargs
+        )
+        return pacsv.read_csv(*args, **kwargs)
+    except pa.lib.ArrowInvalid as e:
+        if e.__str__() == "Empty CSV file":
```
rkenmi, line 67: This is a workaround specific to quirks of our own internal data catalog, which ideally shouldn't be producing empty CSVs anyway. As a general solution, perhaps its better to provide an option to disable/enable this functionality if desired?

raghumdani: Good point.

rkenmi (approved):

PR #209, opened 2023-09-05
Capture memory gb seconds and bug fixes

raghumdani:
- [x] fix delta split
- [x] added inflation to the parquet size
- [x] added more logging
- [x] change storage_type to LOCAL as ParquetFile cannot be pickled
- [x] increase the delay for blocking instance metadata service
- [x] capture total and used memory gb seconds to understand memory efficiency
- [x] add ray_custom_resources compaction param to allow scheduling to particular node types

### DIFF
--- a/deltacat/compute/compactor/model/delta_annotated.py
+++ b/deltacat/compute/compactor/model/delta_annotated.py
@@ -287,7 +287,8 @@ class DeltaAnnotated(Delta):
                         result.append(new_da)
                 else:
                     logger.info(
-                        f"Split was not performed on correct delta with locator: {delta_annotated.locator}"
+                        f"Split was not performed on delta with locator: {delta_annotated.locator} "
+                        "as partial parquet params was not found."
                     )
                     return [delta_annotated]
 
--- a/deltacat/compute/compactor/model/delta_annotated.py
+++ b/deltacat/compute/compactor/model/delta_annotated.py
@@ -286,7 +286,13 @@ class DeltaAnnotated(Delta):
 
                         result.append(new_da)
                 else:
+                    logger.info(
+                        f"Split was not performed on correct delta with locator: {delta_annotated.locator}"
+                    )
                     return [delta_annotated]
+
+        if result:
+            return result
 
         logger.info(
             f"Split was not performed on the delta with locator: {delta_annotated.locator}"
--- a/deltacat/compute/compactor_v2/constants.py
+++ b/deltacat/compute/compactor_v2/constants.py
@@ -23,10 +23,10 @@ AVERAGE_RECORD_SIZE_BYTES = 1000
 # r5.8xlarge EC2 instances.
 TASK_MAX_PARALLELISM = 5367
 
-# The percentage of memory that needs to be estimated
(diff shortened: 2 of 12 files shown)


deltacat/compute/compactor_v2/constants.py
```diff
@@ -35,3 +35,8 @@
 
 # Whether to drop duplicates during merge.
 DROP_DUPLICATES = True
+
+# PARQUET to PYARROW inflation multiplier
+# This is the observed upper bound inflation for parquet
+# size in metadata to pyarrow table size.
+PARQUET_TO_PYARROW_INFLATION = 4
```
Zyiqin-Miranda, line 42: This PARQUET_TO_PYARROW_INFLATION is specific to reading parquet file?

deltacat/utils/resources.py
```diff
@@ -104,14 +107,28 @@ def _update_vcpus(self):
...
                 str(cluster_resources["CPU"] - available_resources["CPU"])
             )
-            self.total_vcpu_seconds = self.total_vcpu_seconds + float(
-                str(cluster_resources["CPU"])
-            )
-        else:
-            self.total_vcpu_seconds = self.total_vcpu_seconds + float(
-                str(cluster_resources["CPU"])
+
+        self.total_vcpu_seconds = self.total_vcpu_seconds + float(
```
Zyiqin-Miranda, line 111: Should this be inside the above if ` if "CPU" not in cluster_resources`?

deltacat/compute/compactor/model/delta_annotated.py
```diff
@@ -286,8 +286,14 @@ def _split_single(delta_annotated: DeltaAnnotated) -> List[DeltaAnnotated]:
 
                         result.append(new_da)
                 else:
+                    logger.info(
+                        f"Split was not performed on correct delta with locator: {delta_annotated.locator}"
```
Zyiqin-Miranda, line 290: What does `correct delta` refer to here?

Zyiqin-Miranda (approved):

deltacat/compute/compactor_v2/constants.py
raghumdani: Yes.

deltacat/compute/compactor/model/delta_annotated.py
raghumdani: I wanted this to be different log message than the below one. Changing this to more meaningful.

deltacat/utils/resources.py
raghumdani: I didn't get this comment.

Zyiqin-Miranda: Is the indention here correct? Will this fail if cluster_resources don't have `"CPU"` field?

raghumdani: Yes, it's correct. The above code will return if "CPU" is not in the list:

```
        if "CPU" not in cluster_resources:
            return
```

Zyiqin-Miranda: I see, L 104-106 was hidden by the UI

PR #210, opened 2023-09-06
update daft to 0.1.16 that uses openssl and correct arrow schema parsing with int96 options

samster25:
* Updates the daft version to 0.1.16 which switches daft over to use OpenSSL over Rustls
* Also includes correct behavior when parsing a parquet schema with int96 options which the `ARROW_SCHEMA` field is also set


raghumdani (approved):
LGTM

PR #213, opened 2023-09-07
Daft Parquet Reader - convert large strings to regular pyarrow strings

samster25:
* Converts large strings to regular strings when exporting a daft table to pyarrow


raghumdani:
If this needs to work, don't we need to return chunked data instead as we discussion over slack?

Issue #150, opened 2023-07-06
[EPIC] Implement a version of compactor that writes files partitioned by pk hash


raghumdani:
As part of this story, we would like to evaluate how materializing the files partitioned by primary key hash of the records it contains will impact efficiency. We will accomplish below tasks as part of this story:

1. Hash bucket to read all files (or only new deltas if destination already materialized in previous round)
2. Combine and dedupe and materialize tasks to run as a single process (though code modularity is still preferred).

PR #214, opened 2023-09-07
Adding compactor version in audit info and rci

raghumdani:
#150 

This commit adds the compactor version in round completion file and audit to allow looking up what version was used and also allows driver to programmatically lookup version to decide whether to run rebase or incremental compaction. E.g., if version changes, a rebase has to be run.

### DIFF
--- a/deltacat/compute/compactor/model/compactor_version.py
+++ b/deltacat/compute/compactor/model/compactor_version.py
@@ -1,6 +1,6 @@
 from enum import Enum
 
 
-class CompactorVersion(Enum):
+class CompactorVersion(str, Enum):
     V1 = "V1"
     V2 = "V2"
--- a/deltacat/compute/compactor/model/compactor_version.py
+++ b/deltacat/compute/compactor/model/compactor_version.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+
+class CompactorVersion(Enum):
+    V1 = "V1"
+    V2 = "V2"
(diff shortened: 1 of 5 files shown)


deltacat/compute/compactor/model/compactor_version.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
+
+
+class CompactorVersion(Enum):
```
pfaraone, line 4: Can we make this a `(str, Enum)` and then we can make comparisons without `.value` method

raghumdani: Good point.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -114,6 +115,7 @@ def _execute_compaction(
     compaction_audit = CompactionSessionAuditInfo(deltacat.__version__, audit_url)
 
     compaction_audit.set_hash_bucket_count(params.hash_bucket_count)
```
pfaraone, line 117: Nit: Could we inline this as `compaction_audit.set_hash_bucket_count(params.hash_bucket_count).set_compactor_version(CompactorVersion.V2.value)` since we are using a builder pattern for the `CompactionSessionAuditInfo` object

pfaraone:
A few small comments but otherwise LGTM

deltacat/compute/compactor_v2/compaction_session.py
raghumdani: Sure.

pfaraone (approved):
LGTM

PR #215, opened 2023-09-11
Add index array size to estimation, fix dedupe and concurrency

raghumdani:
#150 

- [x] Incremental index array size estimation. 
- [x] numpy + operator was appending elements instead of adding due to correct shape inference when all chunks are of same length. 
- [x] Reduce the ray.wait time complexity from O(N^2) to O(N). 
- [x] Set fetch_local=False so that ray.wait can return without downloading the object to node for performance.

### DIFF
--- a/deltacat/compute/compactor_v2/utils/task_options.py
+++ b/deltacat/compute/compactor_v2/utils/task_options.py
@@ -232,7 +232,8 @@ def merge_resource_options_provider(
                     else:
                         pk_size_bytes += pk_size
 
-    # total data downloaded + primary key hash column + primary key column + dict size for merge
+    # total data downloaded + primary key hash column + primary key column
+    # + dict size for merge + incremental index array size
     total_memory = (
         data_size
         + pk_size_bytes
--- a/deltacat/compute/compactor_v2/utils/task_options.py
+++ b/deltacat/compute/compactor_v2/utils/task_options.py
@@ -179,7 +179,9 @@ def merge_resource_options_provider(
     data_size = hash_group_size_bytes.get(hb_group_idx, 0)
     num_rows = hash_group_num_rows.get(hb_group_idx, 0)
 
-    pk_size_bytes = 0
+    # upper bound for pk size of incremental
+    pk_size_bytes = data_size
+    incremental_index_array_size = num_rows * 4
 
     if (
         round_completion_info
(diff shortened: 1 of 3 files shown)


deltacat/utils/ray_utils/concurrency.py
```diff
@@ -49,20 +48,14 @@ def invoke_parallel(
...
-                ray.wait(
-                    list(itertools.chain(*pending_ids)),
-                    num_returns=int(
-                        len(pending_ids[0]) * (len(pending_ids) - max_parallelism)
-                    ),
-                )
-            else:
-                ray.wait(pending_ids, num_returns=len(pending_ids) - max_parallelism)
+        if max_parallelism is not None and len(remaining_refs) > max_parallelism:
+            _, remaining_refs = ray.wait(
```
rkenmi, line 55: Do we need both `pending_ids` and `remaining_refs`?

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -231,7 +233,13 @@ def merge_resource_options_provider(
                         pk_size_bytes += pk_size
 
     # total data downloaded + primary key hash column + primary key column + dict size for merge
```
rkenmi, line 235: Can we update the comment here accordingly?

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -179,7 +179,9 @@ def merge_resource_options_provider(
     data_size = hash_group_size_bytes.get(hb_group_idx, 0)
     num_rows = hash_group_num_rows.get(hb_group_idx, 0)
 
-    pk_size_bytes = 0
+    # upper bound for pk size of incremental
+    pk_size_bytes = data_size
+    incremental_index_array_size = num_rows * 4
```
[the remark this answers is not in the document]
raghumdani: size of int32 is 4 bytes. We create an index array of length num_rows.

deltacat/utils/ray_utils/concurrency.py
raghumdani: Yeah, because pending_ids is what we return.

deltacat/compute/compactor_v2/utils/task_options.py
raghumdani: done.

deltacat/utils/ray_utils/concurrency.py
rkenmi: To clarify, why not just re-use pending_ids for `ray.wait()`?

raghumdani: the remaining_ids list become empty at the end. Hence, we cannot return it as a result.

rkenmi (approved):

PR #216, opened 2023-09-11
Write pa write result even when referencing

raghumdani:
#150 

This commit aims to fix the issue where pa_write_result was not persisted for materialize result.


pfaraone:
LGTM

pfaraone (commented):
LGTM

pfaraone (approved):
Thanks for the changes

PR #218, opened 2023-09-13
Use `coerce_pyarrow_table_to_schema` in pyarrow download

jaychia:
This PR shares `coerce_pyarrow_table_to_schema` across `pyarrow.py` and `daft.py` utility files

This utility will apply a provided schema (`input_schema`) to a table that is read from a file. This application has the following properties, as documented in the function docstring:

```
    1. For each field in `pa_table`, cast it to the field in `input_schema` if one with a matching name
        is available
    2. Reorder the fields in the casted table to the supplied schema, dropping any fields in `pa_table`
        that do not exist in the supplied schema
    3. If any fields in the supplied schema are not present, add a null array of the correct type
```

Note that this is a departure from the current behavior in `pyarrow.py`, which leverages the `_get_compatible_target_schema` function (removed in this PR) and only performs `Step (1)` of the above steps.

Concretely this means that the current behavior will produce tables that have a different fields than the provided schema. This PR ensures that the tables produced will have exactly the same schema as the provided schema.

### DIFF
--- a/deltacat/tests/utils/test_pyarrow.py
+++ b/deltacat/tests/utils/test_pyarrow.py
@@ -56,6 +56,45 @@ class TestS3PartialParquetFileToTable(TestCase):
         partial_parquet_params.row_groups_to_download.pop()
 
         schema = pa.schema(
+            [
+                pa.field("n_legs", pa.string()),
+                pa.field("animal", pa.string()),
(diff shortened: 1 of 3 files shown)


jaychia:
NOTE: This is a follow-on for `coerce_pyarrow_table_to_schema ` introduced in #217

raghumdani (approved):
LGTM! Thanks for this PR.

PR #219, opened 2023-09-16
Derive optional metric configurations from metrics_config

yankevn:
The current implementation explicitly sets metric dimensions and namespace. These configurations are not necessarily desired by everybody, and thus should be derived from the passed-in metrics_config object instead. 

This change will "break" existing metrics, thus necessitating a minor package version bump.

Testing
 - Changes were deployed locally and tested to ensure that both passing in and omitting namespace and dimensions worked properly


raghumdani (approved):
LGTM

PR #220, opened 2023-09-18
Persisting previous delta inflation and avg record sizes

raghumdani:
#150 

As part of this commit, we will persist input inflation and input average record sizes in round completion file. This will allow us to use it to infer sizes of upcoming incremental deltas. Though we have sizes of compacted, we don't have sizes for incremental delta. Hence, this approach has been found to provide most accurate figures instead of parquet uncompressed size in the parquet metadata. Moreover, this solution is generic and applies to any content type.

### DIFF
--- a/deltacat/compute/compactor_v2/compaction_session.py
+++ b/deltacat/compute/compactor_v2/compaction_session.py
@@ -130,7 +130,6 @@ def _execute_compaction(
         cluster_cpus = cluster_resources["CPU"]
         cluster_memory = cluster_resources["memory"]
         task_max_parallelism = cluster_cpus
-        compaction_audit.set_cluster_cpu_max(cluster_cpus)
         compaction_audit.set_total_cluster_memory_bytes(cluster_memory)
 
     # read the results from any previously completed compaction round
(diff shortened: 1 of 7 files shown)


pfaraone:
LGTMs

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -505,6 +530,8 @@ def merge_input_provider(index, item):
         hash_bucket_count=params.hash_bucket_count,
         hb_index_to_entry_range=hb_id_to_entry_indices_range,
         compactor_version=CompactorVersion.V2.value,
+        input_inflation=input_inflation,
```
rkenmi, line 533: Should we also include the output (materialized) inflation / average record sizes?

rkenmi (approved):

deltacat/compute/compactor_v2/compaction_session.py
raghumdani: I did not include it because that can be derived from pyarrow write result.

PR #221, opened 2023-09-18
Version bumping s3fs to match aiobotocore~=2.4.2

pfaraone:


requirements.txt
```diff
@@ -9,7 +9,7 @@ pydantic == 1.10.4
 pymemcache == 4.0.0
 ray[default] ~= 2.0
 redis == 4.6.0
-s3fs == 2022.2.0
+s3fs == 2023.1.0
```
raghumdani, line 12: Did you run compaction to verify there isn't a regression?

dev-requirements.txt
```diff
@@ -5,6 +5,6 @@ isort == 5.10.1
 memray == 1.6.0; platform_system != "Windows" and sys_platform != "darwin" and platform_machine != "arm64" and platform_machine != "aarch64"
 moto==4.1.12
 pre-commit == 2.20.0
-pytest == 7.2.0
-pytest-cov == 4.0.0
+pytest ~= 7.4.2
```
raghumdani, line 8: Why are these changes?

pfaraone: Rolling back - was seeing if there was a latency improvement in the testing stage of CI by bumping up the version

PR #223, opened 2023-09-18
Add functionality for emitting metrics through CloudWatch EMF

yankevn:
CloudWatch Embedded Metric Format (EMF) allows for emitting metrics through logs. Using the `aws_embedded_metrics` library, the CloudWatch agent is leveraged to emit EMF logs. This PR adds this metrics option to deltacat.

Testing
 - Tested locally by varying/omitting each field

### DIFF
--- a/deltacat/utils/metrics.py
+++ b/deltacat/utils/metrics.py
@@ -2,20 +2,27 @@ from dataclasses import dataclass
 
 import logging
 
+from aws_embedded_metrics import metric_scope
+from aws_embedded_metrics.config import get_config
+from aws_embedded_metrics.logger.metrics_logger import MetricsLogger
 from deltacat import logs
 from enum import Enum
 from typing import Dict, Any, List, Callable
 from deltacat.aws.clients import resource_cache
 from datetime import datetime
 
+from ray._private.services import get_node_ip_address
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 DEFAULT_DELTACAT_METRICS_NAMESPACE = "ray-deltacat-metrics"
+DEFAULT_DELTACAT_LOG_GROUP_NAME = "ray-deltacat-metrics-EMF-logs"
+DEFAULT_DELTACAT_LOG_STREAM_CALLABLE = get_node_ip_address
 
 
 class MetricsTarget(str, Enum):
     CLOUDWATCH = "cloudwatch"
+    CLOUDWATCH_EMF = "cloudwatch_emf"
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


deltacat/utils/metrics.py
```diff
@@ -117,8 +128,51 @@ def _emit_cloudwatch_metrics(
...
+    )
+    metrics_name_with_type = _build_metrics_name(metrics_type, metrics_name)
+    try:
+        metrics.set_timestamp(ct)
+        metrics.set_dimensions(dimensions)
+        metrics.set_namespace(metrics_config.metrics_namespace)
+
+        metrics_kwargs = metrics_config.metrics_kwargs
+
+        Config = get_config()
```
raghumdani, line 151: Config is a class?

deltacat/utils/metrics.py
```diff
@@ -117,8 +128,51 @@ def _emit_cloudwatch_metrics(
...
 
+@metric_scope
+def _emit_cloudwatch_emf_metrics(
+    metrics: MetricsLogger,
+    metrics_name: str,
+    metrics_type: Enum,
+    metrics_config: MetricsConfig,
+    value: str,
+) -> None:
+    ct = datetime.now()
```
raghumdani, line 139: Avoid abbreviations in naming variables.

deltacat/utils/metrics.py
```diff
@@ -117,8 +128,51 @@ def _emit_cloudwatch_metrics(
...
+        Config = get_config()
+        Config.log_group_name = (
+            metrics_kwargs["log_group_name"]
+            if "log_group_name" in metrics_kwargs
+            else DEFAULT_DELTACAT_LOG_GROUP_NAME
+        )
+
+        # use a callable to differentiate many possible log streams
+        Config.log_stream_name = (
+            metrics_kwargs["log_stream_name"]()
```
raghumdani, line 160: Document what `metrics_kwargs` can contain. Currently, without looking at this implementation, we have no way of knowing what kwargs are supported and their types.

deltacat/utils/metrics.py
```diff
@@ -128,5 +182,4 @@ def emit_timer_metrics(metrics_name, value, metrics_config, **kwargs):
         metrics_type=MetricsType.TIMER,
         metrics_config=metrics_config,
         value=value,
-        **kwargs,
```
raghumdani, line 131: Why removed this? This can still be useful to pass in additional kwargs to the emit_timer_metrics itself.

raghumdani (commented):
Looks good mostly. Please address minor comments.

deltacat/utils/metrics.py
yankevn: Are you saying it should be renamed to something more generic?

deltacat/utils/metrics.py
yankevn: My thought back then was there would be many cases where kwargs would not be used by the emitter function. I guess that's not really that important considering the upside of more fine-grained metrics control within deltacat.

raghumdani (approved):
LGTM!

PR #225, opened 2023-09-25
Handle case for valid deltas with empty manifest entries

rkenmi:
The compactor today assumes each new delta has more than 0 input files. This patches an auditing bug that incorrectly assumes that manifests exist.


deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -212,6 +212,10 @@ def _execute_compaction(
         logger.info("No input deltas found to compact.")
         return None, None, None
 
+    if not uniform_deltas:
+        logger.info("No uniform deltas found to compact.")
+        return None, None, None
+
```
raghumdani, line 218: This is a valid scenario, we have to reference all files and write back the compacted shard with updated stream position. That would correctly update the completeness. if we return like this, the completeness of the table is never updated.

deltacat/tests/compute/compact_partition_test_cases.py
```diff
@@ -1087,9 +1092,31 @@ def with_compactor_version_func_test_param(
     ),
 }
 
-INCREMENTAL_TEST_CASES = with_compactor_version_func_test_param(INCREMENTAL_TEST_CASES)
 
+NO_RCF_OUTPUT_TEST_CASES: Dict[str, NoRCFOutputCompactionTestCaseParams] = {
```
raghumdani, line 1096: We have to ideally add a test case for rebase then incremental where incremental contains delta without any files in it.

deltacat/tests/local_deltacat_storage/__init__.py
```diff
@@ -86,6 +86,31 @@ def _get_manifest_entry_uri(manifest_entry_id: str) -> str:
     return f"cloudpickle://{manifest_entry_id}"
 
 
+def _create_empty_delta(
```
raghumdani, line 89: This method doesn't belong here. It can be moved to test_utils.

raghumdani, line 89: Moved this to the bottom of the file.

rkenmi: Moved to a separate file.

raghumdani (approved):
LGTM!

deltacat/tests/test_utils/storage.py
```diff
@@ -0,0 +1,34 @@
+from typing import Optional, Dict
+
+from deltacat.aws.redshift import Manifest, ManifestMeta
+from deltacat.storage import Partition, DeltaType, DeltaLocator, Delta
+from deltacat.utils.common import current_time_ms
+
+
+def create_empty_delta(
```
raghumdani, line 8: Minor, this can actually belong to [local_deltacat_storage/__init__.py](https://github.com/ray-project/deltacat/pull/225/files#diff-f03e453d0c87446afab06f5045c36f0a16c23fafb51004f652f266e63059f1f9). I had deleted that earlier comment.

rkenmi: Yeah, I noticed that as well, but I figured separating out test helper methods from the local storage impl. would be easier to read and maintain in the future.

PR #226, opened 2023-09-26
Simplify the hash initialization in the memcached object store

raghumdani:
#150 

This commit simplifies the hasher initialization. This also introduces a new method using which we can close all active connections to the object store. This method is particularly helpful whether we are done with the `object_store` object.

### DIFF
--- a/deltacat/io/memcached_object_store.py
+++ b/deltacat/io/memcached_object_store.py
@@ -36,18 +36,14 @@ class MemcachedObjectStore(IObjectStore):
         self.SEPARATOR = "_"
         self.port = port
         self.storage_node_ips = storage_node_ips
-        self.hasher = None
+        self.hasher = RendezvousHash(nodes=self.storage_node_ips, seed=0)
         self.connect_timeout = connect_timeout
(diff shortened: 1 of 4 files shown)


deltacat/io/memcached_object_store.py
```diff
@@ -73,10 +69,15 @@ def put(self, obj: object, *args, **kwargs) -> Any:
...
 
-        if client.set(uid.__str__(), serialized, noreply=self.noreply):
-            return ref
-        else:
-            raise RuntimeError("Unable to write to cache")
+        try:
+            if client.set(uid.__str__(), serialized, noreply=self.noreply):
+                return ref
+            else:
+                raise RuntimeError("Unable to write to cache")
```
rkenmi, line 76: Minor: Can we log the ref here as well to record the node IP for the failed cache set?

rkenmi (approved):
LGTM

deltacat/io/memcached_object_store.py
raghumdani: yeah sure.

PR #228, opened 2023-09-27
Fix EC2 instance metadata queries not getting caught by the retry handler

rkenmi:
The first call to `_get_url` may fail and raise an exception that is not caught by tenacity's retries.


raghumdani (approved):
LGTM!

PR #229, opened 2023-09-28
Include ConnectionError for retries during EC2 instance metadata calls

rkenmi:


deltacat/aws/clients.py
```diff
@@ -46,9 +46,9 @@ class RetryIfRetryableHTTPStatusCode(retry_if_exception):
 
     def __init__(self):
         def is_retryable_error(exception):
-            return (
+            return isinstance(exception, requests.exceptions.ConnectionError) or (
                 isinstance(exception, requests.exceptions.HTTPError)
-                and exception.response.status_code in RETRYABLE_HTTP_STATUS_CODES
+                and exception.response.status_code in RETRYABLE_HTTP_STATUS_CODES,
```
raghumdani, line 51: why an additional `,` at the end? wouldn't that make it a tuple?

rkenmi: Unintentional change, fixed.

deltacat/aws/clients.py
```diff
@@ -46,7 +46,7 @@ class RetryIfRetryableHTTPStatusCode(retry_if_exception):
 
     def __init__(self):
         def is_retryable_error(exception):
-            return (
+            return isinstance(exception, requests.exceptions.ConnectionError) or (
```
raghumdani, line 49: Are you sure it captures all errors we saw in production?

raghumdani (approved):
LGTM! Please double check all errors are captured.

deltacat/aws/clients.py
rkenmi: We were encountering`HttpError` in production, so it is a safe bet for the time being. Adding `ConnectionError` is more of a precautionary step after looking at the request [docs](https://requests.readthedocs.io/en/latest/user/quickstart/?highlight=exceptions#errors-and-exceptions), but going forward, we can incrementally add new exceptions here as needed.

PR #230, opened 2023-09-30
Add retries and retry_exceptions to overcome throttling

raghumdani:
#150 

This commit will add the `max_retries` and `retry_exceptions` to handle throttling errors.

### DIFF
--- a/deltacat/tests/compute/compactor_v2/utils/test_task_options.py
+++ b/deltacat/tests/compute/compactor_v2/utils/test_task_options.py
@@ -1,0 +1,33 @@
+import unittest
+import ray
+from deltacat.compute.compactor_v2.utils.task_options import get_task_options
+
+
+@ray.remote
+def valid_func():
+    return 2
+
+
+@ray.remote
+def throwing_func():
+    raise ConnectionAbortedError()
+
+
+class TestTaskOptions(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        ray.init(local_mode=True, ignore_reinit_error=True)
+        super().setUpClass()
+
+    def test_get_task_options_sanity(self):
+        opts = get_task_options(0.01, 0.01)
+        result_ref = valid_func.options(**opts).remote()
+        result = ray.get(result_ref)
+
+        self.assertEqual(result, 2)
+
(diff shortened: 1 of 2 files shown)


deltacat/tests/compute/compactor_v2/utils/test_task_options.py
```diff
@@ -0,0 +1,33 @@
...
+        result_ref = valid_func.options(**opts).remote()
+        result = ray.get(result_ref)
+
+        self.assertEqual(result, 2)
+
+    def test_get_task_options_when_exception_is_thrown(self):
+        opts = get_task_options(0.01, 0.01)
+        result_ref = throwing_func.options(**opts).remote()
+
+        self.assertRaises(ConnectionAbortedError, lambda: ray.get(result_ref))
```
yankevn, line 33: How exactly does this test the task options?

raghumdani: It tests that task options are not invalid/ray is not throwing any validation error. Also, you are right. It won't assert `retry_exceptions` actually retried or not.

yankevn: Would it be possible to check if `throwing_func` gets retried the expected amount of times or not?

raghumdani: Yes, it is possible with this code, but this will require ray dashboard to run in the unit test environment. As a result, our tests can become flaky and non-maintainable. Hence, I decided to not use it. 

```
from ray.util.state import summarize_tasks

...
        summary = summarize_tasks()
        # Assert that this task has failed 3 times due to 3 retries
        self.assertEqual(summary["cluster"]["summary"]["throwing_func"]["state_counts"]["FAILED"], 3)
```

raghumdani: More specifically, it requires Ray to be initialized with `ray.init(local_mode=False, ignore_reinit_error=True)`.

yankevn (approved):
LGTM

pfaraone:
LGTM - thanks for this.

PR #232, opened 2023-10-09
Throw an error with null primary keys

pfaraone:
### Motivation
- Null primary keys can't currently be handled by compactor V1 or V2. The compactor is throwing an `AttributeError` within `_append_sha1_hash_to_table` as the iterated value (`hash_value`) is `NoneType`. Before deciding on a strategy for handling null primary keys, the least surprising way to handle it is to throw an error.


raghumdani (approved):
LGTM!

PR #233, opened 2023-10-10
Fix minor bugs preventing deltacat metrics emission

yankevn:
The bugfixes in this PR:
 - `CompactPartitionParams` used in V2 was not extracting `metrics_config`, resulting in metrics never being enabled.
 - The callable for the log stream rather than its return value was passed into `aws_embedded_metrics`, resulting in a JSON serialization exception being thrown
 - The `aws-embedded-metrics` version is bumped to allow for setting the timestamp

Tested via `pytest` and local executions.


pfaraone (approved):
LGTM

PR #234, opened 2023-10-11
Remove a ray private method dependency

yankevn:
Ray has a public interface for obtaining the node IP, so it will be safer to use the interface rather than directly using the private method. Additionally, delete the erroneously added `db_test.sqlite` file.

Tested locally with Ray 2.3.0.


pfaraone (approved):
LGTM

rkenmi (approved):

PR #235, opened 2023-10-11
Audit logging Ray version

pfaraone:
### Changes/Motivation
- The `ray` distribution version is not currently tracked by the audit log. Given we are using a ['compatible release'](https://peps.python.org/pep-0440/#compatible-release) version specifier for the `ray` version it would be helpful to know the full version in the audit log - whether it is 2.3.1 or or 2.6.3


yankevn (approved):
LGTM

deltacat/compute/compactor/compaction_session.py
```diff
@@ -242,7 +242,9 @@ def _execute_compaction_round(
     logger.info(f"Compaction audit will be written to {audit_url}")
 
     compaction_audit = CompactionSessionAuditInfo(
-        deltacat.__version__, audit_url
+        deltacat.__version__,
+        ray.__version__,
```
raghumdani, line 246: Minor: add all new arguments at the end for backward compatibility.

raghumdani (approved):
LGTM

PR #239, opened 2023-10-13
Change "Publish python distribution to PyPI" workflow to only trigger on a GH release and not on drafting a release

pfaraone:


raghumdani (approved):
LGTM

PR #241, opened 2023-10-24
Restructure deltacat metrics to allow for decorators

yankevn:
This change restructures deltacat's metrics code to allow for metrics emission via decorator functions. To accomplish this, the `MetricsConfigSingleton` class is introduced to make the `metrics_config` object globally available while enforcing immutability as much as possible.

Testing:
 - pytest
 - Tested on personal stack and verified metrics were created


deltacat/compute/compactor_v2/model/merge_result.py
```diff
@@ -8,5 +8,4 @@ class MergeResult(NamedTuple):
     materialize_results: List[MaterializeResult]
     deduped_record_count: np.int64
     peak_memory_usage_bytes: np.double
-    telemetry_time_in_seconds: np.double
```
raghumdani, line 11: Why remove this? Won't we be able measure telemetry time from now?

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -438,36 +445,23 @@ def _materialize(
...
-    return MergeResult(
-        merge_result[0],
-        merge_result[1],
-        merge_result[2],
-        np.double(emit_metrics_time),
-        merge_result[4],
-    )
+def merge(
+    input: MergeInput,
+    metrics_config: Optional[MetricsConfig] = None,
```
raghumdani, line 461: why can't this be part of input like before?

deltacat/compute/compactor/compaction_session.py
```diff
@@ -685,8 +696,15 @@ def _execute_compaction_round(
         session_peak_memory
     )
 
+    metrics_telemetry_time = 0
+    try:
+        metrics_telemetry_time = MetricsConfigSingleton.instance().total_telemetry_time
```
raghumdani, line 701: This won't give you telemetry time from each tasks, correct?

deltacat/compute/compactor/steps/materialize.py
```diff
@@ -323,8 +312,49 @@ def _materialize(compacted_tables: List[pa.Table]) -> MaterializeResult:
...
+    read_kwargs_provider: Optional[ReadKwargsProvider] = None,
+    s3_table_writer_kwargs: Optional[Dict[str, Any]] = None,
+    object_store: Optional[IObjectStore] = None,
+    deltacat_storage=unimplemented_deltacat_storage,
+    deltacat_storage_kwargs: Optional[Dict[str, Any]] = None,
+    metrics_config: Optional[MetricsConfig] = None,
+):
+    # initialize singleton on new process
+    if metrics_config:
+        MetricsConfigSingleton.instance(metrics_config)
```
raghumdani, line 342: What is the benefit on this singleton wrapper class? If you are not using any methods of metrics_config nor is metrics_config compute intensive to create, what is the real benefit?

deltacat/utils/metrics.py
```diff
@@ -38,8 +42,63 @@ def __init__(
         self.region = region
         self.metrics_target = metrics_target
         self.metrics_namespace = metrics_namespace
-        self.metrics_dimensions = metrics_dimensions
-        self.metrics_kwargs = metrics_kwargs
+
+        # Pyrsistent is unable to enforce immutability on nested objects
+        # Please avoid modifying stored objects to preserve read-only status
+        self.metrics_dimensions = pyrsistent.v(*metrics_dimensions)
```
raghumdani, line 48: This solution is not complete for enforcing immutability.

raghumdani (commented):
This is not a full review as I have only taken a glance. I have left comments for the main things I saw in the implementation that we may need to discuss.

deltacat/compute/compactor_v2/model/merge_result.py
yankevn: With the move to decorators, you can no longer measure telemetry time (i.e. time taken to emit metrics) from within the merge function.

deltacat/compute/compactor_v2/steps/merge.py
yankevn: `metrics_config` was never part of the input. We could make it part of the `MergeInput` class, but I feel that doesn't make as much sense, as the `metrics_config` object is not actually used in merge.

deltacat/utils/metrics.py
yankevn: Do you mean for just `metrics_dimensions` or also for `metrics_kwargs`? I suppose for `metrics_dimensions` you could enforce perfect immutability, but `metrics_kwargs` is unenforceable due to its open-endedness.

deltacat/compute/compactor/steps/materialize.py
yankevn: The benefit is to create a global `metrics_config` object that cannot be accidentally overwritten.

deltacat/compute/compactor_v2/model/merge_result.py
yankevn: That's a good point, the code will need to be fixed for that.

yankevn:
TODO: add performance test + report numbers here

deltacat/utils/metrics.py
yankevn: After discussion, we decided to remove pyrsistent because it doesn't actually solve the problem of immutability, it just adds bloat to our total packages.

raghumdani:
Resolved by #294

PR #246, opened 2023-11-10
Upload audit to s3 after setting all args

raghumdani:
Currently, we were setting audit after uploading all the args. For incremental cases, this will only set the size of the incremental delta instead of total size of input. This PR fixes this issue.


rkenmi (approved):

yankevn (approved):
LGTM

PR #248, opened 2023-11-17
Remove cluster_util from deltacat

yankevn:
`ClusterUtilizationOverTimeRange` provides misleading behavior when used within deltacat. This is because it estimates cluster-level resource usage rather than partition-specific usage. As a result, it's better to use `ClusterUtilizationOverTimeRange` from within the driver class that calls deltacat functions.

Testing:
 - Tested on personal stack

### DIFF
--- a/deltacat/compute/compactor_v2/compaction_session.py
+++ b/deltacat/compute/compactor_v2/compaction_session.py
@@ -47,7 +47,6 @@ from deltacat.compute.compactor_v2.utils.task_options import (
     hash_bucket_resource_options_provider,
     merge_resource_options_provider,
 )
-from deltacat.utils.resources import ClusterUtilizationOverTimeRange
 from deltacat.compute.compactor.model.compactor_version import CompactorVersion
 
 if importlib.util.find_spec("memray"):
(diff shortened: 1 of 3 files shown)


deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -478,17 +476,6 @@ def merge_input_provider(index, item):
...
-    cluster_util: ClusterUtilizationOverTimeRange = kwargs.get("cluster_util")
-
-    if cluster_util:
-        compaction_audit.set_total_cpu_seconds(cluster_util.total_vcpu_seconds)
-        compaction_audit.set_used_cpu_seconds(cluster_util.used_vcpu_seconds)
-        compaction_audit.set_used_memory_gb_seconds(cluster_util.used_memory_gb_seconds)
-        compaction_audit.set_total_memory_gb_seconds(
-            cluster_util.total_memory_gb_seconds
-        )
-        compaction_audit.set_cluster_cpu_max(cluster_util.max_cpu)
```
raghumdani, line 490: Remove the keys from `compaction_session_audit_info`. Or document on them saying they are not current emitted until we figure out how to measure vCPUSeconds at partition level.

raghumdani (commented):
LGTM! Please address minor comment.

raghumdani (approved):
LGTM

raghumdani (approved):
LGTM

PR #249, opened 2023-11-21
add sliced string cast for hash column generator

samster25:


deltacat/compute/compactor_v2/utils/primary_key_index.py
```diff
@@ -167,6 +167,40 @@ def group_by_pk_hash_bucket(
...
+    MAX_BYTES = 2147483646
+    max_str_len = None
+    if pa.types.is_integer(dtype):
+        max_str_len = 21  # -INT_MAX
+    elif pa.types.is_floating(dtype):
+        max_str_len = 24
+    elif pa.types.is_decimal128(dtype):
+        max_str_len = 39
+    elif pa.types.is_decimal256(dtype):
+        max_str_len = 77
```
pdames, line 184: A couple minor asks:
1. I think both signed integers (-2^63) and unsigned integers (2^64) have a max length of 20 bytes. Am I missing something here that led to the conclusion of 21 bytes?
2. Can we add some comments giving a bit more context to each figure here... maybe something like the following:
```suggestion
        max_str_len = 21  # UINT64_MAX (2^64) / -INT64_MAX(-2^63)
    elif pa.types.is_floating(dtype):
        max_str_len = 24  # ??? TODO
    elif pa.types.is_decimal128(dtype):
        max_str_len = 39. # ??? TODO
    elif pa.types.is_decimal256(dtype):
        max_str_len = 77 # ??? TODO
```
Or, to provide this clarity implicitly, what do you think about having constants where we define the pyarrow type that is assumed to have the max string length, cast it to a string, and count the digits once, for example:
```
PA_UINT64_MAX_STR_BYTES = pa.compute.binary_length(pa.compute.cast(pa.scalar(2**64-1, type=pa.uint64()), pa.string())).as_py()
PA_INT64_MAX_STR_BYTES = pa.compute.binary_length(pa.compute.cast(pa.scalar(-2**63, type=pa.int64()), pa.string())).as_py()
PA_INT_MAX_STR_BYTES = max(PA_UINT64_MAX_STR_BYTES, PA_INT64_MAX_STR_BYTES)
```

deltacat/compute/compactor_v2/utils/primary_key_index.py
```diff
@@ -167,6 +167,40 @@ def group_by_pk_hash_bucket(
     return result
 
 
+def _sliced_string_cast(array: pa.ChunkedArray) -> pa.ChunkedArray:
```
raghumdani, line 170: would moving it to pyarrow utils and take in second arg as type make it more generic?

raghumdani (commented):
Thanks for the quick fix.

deltacat/compute/compactor_v2/utils/primary_key_index.py
samster25: I mistakenly did `(-2^64)` which led to that. I moved to cached functions that compute the various lengths. (Moved to `deltacat/utils/pyarrow.py`)

deltacat/compute/compactor_v2/utils/primary_key_index.py
samster25: moved over to the utils. What would the second arg express?

deltacat/compute/compactor_v2/utils/primary_key_index.py
```diff
@@ -167,6 +167,40 @@ def group_by_pk_hash_bucket(
...
+    """performs slicing of a pyarrow array prior casting to a string.
+    This prevents a pyarrow from allocating too large of an array causing a failure.
```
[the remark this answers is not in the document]
samster25: done!

deltacat/compute/compactor_v2/utils/primary_key_index.py
raghumdani: The type to cast it to i.e., `pa.string()`?

samster25: this function is only relevant for casting to `string` types. So would the second arg be either `pa.string()` and `pa.large_string()`?

Or are you thinking to make this a generic casting function for all types, special casing when we see `pa.string()`?

raghumdani: Yeah, I was thinking this sliced cast is relevant for all target types, but not just strings. We could only support strings for now.

raghumdani (approved):
LGTM!

deltacat/utils/pyarrow.py
```diff
@@ -738,3 +741,69 @@ def clear_remaining(self) -> None:
...
+
+
+def _max_decimal256_string_len():
+    return 78  # "-" + 76 digits + decimal
+
+
+def sliced_string_cast(array: pa.ChunkedArray) -> pa.ChunkedArray:
+    """performs slicing of a pyarrow array prior casting to a string.
+    This prevents a pyarrow from allocating too large of an array causing a failure.
+    Issue: https://github.com/apache/arrow/issues/38835
```
raghumdani, line 779: The issue has an answer. Looks like casting to `large_string()` works fine.

rkenmi: Do we really want to cast to this type? Seems like it adds a decent amount of memory overhead.

```python
>>> parr = pa.array(list(range(100000000)))
par_s = pc.cast(parr, pa.string())
par_ls = pc.cast(parr, pa.large_string())
print(par_s.nbytes, par_ls.nbytes)
1188888890 1588888890
```

samster25: So are we favoring the slicing route in this PR or take the extra memory overhead route with `large_string`

rkenmi (approved):

deltacat/utils/pyarrow.py
rkenmi: I think this looks good to me 🙂 we can add an enhancement issue for this in the meantime

PR #258, opened 2024-01-31
Add debug logs to capture params used for memory calculation

yankevn:
This change adds debug logs for params used when calculating the allocated memory for merge tasks. These logs will help debugging efforts for nodes whose process memory consistently surpasses the allocated memory, thus implying the allocated memory calculation was incorrect.

### DIFF
--- a/deltacat/compute/compactor_v2/utils/task_options.py
+++ b/deltacat/compute/compactor_v2/utils/task_options.py
@@ -1,5 +1,7 @@
 import botocore
+import logging
 from typing import Dict, Optional, List, Tuple
+from deltacat import logs
 from deltacat.types.media import ContentEncoding, ContentType
 from deltacat.types.partial_download import PartialParquetParameters
 from deltacat.storage import (
@@ -16,6 +18,8 @@ from deltacat.compute.compactor_v2.constants import (
     TOTAL_MEMORY_BUFFER_PERCENTAGE,
     PARQUET_TO_PYARROW_INFLATION,
 )
+
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 
 def _get_parquet_type_params_if_exist(
@@ -133,11 +137,15 @@ def hash_bucket_resource_options_provider(
     ray_custom_resources: Optional[Dict] = None,
     **kwargs,
 ) -> Dict:
+    debug_memory_params = {"hash_bucket_task_index": index}
     size_bytes = 0.0
     num_rows = 0
     total_pk_size = 0
 
     if not item.manifest or not item.manifest.entries:
+        logger.debug(
+            f"[Hash bucket task {index}]: No manifest entries, skipping memory allocation calculation"
+        )
         return {"CPU": 0.01}
 
     for entry in item.manifest.entries:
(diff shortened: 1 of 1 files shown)


raghumdani (commented):
Two points:
1. Do you also want to log hash bucket memory calculation params?
2. In merge logs, we may need to add merge index so that we can easily connect.

raghumdani (commented):
LGTM! Few minor comments.

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -165,9 +171,17 @@ def hash_bucket_resource_options_provider(
...
+    debug_memory_params["size_bytes"] = size_bytes
+    debug_memory_params["num_rows"] = num_rows
+    debug_memory_params["total_pk_size"] = total_pk_size
+    debug_memory_params["total_memory"] = total_memory
 
     # Consider buffer
     total_memory = total_memory * (1 + TOTAL_MEMORY_BUFFER_PERCENTAGE / 100.0)
+    debug_memory_params["total_memory_with_buffer"] = total_memory
+    logger.debug(
+        f"Params used for calculating hash bucketing memory: {debug_memory_params}"
```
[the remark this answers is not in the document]
yankevn: Sure, makes sense

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -133,11 +137,15 @@ def hash_bucket_resource_options_provider(
     ray_custom_resources: Optional[Dict] = None,
     **kwargs,
 ) -> Dict:
+    debug_memory_params = {"hash_bucket_task_index": index}
```
raghumdani, line 140: This is now redundant!

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -186,10 +202,13 @@ def merge_resource_options_provider(
     deltacat_storage_kwargs: Optional[Dict] = {},
     **kwargs,
 ) -> Dict:
+    debug_memory_params = {"merge_task_index": index}
```
raghumdani, line 205: Same here, as this is now redundant.

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -256,7 +278,14 @@ def merge_resource_options_provider(
...
     )
+    debug_memory_params["data_size"] = data_size
+    debug_memory_params["num_rows"] = num_rows
+    debug_memory_params["pk_size_bytes"] = pk_size_bytes
+    debug_memory_params["incremental_index_array_size"] = incremental_index_array_size
+    debug_memory_params["total_memory"] = total_memory
 
     total_memory = total_memory * (1 + TOTAL_MEMORY_BUFFER_PERCENTAGE / 100.0)
+    debug_memory_params["total_memory_with_buffer"] = total_memory
+    logger.debug(f"Params used for calculating merge memory: {debug_memory_params}")
```
raghumdani, line 289: Merge index not added here.

yankevn: Whoops, forgot to hit save 🙃

deltacat/compute/compactor_v2/utils/task_options.py
yankevn: I think it's ok to keep these, it makes the printed objects self-identifiable.

raghumdani: I don't see them being read outside of the log. So, not sure why they have to self-identifiable. I have no strong bias either ways.

raghumdani (approved):
LGTM!

PR #263, opened 2024-02-20
Add task index into logs for hb and merge

yankevn:
Add the task index into the hb and merge logs. This makes which tasks have started/ended clearer while also logging the hb task index, which was not logged before.

### DIFF
--- a/deltacat/compute/compactor_v2/model/hash_bucket_input.py
+++ b/deltacat/compute/compactor_v2/model/hash_bucket_input.py
@@ -15,6 +15,7 @@ class HashBucketInput(Dict):
         primary_keys: List[str],
         num_hash_buckets: int,
         num_hash_groups: int,
+        hb_task_index: Optional[int] = 0,
         enable_profiler: Optional[bool] = False,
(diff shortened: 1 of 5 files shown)


deltacat/compute/compactor_v2/model/hash_bucket_input.py
```diff
@@ -15,6 +15,7 @@ def of(
         primary_keys: List[str],
         num_hash_buckets: int,
         num_hash_groups: int,
+        hb_task_index: Optional[int] = 0,
```
raghumdani, line 18: Is this passed only for logging? Isn't taskID sufficient for identifying logs?

yankevn: No, the problem is the memory allocation logs are emitted by the driver, whereas the actual hash bucketing logic is done by the worker. The workers currently do not emit which hash bucket index they are, so it is impossible (or at least much more difficult than it should be, if a way exists) to tell which worker corresponds to which memory allocation log.

raghumdani (approved):
LGTM

Issue #265, opened 2024-02-22
[V2][API] Daft reader for DeltaCAT catalogs in general


raghumdani:
This feature allows reading the DeltaCAT catalog (Iceberg, internal catalog etc.) into a Daft distributed dataframe. It also provides hints regarding data layout to Daft reader so that the reads are optimized and no worse than existing compactor.  

### Expected Features

- [ ] Reading entire table into memory. 
- [ ] Leverage Partition Pruning using DeltaCAT metadata. 
- [ ] Leverage sort order policy using DeltaCAT metadata. 
- [ ] Support merge on read and hence, delete deltas.
- [ ] Unblocks #170.

PR #266, opened 2024-02-22
Initial draft of Daft dataframe support for DeltaCAT catalog

raghumdani:
## Basic Example

```
from deltacat.catalog import default_catalog_impl as dc

dc.initialize(ds=ds)
df = dc.read_table(
    table=TABLE_NAME,
    namespace=self.TABLE_NAMESPACE,
    catalog=self.LOCAL_CATALOG_NAME,
    distributed_dataset_type=DistributedDatasetType.DAFT,
    merge_on_read=False,
    deltacat_storage_kwargs=self.kwargs,
)

df.show()
```

Part of #265

### DIFF
--- a/deltacat/aws/s3u.py
+++ b/deltacat/aws/s3u.py
@@ -357,17 +357,7 @@ def download_manifest_entry(
     content_encoding: Optional[ContentEncoding] = None,
 ) -> LocalTable:
 
-    conf = Config(retries={"max_attempts": BOTO_MAX_RETRIES, "mode": "adaptive"})
-    s3_client_kwargs = (
-        {
-            "aws_access_key_id": token_holder["accessKeyId"],
-            "aws_secret_access_key": token_holder["secretAccessKey"],
-            "aws_session_token": token_holder["sessionToken"],
-            "config": conf,
-        }
-        if token_holder
-        else {"config": conf}
-    )
+    s3_client_kwargs = _get_s3_client_kwargs_from_token(token_holder=token_holder)
     if not content_type:
         content_type = manifest_entry.meta.content_type
         assert (
@@ -477,7 +467,7 @@ def download_manifest_entries_distributed(
     if distributed_dataset_type == DistributedDatasetType.RAY_DATASET:
         return _download_manifest_entries_ray_data_distributed(**params)
     elif distributed_dataset_type is not None:
-        return _download_manifest_entries_any_distributed(**params)
+        return _download_manifest_entries_all_dataset_distributed(**params)
     else:
         raise ValueError(
             f"Distributed dataset type {distributed_dataset_type} not supported."
@@ -579,6 +569,20 @@ def _block_metadata(block: Block) -> BlockMetadata:
     )
 
 
+def _get_s3_client_kwargs_from_token(token_holder) -> Dict[Any, Any]:
+    conf = Config(retries={"max_attempts": BOTO_MAX_RETRIES, "mode": "adaptive"})
+    return (
+        {
+            "aws_access_key_id": token_holder["accessKeyId"],
+            "aws_secret_access_key": token_holder["secretAccessKey"],
+            "aws_session_token": token_holder["sessionToken"],
+            "config": conf,
+        }
+        if token_holder
+        else {"config": conf}
+    )
+
+
 def _get_metadata(
     table: Union[LocalTable, DistributedDataset],
     write_paths: List[str],
@@ -652,7 +656,7 @@ def _download_manifest_entries_ray_data_distributed(
     return TABLE_TYPE_TO_DATASET_CREATE_FUNC_REFS[table_type](table_pending_ids)
 
 
-def _download_manifest_entries_any_distributed(
+def _download_manifest_entries_all_dataset_distributed(
     manifest: Manifest,
     token_holder: Optional[Dict[str, Any]] = None,
     table_type: TableType = TableType.PYARROW,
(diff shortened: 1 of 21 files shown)


deltacat/storage/interface.py
```diff
@@ -178,9 +183,10 @@ def download_delta(
     columns: Optional[List[str]] = None,
     file_reader_kwargs_provider: Optional[ReadKwargsProvider] = None,
     ray_options_provider: Callable[[int, Any], Dict[str, Any]] = None,
+    distributed_dataset_type: DistributedDatasetType = DistributedDatasetType.RAY_DATASET,
```
rkenmi, line 186: I feel this overcomplicates the interface slightly. Rather than having a new parameter, could we not re-use `table_type` and `storage_type`? For example, if we want our deltas read into a Daft DataFrame, could we just specify something like `TableType.DAFT_DATAFRAME` and `StorageType.DISTRIBUTED` to achieve the same result?

deltacat/aws/s3u.py
```diff
@@ -574,3 +522,200 @@ def download(
...
+            raise ValueError(
+                f"Mixed content encoding of {entry_content_encoding},"
+                f" {entry.meta.content_encoding} is not supported."
+            )
+
+        entry_content_type = entry.meta.content_type
+        entry_content_encoding = entry.meta.content_encoding
+        uris.append(entry.uri)
+
+    conf = Config(retries={"max_attempts": BOTO_MAX_RETRIES, "mode": "adaptive"})
```
rkenmi, line 695: Is the S3 config and client kwargs duplicated from somewhere?

deltacat/aws/s3u.py
```diff
@@ -574,3 +522,200 @@ def download(
...
+            column_names,
+            include_columns,
+            file_reader_kwargs_provider,
+            max_parallelism=max_parallelism,
+            options_provider=ray_options_provider,
+        )
+    return TABLE_TYPE_TO_DATASET_CREATE_FUNC_REFS[table_type](table_pending_ids)
+
+
+def _download_manifest_entries_any_distributed(
```
Zyiqin-Miranda, line 655: Minor: The naming `any_distributed` seems a bit confusing, could we just remove the `any`?

deltacat/aws/s3u.py
```diff
@@ -574,3 +522,200 @@ def download(
...
+    ray_options_provider: Callable[[int, Any], Dict[str, Any]] = None,
+    distributed_dataset_type: Optional[
+        DistributedDatasetType
+    ] = DistributedDatasetType.RAY_DATASET,
+) -> DistributedDataset:
+
+    entry_content_type = None
+    entry_content_encoding = None
+    uris = []
+    for entry in manifest.entries or []:
```
Zyiqin-Miranda, line 672: What does the empty array `or []` do here?

deltacat/catalog/default_catalog_impl/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+        partition_filter = STORAGE.list_partitions(
+            table_name=table,
+            namespace=namespace,
+            table_version=table_version,
+            **deltacat_storage_kwargs,
+        ).all_items()
+
+    qualified_deltas = _get_deltas_from_partition_filter(
+        stream_position_range_inclusive=stream_position_range_inclusive,
```
Zyiqin-Miranda, line 117: Should this `stream_position_range_inclusive` become a map of `partition_filter:{stream_position_range}` instead? Given that we can download deltas from different partitions?

deltacat/aws/s3u.py
raghumdani: manifest.entries can be None. Hence, this avoid failing this loop.

deltacat/catalog/default_catalog_impl/__init__.py
raghumdani: I thought about it. Users, most likely won't have use-case to download from different stream position range from different partitions. We can add support for that as well if required. Right, keeping the API simple.

deltacat/catalog/default_catalog_impl/__init__.py
```diff
@@ -0,0 +1,369 @@
...
+
+def default_namespace() -> str:
+    """Returns the default namespace for the catalog."""
+    raise NotImplementedError("default_namespace not implemented")
+
+
+# catalog functions
+def initialize(ds: deltacat_storage, *args, **kwargs) -> None:
+    """Initializes the data catalog with the given arguments."""
+    global STORAGE
```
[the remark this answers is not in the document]
raghumdani: Another way is to pass in storage as kwargs.

deltacat/storage/interface.py
raghumdani: The reason why we needed a enums was, you can have RAY_DATASET with table type being pandas, arrow, or even numpy.

deltacat/aws/s3u.py
raghumdani: Yes, I created a separate function reusing them.

deltacat/aws/s3u.py
raghumdani: There already exists a method named `download_manifest_entries_distributed`. I changed it to `_download_manifest_entries_all_dataset_distributed`.

rkenmi (approved):

Zyiqin-Miranda (approved):
LGTM!

PR #269, opened 2024-02-22
Increase memory allocation calculation based on missed memory inefficiencies

yankevn:
The existing memory allocation logic does not account for several unexpected inefficiencies. 

These inefficiencies are:

1. When converting the pyarrow table to numpy, the process memory usage is doubling, indicating that pyarrow is not performing a zero-copy conversion like we expected. To account for this, we add the pk_size_bytes attribute again.
2. When generating sha1 hashes using hashlib, the memory inefficiencies of the library accumulate for tables with large numbers of primary keys. To account for this, we add num_rows * 20 again.

Testing:
 - These changes were tested by running on personal stack and verifying the memory allocation limit was increased.

### DIFF
--- a/deltacat/compute/compactor_v2/utils/task_options.py
+++ b/deltacat/compute/compactor_v2/utils/task_options.py
@@ -170,9 +170,16 @@ def hash_bucket_resource_options_provider(
             else:
                 total_pk_size += pk_size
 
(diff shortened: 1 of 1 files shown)


raghumdani (approved):
LGTM!

PR #270, opened 2024-02-27
bump deltacat major version 0.3.0

raghumdani:
This version contains a potentially backward incompatible changes for sungate python.


rkenmi (commented):
LGTM, should we update requirements.txt as well?

PR #272, opened 2024-03-05
Downgrade Daft and patch version 0.2.11

raghumdani:


Zyiqin-Miranda (approved):
LGTM!

samster25:
@raghumdani This PR https://github.com/ray-project/deltacat/pull/274 should prevent API calls hitting the local IMDS (with version `getdaft>=0.2.15` which we think caused the DNS errors you were seeing. I believe deltacat/main is currently at `getdaft==0.2.16` which should be compatible with this change. I also made a PR that upgrades daft the latest [here](https://github.com/ray-project/deltacat/pull/275) but you shouldn't need it to test the fix.

pfaraone:
Is this PR still relevant?

raghumdani:
No, cancelling this.

PR #273, opened 2024-03-07
Handling equality-based columnar deletes

pfaraone:
## Overview
This PR makes backward-incompatible changes to how delete-type deltas are handled. Currently, the deltaCAT specification requires that DELETE-type deltas must contain the primary key of the records that they are affecting. This PR introduces a different way of handling row-level deletes. Delete-type deltas deletes all records that match the specified delete column values. In other words they mark a row deleted by one or more column values - like `id = 5`

For a given DELETE-type delta, 
- The delete columns are the columns of the delete file used to match data rows and are retrieved from the input deltas. The string list of canonical delete column names is retrieved from the `Delta.properties`
- The underlying DELETE-type delta storage files data identify deleted rows by one or more column values

Example: 
```
Example: 

(Upsert, 1)
(Delete, 2)
(Upsert, 3)
(Upsert, 4)
(Delete, 5)
(Delete, 6)

U1 will apply D2
U3 will apply D5/D6
U4 will apply D5/D6 (edited)
```


deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,92 @@
...
+            deletes_at_this_stream_position
+        )
+        stream_position_of_earliest_delete_in_sequence: int = uniform_deltas[
+            window_start
+        ].stream_position
+        deletes_obj_ref_by_stream_position[
+            stream_position_of_earliest_delete_in_sequence
+        ] = ray.put(consolidated_deletes)
+        window_start = window_end
+        window_end = window_start
```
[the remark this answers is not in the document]
pfaraone: Good catch - it's not necessary.

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -348,15 +429,19 @@ def _timed_merge(input: MergeInput) -> MergeResult:
...
+                not merge_file_group.dfe_groups
+                and not input.deletes_to_apply_by_stream_positions
+            ):
                 hb_index_copy_by_ref_ids.append(merge_file_group.hb_index)
                 continue
-
             table, input_records, deduped_records = _compact_tables(
                 input, merge_file_group.dfe_groups, merge_file_group.hb_index
             )
+            if table is None:
```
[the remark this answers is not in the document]
pfaraone: Currently,  we copy the data by reference in the above if statement, preventing this scenario from occurring.

However, the current implementation allows deletes to apply only to the compacted table without affecting the incremental table (no dfe_groups). When deletes are present, we need to download the table for every hb_index if a compacted table corresponding to the hb_index exists. If the deletes apply to the table, we need to drop rows

The only case where the table is None is when there are no dfe_groups corresponding to the merge_file_group and no compacted table exists for the manifest_entry mapped by the hb_index.

Zyiqin-Miranda:
A few general question after a brief look at this PR and to better understand approach here:
1. On prepare_deletes, is there any particular reason why it's done on annotated_deltas which could be around 20k+ for large tables instead of original delta (which is just 1 in rebase case)?
2. Are we separating out the Delete operation in driver in this implementation? 
For example, For deltas coming in the order of Upsert, Delete, Upsert. We will only compact up to where Delete delta stream position? Cause I see in merge step, we just sort by stream_position and delete accordingly.

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,92 @@
...
+import pyarrow as pa
+from deltacat.compute.compactor import (
+    DeltaAnnotated,
+)
+
+
+def prepare_deletes(
+    params: CompactPartitionParams,
+    uniform_deltas: List[DeltaAnnotated],
+    deletes_obj_ref_by_stream_position: IntegerRangeDict,
```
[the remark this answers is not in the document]
pfaraone: Changed the signature - instantiating `IntegerRangeDict` in prepare_deletes

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,92 @@
...
+            window_start += 1
+            window_end = window_start
+            continue
+        # extend out the delete window to include all consecutive deletes
+        while (
+            window_end < len(uniform_deltas)
+            and uniform_deltas[window_end].annotations[0].annotation_delta_type
+            is DeltaType.DELETE
+        ):
+            window_end += 1
```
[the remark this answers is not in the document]
pfaraone: A window represents consecutive DELETE-type deltas sequence within the `uniform_deltas`

For example:
```
DeltaType | Stream Position
U | 1
D | 2
U | 3
U | 4
D | 5
D | 6
D | 7
U | 8
```

In this case the windows will  [(D | 2), (D | 5, D | 6, D | 7)]

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -279,21 +373,15 @@ def _compact_tables(
...
     compacted_table = None
-
-    if (
-        input.round_completion_info
-        and input.round_completion_info.hb_index_to_entry_range
-        and input.round_completion_info.hb_index_to_entry_range.get(str(hb_idx))
-        is not None
-    ):
+    if _does_hash_bucket_idx_have_compacted_table(input, hb_idx):
         compacted_table = _download_compacted_table(
```
Zyiqin-Miranda, line 377: This `_does_hash_bucket_idx_have_compacted_table` is called twice if it returns True, then we download the table twice?

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -48,6 +50,22 @@
...
+def _does_hash_bucket_idx_have_compacted_table(input: MergeInput, hb_idx: int) -> bool:
+    return (
+        input.round_completion_info
+        and input.round_completion_info.hb_index_to_entry_range
+        and input.round_completion_info.hb_index_to_entry_range.get(str(hb_idx))
+        is not None
+    )
+
+
+def _get_all_deletes(
```
Zyiqin-Miranda, line 62: We probably need to consider the DELETE delta size too. If the delete delta is too large, it should affect the remote task memory requirement to make sure it doesn't cause OOM during merge. Maybe we could have a separate PR to after testing on production deltas and address this potential issue.

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -279,21 +373,15 @@ def _compact_tables(
...
 
     compacted_table = None
-
-    if (
-        input.round_completion_info
-        and input.round_completion_info.hb_index_to_entry_range
-        and input.round_completion_info.hb_index_to_entry_range.get(str(hb_idx))
-        is not None
-    ):
+    if _does_hash_bucket_idx_have_compacted_table(input, hb_idx):
```
Zyiqin-Miranda, line 376: Is this if `_does_hash_bucket_idx_have_compacted_table(input, hb_idx)` duplicates above  `if dfe_list is None and _does_hash_bucket_idx_have_compacted_table(input, hb_idx)` work and compacted_table get downloaded twice?

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -168,32 +242,38 @@ def _download_compacted_table(
...
             **deltacat_storage_kwargs,
         )
-
         tables.append(table)
-
-    return pa.concat_tables(tables)
+    compacted_table = pa.concat_tables(tables)
+    if deletes_to_apply_by_stream_positions:
+        for delete_table in _get_all_deletes(deletes_to_apply_by_stream_positions):
+            compacted_table, number_of_rows_dropped = drop_rows_affected_by_deletes(
```
Zyiqin-Miranda, line 268: Could we separate the `drop_rows_affected_by_deletes` apart from `_download_compacted_table` to another function like `drop_rows_from_compacted_table` so that it's clearer we are deleting from compacted table right after we download it?

pfaraone: Good idea - addressing this in the next commit

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,92 @@
...
+            deletes_at_this_stream_position.extend(delete_dataset)
+        consolidated_deletes: pa.Table = pa.concat_tables(
+            deletes_at_this_stream_position
+        )
+        stream_position_of_earliest_delete_in_sequence: int = uniform_deltas[
+            window_start
+        ].stream_position
+        deletes_obj_ref_by_stream_position[
+            stream_position_of_earliest_delete_in_sequence
+        ] = ray.put(consolidated_deletes)
```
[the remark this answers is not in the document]
pfaraone: Now using the parameterized object_store

```
        deletes_obj_ref_by_stream_position[
            stream_position_of_earliest_delete_in_sequence
        ] = params.object_store.put(consolidated_deletes)
```

pfaraone:
> 2. Are we separating out the Delete operation in driver in this implementation? For example, For deltas coming in the ? > order of Upsert, Delete, Upsert. We will only compact up to where Delete delta stream position? Cause I see in merge step, > we just sort by stream_position and delete accordingly.

These DeltaCAT changes are sufficient and general enough to support DELETEs without driver changes. In the (Upsert, 100), (Delete,105), (Upsert,110) example, 
1. All input deltas are discovered
2. During prepare_delete, a single entry is added to `deletes_obj_ref_by_stream_position` of {(105, Delete table ref)}. The uniform delta now only contains Upsert
3. During the `merge` step, rows are dropped while building the incremental table and after downloading the the compacted table if they match the DELETE delta rows

deltacat/compute/compactor_v2/steps/merge.py
pfaraone: Yes - a separate PR after testing sounds ideal

raghumdani (commented):
Overall, looks like we will need top add few enhancements. Will look into unit tests once the core business logic is appropriate.

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -70,36 +90,92 @@ def _drop_delta_type_rows(table: pa.Table, delta_type: DeltaType) -> pa.Table:
...
+        )
+        for delete_column_name in delete_column_names
+    ]
+    for i, droppable in enumerate(drop_masks[: len(drop_masks) - 1]):
+        curr_drop_mask = drop_masks[i]
+        next_drop_mask = drop_masks[i + 1]
+        # logical OR the mask arrays
+        result_mask = pa.compute.or_(curr_drop_mask, next_drop_mask)
+        drop_masks[i + 1] = result_mask
+    mask_with_all_droppable = drop_masks[-1].combine_chunks()
```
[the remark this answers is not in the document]
pfaraone: The `ChunkedArray` type doesn't expose a method for counting the 'True' elements in the mask which we use for getting the number of rows dropped. 

See https://arrow.apache.org/docs/python/generated/pyarrow.BooleanArray.html

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,95 @@
...
+    window_start, window_end = 0, 0
+    non_delete_deltas = []
+    while window_end < len(uniform_deltas):
+        # skip over non-delete type deltas
+        if (
+            uniform_deltas[window_end].annotations[0].annotation_delta_type
+            is not DeltaType.DELETE
+        ):
+            non_delete_deltas.append(uniform_deltas[window_end])
+            window_start += 1
```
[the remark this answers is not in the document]
pfaraone: There's a test case for this `test_prepare_delete` - see "5-test-upsert-deletesequence-upsert-delete"

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,95 @@
...
+                delete_delta,
+                file_reader_kwargs_provider=params.read_kwargs_provider,
+                columns=delete_columns,
+                storage_type=StorageType.LOCAL,
+                max_parallelism=1,
+                **params.deltacat_storage_kwargs,
+            )
+            deletes_at_this_stream_position.extend(delete_dataset)
+        consolidated_deletes: pa.Table = pa.concat_tables(
+            deletes_at_this_stream_position
```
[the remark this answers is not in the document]
pfaraone: If there are no deletes the while statement will reach the end of the uniform deltas without ever reaching this line. So by definition reaching this line means there is at least one delete

```
# skip over non-delete type deltas
        if (
            uniform_deltas[window_end].annotations[0].annotation_delta_type
            is not DeltaType.DELETE
        ):
            non_delete_deltas.append(uniform_deltas[window_end])
            window_start += 1
            window_end = window_start
            continue
```

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -215,6 +221,10 @@ def _execute_compaction(
     if not input_deltas:
         logger.info("No input deltas found to compact.")
         return None, None, None
+    deletes_to_apply_by_stream_position = IntegerRangeDict()
+    deletes_to_apply_by_stream_position = prepare_deletes(
+        params, uniform_deltas, deletes_to_apply_by_stream_position
+    )
```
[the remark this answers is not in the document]
pfaraone: This is outdated

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,95 @@
...
+    CompactPartitionParams,
+)
+import pyarrow as pa
+from deltacat.compute.compactor import (
+    DeltaAnnotated,
+)
+
+
+def prepare_deletes(
+    params: CompactPartitionParams, uniform_deltas: List[DeltaAnnotated]
```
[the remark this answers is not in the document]
pfaraone: Done
```
...
    if not uniform_deltas:
        return uniform_deltas, None
    assert all(
        uniform_deltas[i].stream_position <= uniform_deltas[i + 1].stream_position
        for i in range(len(uniform_deltas) - 1)
    ),  "Uniform deltas must be in non-decreasing order by stream position"
...
```

deltacat/compute/compactor_v2/steps/merge.py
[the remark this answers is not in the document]
raghumdani: I thought this case occurs when all the records are deleted from compacted table.

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -49,10 +49,7 @@ def _calculate_parquet_column_size(
             if column_meta.path_in_schema in columns:
                 columns_found += 1
                 column_size += column_meta.total_uncompressed_size
-        assert columns_found == len(columns), (
-            "Columns not found in the parquet data as "
-            f"{columns_found} != {len(columns)}"
-        )
+        # NOTE: as delete deltas may only contain a subset of columns we should no longer assert that assert columns_found == len(columns)
```
raghumdani, line 52: I don't think this assertion will be invalid if we propagate right set of columns which is my another comment above.

deltacat/utils/rangedictionary.py
```diff
@@ -0,0 +1,92 @@
...
+        if not isinstance(key, int):
+            raise ValueError("key must be an int type")
+        return super().__setitem__(key, value)
+
+    def __getitem__(self, key: int) -> Any:
+        if not isinstance(key, int):
+            raise ValueError("key must be an int type")
+        try:
+            return super().__getitem__(key)
+        except KeyError as ke:
```
raghumdani, line 67: Catching exceptions is expensive. I see that all we are doing is binary search which shouldn't require throwing and catching exceptions.

deltacat/utils/rangedictionary.py
```diff
@@ -0,0 +1,92 @@
...
+            keys = list(super().keys())
+            for i, existing_key in enumerate(keys):
+                if key < existing_key:
+                    key_of_next_greater_element = keys[i]
+                    return super().__getitem__(key_of_next_greater_element)
```
raghumdani, line 72: Wait, is this linear search? I don't see a need for binary search for our use-case but for general purpose utilities like these, it's better to implement them in an optimal way as we don't know who will end up using it.

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+        stream_position_of_earliest_delete_in_sequence: int = delete_deltas_sequence[
+            0
+        ].stream_position
+        deletes_obj_ref_by_stream_position[
+            stream_position_of_earliest_delete_in_sequence
+        ] = params.object_store.put(consolidated_deletes)
+        window_start = window_end
+        # store all_deletes
+    return non_delete_deltas, deletes_obj_ref_by_stream_position
```
raghumdani, line 101: Should we always rebalance this before returning? I know we will always put the keys in increasing order but it's implicit. If for some reason that constraint is violated above, the code will return incorrect results.

deltacat/utils/rangedictionary.py
```diff
@@ -0,0 +1,92 @@
+from collections import OrderedDict
+from typing import Any, Optional, Union
+
+
+class IntegerRangeDict(OrderedDict):
```
raghumdani, line 5: Is `SortedDict` more accurate name? Also, there seems to be a [SortedDict](https://www.geeksforgeeks.org/python-sorted-containers-an-introduction/) which does exactly what you are trying to do. Can we leverage the same?

deltacat/compute/compactor_v2/steps/merge.py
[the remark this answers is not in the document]
pfaraone: Yes - it's only for logging at the moment

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Raises:
+        AssertionError: If a delete operation does not have the required properties defined.
+    """
+    if not uniform_deltas:
+        return uniform_deltas, None
+    deletes_obj_ref_by_stream_position = IntegerRangeDict()
+    window_start, window_end = 0, 0
+    non_delete_deltas = []
+    while window_end < len(uniform_deltas):
```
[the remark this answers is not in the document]
pfaraone: The intention was to to do this in one pass using a fairly conventional sliding window - so in one pass we can collect /group all the deletes, extract out the upserts in order. Point taken though - the maintainability/single responsibility principle trade-off may not be worth it

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -49,10 +49,7 @@ def _calculate_parquet_column_size(
             if column_meta.path_in_schema in columns:
                 columns_found += 1
                 column_size += column_meta.total_uncompressed_size
-        assert columns_found == len(columns), (
```
raghumdani, line 52: The delete memory needs to be accounted for in the merge tasks memory estimation as well.

deltacat/compute/compactor_v2/deletes/prepare_deletes.py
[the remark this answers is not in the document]
raghumdani: Really appreciate the thought that has gone into this.

deltacat/compute/compactor_v2/deletes/model.py
```diff
@@ -0,0 +1,157 @@
...
+        delete_file_envelopes: List[DeleteFileEnvelope],
+        *args,
+        **kwargs
+    ) -> Tuple[List[int], Dict[str, Any]]:
+        pass
+
+    @abstractmethod
+    def rebatch_df_envelopes(
+        self,
+        index_identifier: int,
```
raghumdani, line 130: These method does not seem meaningful here in DeleteStrategy.

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -306,13 +299,114 @@ def _compact_tables(
         compacted_table=compacted_table,
     )
     total_deduped_records = hb_table_record_count - len(table)
+    return table, incremental_len, total_deduped_records, merge_time
+
+
+def _compact_incremental_and_delete_table(
```
raghumdani, line 305: As discussed, let's reuse the `build_incremental_table` and `merge_tables` and `apply_deletes` from `DeleteStrategy`.

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -306,13 +299,114 @@ def _compact_tables(
...
+    input: MergeInput,
+    df_envelopes: Optional[List[DeltaFileEnvelope]],
+    hb_idx: int,
+) -> Tuple[Any, int, int]:
+    table = None
+    total_incremental_len = 0
+    total_deduped_records = 0
+    total_dropped_records = 0
+    total_merge_time = 0
+    if not df_envelopes:
```
raghumdani, line 315: We don't need to have this special case if we just create an holistic dfe list including deletes.

deltacat/storage/model/delete_parameters.py
```diff
@@ -0,0 +1,45 @@
...
+        if equality_column_names is not None:
+            delete_parameters["equality_column_names"] = equality_column_names
+        return delete_parameters
+
+    @property
+    def equality_column_names(self) -> Optional[List[str]]:
+        return self.get("equality_column_names")
+
+    @staticmethod
+    def merge_delete_parameters(
```
raghumdani, line 27: Write UT for this.

deltacat/storage/model/delete_parameters.py
```diff
@@ -0,0 +1,45 @@
...
+        delete_parameters: List[DeleteParameters],
+    ) -> Optional[DeleteParameters]:
+        def _merge_equality_column_names(equality_column_names: List[List[str]]):
+            intersection = set([equality_column_names[0]])
+            for column_name in equality_column_names[1:]:
+                intersection &= set(column_name)
+            return list(intersection)
```
raghumdani, line 34: This is not correct behavior. We cannot merge two dfes if they equality column names are different.

raghumdani (commented):
Correctness wise, looks good. As discussed, let's break it into 4 PRs and separately review and merge them over next 2-3 days.

deltacat/storage/model/delete_parameters.py
pfaraone: Fixed now

deltacat/storage/model/delete_parameters.py
pfaraone: Added

PR #274, opened 2024-03-07
Pass region into daft io-config to prevent s3 client hitting imds

samster25:
* Pass AWS region into Daft `IOConfig` which will allow daft to avoid hitting the local IMDS.
* Hitting the IMDS server under load has yielded errors such as `hyper::Error(Connect, ConnectError("dns error", Custom { kind: Uncategorized, error: "failed to lookup address information: Temporary failure in name resolution" }))`
* However to skip IMDS, we have to use a version of daft that was build after this merged https://github.com/Eventual-Inc/Daft/pull/1886 which should be `getdaft== 0.2.15`
* I will make a follow on PR that upgrades daft to this version as well


raghumdani (approved):
LGTM! Thanks for making this change on our behalf.

PR #276, opened 2024-03-11
Add wait support for IMDSv2

rkenmi:
This change fixes the IMDS health check for IMDSv2 backed EC2 instances, before generating a AWS session.

### DIFF
--- a/deltacat/aws/clients.py
+++ b/deltacat/aws/clients.py
@@ -93,7 +93,15 @@ def retrying_get(
             after=_log_attempt_number,
         ):
             with attempt:
-                resp = _get_url(url)
+                try:
+                    resp = _get_url(url)
+                except requests.exceptions.HTTPError as exception:
+                    # Unauthorized errors imply IMDSv2 is being used
+                    if exception.response.status_code == HTTPStatus.UNAUTHORIZED:
+                        return None
+
+                    raise exception
+
                 return resp
     except RetryError as re:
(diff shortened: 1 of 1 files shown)


deltacat/aws/clients.py
```diff
@@ -93,7 +93,15 @@ def retrying_get(
             after=_log_attempt_number,
         ):
             with attempt:
-                resp = _get_url(url)
+                try:
+                    resp = _get_url(url)
+                except requests.exceptions.HTTPError as exception:
+                    # Unauthorized errors imply IMDSv2 is being used
+                    if exception.response.status_code == HTTPStatus.UNAUTHORIZED:
+                        return None
```
raghumdani, line 101: I believe this would still be required.

raghumdani (commented):
LGTM! What is your proposal for pinging IMDSv2?

deltacat/aws/clients.py
rkenmi: The idea here is to allowlist this case - let the ping exit successfully, instead of fail or retry numerous times. This exception only occurs if IMDSv2 is enabled. 

> What is your proposal for pinging IMDSv2?

We will already be pinging IMDSv2 if it is enabled on the EC2 instance. The main difference with IMDSv1 and IMDSv2 is that an additional call to fetch the session token is required, which will then be passed to subsequent EC2 metadata API calls. For pinging purposes, I avoided this step.

raghumdani (approved):
LGTM!

PR #283, opened 2024-03-28
Provide resource estimates for local merge

rkenmi:
This change provides resource estimates for input deltas during local merge, so that Ray can schedule tasks responsibly when there are many concurrent jobs. It also addresses a bug where in some instances, `ray.available_resources()` may not yet report anything, causing the ClusterUtilization to throw exceptions during resource calculation and fail jobs prematurely.

### DIFF
--- a/deltacat/compute/compactor_v2/utils/task_options.py
+++ b/deltacat/compute/compactor_v2/utils/task_options.py
@@ -1,7 +1,10 @@
 import botocore
 import logging
-from typing import Dict, Optional, List, Tuple
+from typing import Dict, Optional, List, Tuple, Any
 from deltacat import logs
+from deltacat.compute.compactor_v2.model.merge_file_group import (
+    LocalMergeFileGroupsProvider,
+)
 from deltacat.types.media import ContentEncoding, ContentType
 from deltacat.types.partial_download import PartialParquetParameters
 from deltacat.storage import (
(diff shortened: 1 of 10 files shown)


deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -279,6 +282,89 @@ def merge_resource_options_provider(
...
+        estimated_da_size,
+        pk_size_bytes,
+        estimated_num_rows,
+        incremental_index_array_size,
+        debug_memory_params,
+        ray_custom_resources,
+    )
+
+
+def merge_estimate_task_options(
```
raghumdani, line 359: Can you make `TOTAL_MEMORY_BUFFER_PERCENTAGE` buffer configurable via scaling params as you already have an open CR? I know this is a different task but it may be easier for you to add it.

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -279,6 +282,89 @@ def merge_resource_options_provider(
...
+    deltacat_storage=unimplemented_deltacat_storage,
+    deltacat_storage_kwargs: Optional[Dict] = {},
+    **kwargs,
+) -> Dict:
+    index = LocalMergeFileGroupsProvider.LOCAL_HASH_BUCKET_INDEX
+    debug_memory_params = {"merge_task_index": index}
+    pk_size_bytes = estimated_da_size
+    incremental_index_array_size = estimated_num_rows * 4
+
+    if round_completion_info and compacted_delta_manifest:
```
raghumdani, line 312: Can't this code be reused from above method?

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -279,6 +282,89 @@ def merge_resource_options_provider(
...
+        previous_inflation = (
+            round_completion_info.compacted_pyarrow_write_result.pyarrow_bytes
+            / round_completion_info.compacted_pyarrow_write_result.file_bytes
+        )
+        debug_memory_params["previous_inflation"] = previous_inflation
+
+        average_record_size = (
```
raghumdani, line 319: It looks like only thing different here is the additional `estimated_da_bytes`.

raghumdani (commented):
LGTM! Please address minor comments.

deltacat/compute/compactor_v2/utils/task_options.py
rkenmi: Sure, added.

pfaraone:
LGTM

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -191,8 +195,9 @@ def hash_bucket_resource_options_provider(
     # Consider buffer
     total_memory = total_memory * (1 + TOTAL_MEMORY_BUFFER_PERCENTAGE / 100.0)
```
raghumdani, line 196: should we update here too?

raghumdani:
Can you verify if the performance regression is a false positive?

raghumdani (approved):
LGTM. Please verify if the performance regression is false positive.

rkenmi:
> LGTM. Please verify if the performance regression is false positive.

False positive, the summary was for a previous failing commit.

PR #284, opened 2024-03-28
Add memory_logs_enabled param

yankevn:
This change adds the `memory_logs_enabled` param. When set to true, low-level debug logs related to memory usage will be enabled. By using this param, we can avoid unnecessary logs on tables that do not require special attention to memory usage.

Testing:
 - tested on personal stack

### DIFF
--- a/deltacat/compute/compactor/model/compact_partition_params.py
+++ b/deltacat/compute/compactor/model/compact_partition_params.py
@@ -362,8 +362,8 @@ class CompactPartitionParams(dict):
         return self.get("memory_logs_enabled")
 
     @memory_logs_enabled.setter
-    def memory_logs_enabled(self, are_memory_logs_enabled: bool) -> None:
-        self["memory_logs_enabled"] = are_memory_logs_enabled
+    def memory_logs_enabled(self, value: bool) -> None:
+        self["memory_logs_enabled"] = value
 
     @property
     def metrics_config(self) -> Optional[MetricsConfig]:
--- a/deltacat/compute/compactor/model/compact_partition_params.py
+++ b/deltacat/compute/compactor/model/compact_partition_params.py
@@ -91,6 +91,8 @@ class CompactPartitionParams(dict):
         result.drop_duplicates = params.get("drop_duplicates", DROP_DUPLICATES)
         result.ray_custom_resources = params.get("ray_custom_resources")
 
+        result.memory_logs_enabled = params.get("memory_logs_enabled", False)
+
         result.metrics_config = params.get("metrics_config")
 
         if not importlib.util.find_spec("memray"):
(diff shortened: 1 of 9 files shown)


deltacat/compute/compactor/model/compact_partition_params.py
```diff
@@ -355,6 +357,14 @@ def sort_keys(self) -> Optional[List[SortKey]]:
     def sort_keys(self, keys: List[SortKey]) -> None:
         self["sort_keys"] = keys
 
+    @property
+    def memory_logs_enabled(self) -> bool:
+        return self.get("memory_logs_enabled")
+
+    @memory_logs_enabled.setter
+    def memory_logs_enabled(self, are_memory_logs_enabled: bool) -> None:
```
pfaraone, line 365: Use `(self, value: bool):` to be consistent with other boolean properties
```
def memory_logs_enabled(self, value: bool) -> None:
     self["memory_logs_enabled"] = value
```

deltacat/compute/compactor_v2/model/hash_bucket_input.py
```diff
@@ -22,6 +22,7 @@ def of(
         object_store: Optional[IObjectStore] = None,
         deltacat_storage=unimplemented_deltacat_storage,
         deltacat_storage_kwargs: Optional[Dict[str, Any]] = None,
+        memory_logs_enabled: bool = False,
```
raghumdani, line 25: Do not default at multiple places. It makes changing defaults harder.

deltacat/compute/compactor_v2/utils/task_options.py
```diff
@@ -298,8 +301,9 @@ def merge_resource_options_provider(
 
     total_memory = total_memory * (1 + TOTAL_MEMORY_BUFFER_PERCENTAGE / 100.0)
     debug_memory_params["total_memory_with_buffer"] = total_memory
-    logger.debug(
-        f"[Merge task {index}]: Params used for calculating merge memory: {debug_memory_params}"
-    )
+    if memory_logs_enabled:
+        logger.debug(
+            f"[Merge task {index}]: Params used for calculating merge memory: {debug_memory_params}"
+        )
```
raghumdani, line 307: Please create a higher order function on logger that can take in `do_print` and calls logger if required.

raghumdani (commented):
LGTM! Please address couple of minor comments.

deltacat/compute/compactor_v2/model/hash_bucket_input.py
yankevn: What would the alternative be? I imagine it would have to be Optional, but that would essentially be the same behavior, right?

raghumdani: Use like below:
```suggestion
        memory_logs_enabled: Optional[bool] = None,
```

raghumdani (approved):
LGTM

pfaraone (approved):
LGTM!

raghumdani (approved):

PR #285, opened 2024-04-05
Implement clear() within memcached object store

rkenmi:


deltacat/io/memcached_object_store.py
```diff
@@ -181,15 +181,29 @@ def get(self, ref: Any, *args, **kwargs) -> object:
         for chunk_index in range(chunk_count):
             ref = self._create_ref(uid, ip, chunk_index)
             chunk = client.get(ref)
+            if chunk is None:
+                raise RuntimeError(
```
raghumdani, line 185: Throw `ValueError` instead of broad `RuntimeError`. It helps to retry.

deltacat/io/object_store.py
```diff
@@ -43,7 +43,7 @@ def get_many(self, refs: List[Any], *args, **kwargs) -> List[object]:
         or may not return ordered results.
         """
 
-    def clear(self, *args, **kwargs) -> bool:
+    def clear(self, *args, **kwargs) -> None:
```
raghumdani, line 46: Why changed the signature? You can make it Optional[bool] to extend.

raghumdani (commented):
LGTM! Please address minor comment.

deltacat/io/object_store.py
rkenmi: This is generally fine if `noreply=True`. Since we're explicitly waiting for the ACK here, we should also capture the status of the flush calls.

raghumdani (approved):
LGTM

PR #280, opened 2024-03-23
[->main] Delete Strategy Implementation and Compactor V2 merge algorithm changes

pfaraone:
## Description
- Pass

## Linked PR
- https://github.com/ray-project/deltacat/pull/273
- https://github.com/ray-project/deltacat/pull/278
- https://github.com/ray-project/deltacat/pull/279

## Testing
- Added additional test cases for the single hash-bucket case, no primary key, and a test case which is compatible with compactor v1 and v2

PR #286, opened 2024-04-05
Audit logging records deleted (`recordsDeleted`)

pfaraone:
## Description
- With the changes in compactor v2 delete algorithm introduced in https://github.com/ray-project/deltacat/pull/280, the total number of records dropped is now returned from the `merge` step. This should be audit logged like record deduped (`recordsDeduped`)

### DIFF
--- a/deltacat/compute/compactor/model/compaction_session_audit_info.py
+++ b/deltacat/compute/compactor/model/compaction_session_audit_info.py
@@ -85,11 +85,11 @@ class CompactionSessionAuditInfo(dict):
         return self.get("recordsDeduped")
 
     @property
-    def records_dropped(self) -> int:
+    def records_deleted(self) -> int:
         """
         The total count of dropped records in a compaction session if delete deltas are present.
         """
-        return self.get("recordsDropped")
+        return self.get("recordsDeleted")
 
     @property
     def input_size_bytes(self) -> float:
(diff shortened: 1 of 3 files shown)


deltacat/compute/compactor/model/compaction_session_audit_info.py
```diff
@@ -84,6 +84,13 @@ def records_deduped(self) -> int:
         """
         return self.get("recordsDeduped")
 
+    @property
+    def records_dropped(self) -> int:
```
raghumdani, line 88: Should we explicitly name it to `records_deleted`? We have `records_deduped` are also dropped.

pfaraone: Sounds like a good idea. Making the change

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -532,7 +532,7 @@ def merge_input_provider(index, item):
     record_info_msg = (
         f"Hash bucket records: {total_hb_record_count},"
         f" Deduped records: {total_dd_record_count}, "
-        f" Dropped records: {total_dropped_record_count}, "
+        f" Dropped records: {total_deleted_record_count}, "
```
raghumdani, line 535: Minor: `Deleted records`

raghumdani (approved):
LGTM!

PR #289, opened 2024-04-08
Trigger copy by reference only if previous compacted table exists

raghumdani:
Earlier, we would copy by reference even if the RCF compactor delta locator stream ID does not match the provided destination stream id. This would result in an error from the catalog as the manifest entries use unique prefixes for each stream and permissions are isolated. This pull request will enable copy by reference only if stream ID matches.

### DIFF
--- a/deltacat/compute/compactor_v2/steps/merge.py
+++ b/deltacat/compute/compactor_v2/steps/merge.py
@@ -12,6 +12,7 @@ from uuid import uuid4
 from deltacat import logs
 from typing import Callable, Iterator, List, Optional, Tuple
 from deltacat.compute.compactor_v2.model.merge_result import MergeResult
+from deltacat.compute.compactor_v2.model.merge_file_group import MergeFileGroup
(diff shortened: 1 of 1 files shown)


pfaraone (approved):
LGTM!

Zyiqin-Miranda:
LGTM, thanks for the fix

Zyiqin-Miranda (approved):

DmitriGekhtman:
Hi! Starting to learn about the code in this repository.
Noticed this PR didn't include a test (not judging this to be a bad thing). What's the testing philosophy? What are the considerations for including or not including a unit test for a bug fix?

raghumdani:
We should include it in all cases, unless it is for experiments/POCs or short term hacks to reduce the throwaway effort if those fail and hence, it allows us to fail fast. What is considered a POC is often evident and is decided on a case by case basis between reviewer and publisher. As for this instance, it not supported in all parts of our system and we don't have plans to fully support it either as we only added it to perform a one-time internal migration. Thanks for calling out. We should have added a TODO in the code here to remove it later btw.

PR #292, opened 2024-04-16
Updating the `rcf_source_partition_locator` if the table is in-place compacted

pfaraone:
## Changes
- Overriding the `rcf_source_partition_locator` for the in-place compaction case. Currently, we write the round completion file based of the rebase_source_partition_locator or source_partition_locator only. In the in-place case, we should update the rcf_source_partition_locator to the newly created partition so that incremental jobs can find the round completion file


raghumdani (commented):
LGTM overall. Please address minor comments.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -627,11 +627,23 @@ def merge_input_provider(index, item):
...
         f"partition-{params.source_partition_locator.partition_values},"
         f"compacted at: {params.last_stream_position_to_compact},"
     )
+    is_inplace_compacted: bool = (
+        params.source_partition_locator == params.destination_partition_locator
+    )
+    if (
+        is_inplace_compacted
+        and compacted_partition != rcf_source_partition_locator
+    ):
```
pfaraone, line 640: Add context/document the condition and if it changes then describe how the second concurrent compaction may not succeed

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -467,27 +472,23 @@ def merge_input_provider(index, item):
...
-
-    total_dd_record_count = sum([ddr.deduped_record_count for ddr in merge_results])
-    total_deleted_record_count = sum(
-        [ddr.deleted_record_count for ddr in merge_results]
+    total_deduped_record_count += sum(
+        [ddr.deduped_record_count for ddr in merge_results]
     )
-    logger.info(
-        f"Deduped {total_dd_record_count} records and deleted {total_deleted_record_count} records..."
+    total_deleted_record_count += sum(
```
raghumdani, line 478: Why += instead of =?

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -152,7 +152,11 @@ def _execute_compaction(
     # read the results from any previously completed compaction round
     round_completion_info = None
     high_watermark = None
-    previous_compacted_delta_manifest = None
+    previous_compacted_delta_manifest: Optional[Manifest] = None
+    total_input_records_count = np.int64(0)
+    total_hb_record_count = np.int64(0)
+    total_deduped_record_count = np.int64(0)
+    total_deleted_record_count = np.int64(0)
```
raghumdani, line 159: Why initializing them at the top if you only need them at the very bottom?

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -200,6 +204,9 @@ def _execute_compaction(
     )
     if not input_deltas:
         logger.info("No input deltas found to compact.")
+        compaction_audit.set_input_records(total_input_records_count.item())
+        compaction_audit.set_records_deduped(total_deduped_record_count.item())
+        compaction_audit.set_records_deleted(total_deleted_record_count.item())
```
raghumdani, line 209: can hardcode zero to make it easier to process code

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -627,11 +631,17 @@ def merge_input_provider(index, item):
...
     )
-
-    logger.info(
-        f"partition-{params.source_partition_locator.partition_values},"
-        f"compacted at: {params.last_stream_position_to_compact},"
-    )
+    if (
+        is_inplace_compacted
+        and compacted_partition.partition_id
+        != rcf_source_partition_locator.partition_id
```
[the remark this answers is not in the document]
pfaraone: Added a note about the potential race condition here.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -629,11 +630,24 @@ def merge_input_provider(index, item):
...
         input_average_record_size_bytes=input_average_record_size_bytes,
     )
-
     logger.info(
         f"partition-{params.source_partition_locator.partition_values},"
         f"compacted at: {params.last_stream_position_to_compact},"
     )
+    is_inplace_compacted: bool = (
+        params.source_partition_locator == params.destination_partition_locator
+    )
```
raghumdani, line 639: There is a problem if two jobs run concurrently which is still not resolved. Say job 1 ran first where this check will yield true but for 2nd job, this could yield False which is result in incorrect RCF and eventually triggering rebases more often than required.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -629,11 +630,24 @@ def merge_input_provider(index, item):
...
+    if (
+        is_inplace_compacted
+        and compacted_partition.locator != rcf_source_partition_locator
+    ):
+        # NOTE: Concurrent compactions may affect the partitions pointed to in these equality checks
+        logger.warning(
```
raghumdani, line 645: This is info log as it's expected behavior for inplace.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -629,11 +630,24 @@ def merge_input_provider(index, item):
...
+            "Overriding round completion file source partition locator as in-place compacted. "
+            + f"Got compacted partition partition_id of {compacted_partition.locator.partition_id} "
+            f"and rcf source partition_id of {rcf_source_partition_locator.partition_id}."
```
raghumdani, line 648: We are missing unit tests for this functionality.

raghumdani (commented):
Looks good mostly. As a solution to concurrency problem, you can only check if partition names match and derive it as inplace.

deltacat/compute/compactor_v2/compaction_session.py
pfaraone: Now added

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -634,6 +640,22 @@ def merge_input_provider(index, item):
...
     )
+    is_inplace_compacted: bool = (
+        params.source_partition_locator.partition_values
+        == params.destination_partition_locator.partition_values
+        and params.source_partition_locator.stream_id
+        == params.destination_partition_locator.stream_id
+    )
+    if (
+        is_inplace_compacted
+        and compacted_partition.locator != rcf_source_partition_locator
```
raghumdani, line 651: Minor: 2nd check is redundant.

pfaraone: Fixed

PR #293, opened 2024-04-20
Add retries on CredentialRetrievalError

rkenmi:
Using the boto resource or client on many concurrent Ray tasks (1000+) intermittently throws a `CredentialRetrievalError: Error when retrieving credentials from Ec2InstanceMetadata: No credentials found in credential_source referenced in profile X`. This error is thrown even when the IMDS server is up and health checks have succeeded. This is likely due to query throttling as defined in the [docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instancedata-throttling). This change adds retries with exponential backoff to remedy the issue.

### DIFF
--- a/deltacat/aws/clients.py
+++ b/deltacat/aws/clients.py
@@ -4,6 +4,7 @@ from typing import Optional
 from http import HTTPStatus
 
 import boto3
+from botocore.exceptions import CredentialRetrievalError
 from boto3.exceptions import ResourceNotExistsError
 from boto3.resources.base import ServiceResource
 from botocore.client import BaseClient
@@ -15,6 +16,8 @@ from tenacity import (
     wait_fixed,
     retry_if_exception,
     stop_after_delay,
+    retry_if_exception_type,
+    wait_random_exponential,
 )
 
 from deltacat import logs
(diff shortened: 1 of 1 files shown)


deltacat/aws/clients.py
```diff
@@ -183,10 +193,10 @@ def _client(name: str, region: Optional[str], **kwargs) -> BaseClient:
 def resource_cache(name: str, region: Optional[str], **kwargs) -> ServiceResource:
     # we don't use the @lru_cache decorator because Ray can't pickle it
     cached_function = lru_cache()(_resource)
-    return cached_function(name, region, **kwargs)
+    return boto_retry_wrapper(cached_function, name, region, **kwargs)
```
raghumdani, line 196: I have a meta question. Why are we having to hit IMDS with high TPS when we are caching this resource?

rkenmi: From the error stack traces, the throttling is occurring during cache misses.

raghumdani: Ah.. Each worker is a separate process. May be that's why. I am wondering if we should instead be thinking of per-node cache. Thoughts?

rkenmi: For per-node cache, are you thinking about a Ray actor? I've found that the boto resource or session can't be cached outside of it's process due to serialization issues.

raghumdani: Yeah, we can use per-node memcached/redis/ray plasma. Feel free to pick one. We just have to make sure it's not transferred over the wire. Also, I could pickle session though.  

```
>>> s = boto3.Session()
>>> dmps = cloudpickle.dumps(s)
>>> loads = cloudpickle.loads(dmps)
```

rkenmi: The boto resource isn't serializable, but yeah you're right that the session can be serialized with the Ray pickler at least (pymemcache's default serde didn't work for me). 

I played around with sessions using memcached, but I still find the throttling issue to persist for heavy workloads. From the error traces, it seems to be due to how there will always be a credentials check when we instantiate a new service client or resource, even from a re-used session.

PR #296, opened 2024-04-30
Use class cache instead of ray actor as a more scalable approach

raghumdani:
When there are too many metrics being emitted, actor dies due to a very high load it gets. This commit will aim to use a separate feature that allows us to configure metrics while not using actors.


yankevn:
Did you test this?

raghumdani:
> Did you test this?

Yes, tested it. Scale testing in progress.

yankevn (approved):
LGTM, maybe wait until scale test finishes before merging?

rkenmi (approved):

raghumdani:
Seeing all the emf logs in the log group without any issues.

PR #301, opened 2024-05-15
Bump version to 1.1.5

pfaraone:


DmitriGekhtman (approved):
What does this do? When do we upgrade the version?

PR #302, opened 2024-05-16
Fix compaction jobs latency increase due to changes in scheduling algo in ray 2.20.0

raghumdani:
With Ray version 2.20.0, the scheduling strategy does not scale up to full utilization due the algorithm mentioned in https://docs.ray.io/en/latest/ray-core/scheduling/index.html for default strategy. We were seeing Ray only spinning up nodes to handle just 20% of tasks in queue. With this change we will be using SPREAD strategy which will proactively spin up nodes to handle all of the load in queue. Especially for merge tasks, SPREAD is the ideal strategy because we do not care about node locality. 

NOTE: Existing tests are exercised.

### DIFF
--- a/deltacat/__init__.py
+++ b/deltacat/__init__.py
@@ -44,7 +44,7 @@ from deltacat.types.tables import TableWriteMode
 
 deltacat.logs.configure_deltacat_logger(logging.getLogger(__name__))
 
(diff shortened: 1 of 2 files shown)


pfaraone (approved):
LGTM!

PR #304, opened 2024-05-18
Additional retryable exceptions for s3 put/get

rkenmi:
New transient exceptions (e.g. SSL handshake errors) were discovered from stress tests, which were incorrectly classified as non-retryable errors.

### DIFF
--- a/deltacat/aws/constants.py
+++ b/deltacat/aws/constants.py
@@ -1,13 +1,26 @@
-from typing import List, Set
-
-from botocore.exceptions import NoCredentialsError, ReadTimeoutError
+import botocore
+from typing import Set
 
 from deltacat.utils.common import env_integer, env_string
 
 DAFT_MAX_S3_CONNECTIONS_PER_FILE = env_integer("DAFT_MAX_S3_CONNECTIONS_PER_FILE", 8)
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 5)
-TIMEOUT_ERROR_CODES: List[str] = ["ReadTimeoutError", "ConnectTimeoutError"]
-RETRYABLE_PUT_OBJECT_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
-RETRYABLE_TRANSIENT_ERRORS = (NoCredentialsError, ReadTimeoutError, ConnectionError)
+TIMEOUT_ERROR_CODES: Set[str] = {"ReadTimeoutError", "ConnectTimeoutError"}
+THROTTLING_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+RETRYABLE_TRANSIENT_ERRORS = (
+    ConnectionError,
+    TimeoutError,
+    OSError,
+    botocore.exceptions.ConnectionError,
+    botocore.exceptions.HTTPClientError,
+    botocore.exceptions.NoCredentialsError,
+    botocore.exceptions.ConnectTimeoutError,
+    botocore.exceptions.ReadTimeoutError,
+)
 AWS_REGION = env_string("AWS_REGION", "us-east-1")
-RETRY_STOP_AFTER_DELAY = env_integer("RETRY_STOP_AFTER_DELAY", 10 * 60)
+UPLOAD_DOWNLOAD_RETRY_STOP_AFTER_DELAY = env_integer(
+    "UPLOAD_DOWNLOAD_RETRY_STOP_AFTER_DELAY", 10 * 60
+)
+UPLOAD_SLICED_TABLE_RETRY_STOP_AFTER_DELAY = env_integer(
+    "UPLOAD_SLICED_TABLE_RETRY_STOP_AFTER_DELAY", 30 * 60
+)
--- a/deltacat/aws/constants.py
+++ b/deltacat/aws/constants.py
@@ -5,11 +5,9 @@ from deltacat.utils.common import env_integer, env_string
 
 DAFT_MAX_S3_CONNECTIONS_PER_FILE = env_integer("DAFT_MAX_S3_CONNECTIONS_PER_FILE", 8)
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 5)
-TIMEOUT_ERROR_CODES: Set[str] = {"ReadTimeoutError", "ConnectTimeoutError"}
-THROTTLING_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+BOTO_TIMEOUT_ERROR_CODES: Set[str] = {"ReadTimeoutError", "ConnectTimeoutError"}
+BOTO_THROTTLING_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
 RETRYABLE_TRANSIENT_ERRORS = (
-    ConnectionError,
-    TimeoutError,
     OSError,
     botocore.exceptions.ConnectionError,
     botocore.exceptions.HTTPClientError,
@@ -23,4 +21,7 @@ UPLOAD_DOWNLOAD_RETRY_STOP_AFTER_DELAY = env_integer(
 )
 UPLOAD_SLICED_TABLE_RETRY_STOP_AFTER_DELAY = env_integer(
     "UPLOAD_SLICED_TABLE_RETRY_STOP_AFTER_DELAY", 30 * 60
+)
+DOWNLOAD_MANIFEST_ENTRY_RETRY_STOP_AFTER_DELAY = env_integer(
+    "DOWNLOAD_MANIFEST_ENTRY_RETRY_STOP_AFTER_DELAY", 30 * 60
 )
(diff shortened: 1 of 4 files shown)


deltacat/aws/constants.py
```diff
@@ -1,10 +1,13 @@
...
 
+from botocore.exceptions import NoCredentialsError, ReadTimeoutError
+
 from deltacat.utils.common import env_integer, env_string
 
 DAFT_MAX_S3_CONNECTIONS_PER_FILE = env_integer("DAFT_MAX_S3_CONNECTIONS_PER_FILE", 8)
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 5)
 TIMEOUT_ERROR_CODES: List[str] = ["ReadTimeoutError", "ConnectTimeoutError"]
 RETRYABLE_PUT_OBJECT_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+RETRYABLE_TRANSIENT_ERRORS = (NoCredentialsError, ReadTimeoutError, ConnectionError)
```
pfaraone, line 11: Should we stick with error codes instead of using the errors directly?

rkenmi: boto won't capture every error as a ClientError with error codes supplied, unfortunately. There was one instance where `s3.put_object` threw a SSLError as an example.

deltacat/aws/constants.py
```diff
@@ -1,10 +1,26 @@
...
 DAFT_MAX_S3_CONNECTIONS_PER_FILE = env_integer("DAFT_MAX_S3_CONNECTIONS_PER_FILE", 8)
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 5)
-TIMEOUT_ERROR_CODES: List[str] = ["ReadTimeoutError", "ConnectTimeoutError"]
-RETRYABLE_PUT_OBJECT_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+TIMEOUT_ERROR_CODES: Set[str] = {"ReadTimeoutError", "ConnectTimeoutError"}
+THROTTLING_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+RETRYABLE_TRANSIENT_ERRORS = (
+    ConnectionError,
+    TimeoutError,
+    OSError,
```
raghumdani, line 13: `ConnectionError` and `TimeoutError` are a subclass of `OSError`.

deltacat/aws/constants.py
```diff
@@ -1,10 +1,26 @@
...
+    botocore.exceptions.ConnectionError,
+    botocore.exceptions.HTTPClientError,
+    botocore.exceptions.NoCredentialsError,
+    botocore.exceptions.ConnectTimeoutError,
+    botocore.exceptions.ReadTimeoutError,
```
raghumdani, line 18: Can we also enrich it by retrieving exceptions using exception env variables? For example, we can specify `botocore.exceptions.ReadTimeoutError` as value of env variable and the code here imports it at runtime using importlib library. I wanted to do this, but never got time to do it. This feature may be way more useful now than ever because, you don't need to create a new deltacat version if you find that you missed any transient error.

deltacat/aws/constants.py
```diff
@@ -1,10 +1,26 @@
...
+from typing import Set
 
 from deltacat.utils.common import env_integer, env_string
 
 DAFT_MAX_S3_CONNECTIONS_PER_FILE = env_integer("DAFT_MAX_S3_CONNECTIONS_PER_FILE", 8)
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 5)
-TIMEOUT_ERROR_CODES: List[str] = ["ReadTimeoutError", "ConnectTimeoutError"]
-RETRYABLE_PUT_OBJECT_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+TIMEOUT_ERROR_CODES: Set[str] = {"ReadTimeoutError", "ConnectTimeoutError"}
+THROTTLING_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
```
raghumdani, line 9: Let's rename these to explicitly call out `BOTO_THROTTLING_ERROR_CODES`.

deltacat/aws/constants.py
```diff
@@ -1,10 +1,26 @@
...
+import botocore
+from typing import Set
 
 from deltacat.utils.common import env_integer, env_string
 
 DAFT_MAX_S3_CONNECTIONS_PER_FILE = env_integer("DAFT_MAX_S3_CONNECTIONS_PER_FILE", 8)
 BOTO_MAX_RETRIES = env_integer("BOTO_MAX_RETRIES", 5)
-TIMEOUT_ERROR_CODES: List[str] = ["ReadTimeoutError", "ConnectTimeoutError"]
-RETRYABLE_PUT_OBJECT_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+TIMEOUT_ERROR_CODES: Set[str] = {"ReadTimeoutError", "ConnectTimeoutError"}
```
raghumdani, line 8: Same here regarding the naming as these are botocore exception error codes.

deltacat/aws/constants.py
```diff
@@ -1,10 +1,26 @@
...
+THROTTLING_ERROR_CODES: Set[str] = {"Throttling", "SlowDown"}
+RETRYABLE_TRANSIENT_ERRORS = (
+    ConnectionError,
+    TimeoutError,
+    OSError,
+    botocore.exceptions.ConnectionError,
+    botocore.exceptions.HTTPClientError,
```
raghumdani, line 15: Where is SSLError captured?

deltacat/aws/s3u.py
```diff
@@ -257,10 +257,18 @@ def read_file(
...
+                f"Retry table download from: {s3_url} after receiving {type(e).__name__}"
+            ) from e
+        raise NonRetryableError(
+            f"Failed table download from: {s3_url} after receiving {type(e).__name__}"
+        ) from e
+    except RETRYABLE_TRANSIENT_ERRORS as e:
+        raise RetryableError(
+            f"Retry upload for: {s3_url} after receiving {type(e).__name__}"
+        ) from e
     except BaseException as e:
```
raghumdani, line 272: Now that we can dynamically set error classes via env variables, can retry only on `RetryableError` in the parent method?

deltacat/aws/s3u.py
```diff
@@ -285,7 +293,7 @@ def upload_sliced_table(
     # @retry decorator can't be pickled by Ray, so wrap upload in Retrying
     retrying = Retrying(
         wait=wait_random_exponential(multiplier=1, max=60),
-        stop=stop_after_delay(30 * 60),
+        stop=stop_after_delay(UPLOAD_SLICED_TABLE_RETRY_STOP_AFTER_DELAY),
```
raghumdani, line 296: This delay is also hard-coded on line number 441.

deltacat/tests/aws/test_s3u.py
```diff
@@ -87,6 +97,51 @@ def test_upload_throttled(self, mock_s3_client_cache):
...
+    ):
+        mock_manifest_entry.from_s3_obj_url.side_effect = OSError(
+            "Please reduce your request rate.."
+        )
+        mock_get_metadata.return_value = [mock.MagicMock()]
+        cbwp = CapturedBlockWritePaths()
+        cbwp._write_paths = ["s3_write_path"]
+        cbwp._block_refs = [mock.MagicMock()]
+        mock_captured_block_write_paths.return_value = cbwp
+        with pytest.raises(RetryError):
```
raghumdani, line 118: I always assumed it would throw the underlying exception when retries are exhausted. Should we then add `RetryError` to ray task level retries here? https://github.com/ray-project/deltacat/blob/main/deltacat/compute/compactor_v2/utils/task_options.py

raghumdani (commented):
Looks good overall. Please address few comments.

deltacat/aws/constants.py
rkenmi: Perhaps you could raise this as a separate feature request / PR? I feel that this is a more nuanced use case and we'll want to test that feature more carefully (e.g. if you try to specify an exception class that doesn't exist or is no longer used in the dependencies, we should ignore it gracefully)

deltacat/aws/constants.py
rkenmi: Good catch, this is redundant.

deltacat/aws/constants.py
rkenmi: SSLError is also a subclass of ConnectionError

deltacat/tests/aws/test_s3u.py
rkenmi: Good idea, added.

deltacat/aws/constants.py
raghumdani: it can be a separate PR but we agreed the new deltacat release should go with this feature to improve release efficiency when we find a new error we want to retry.

raghumdani (approved):
LGTM

PR #306, opened 2024-05-24
Added timeout for merge and hash bucket tasks as a symptomatic fix

raghumdani:
We are seeing a resource leak from daft `read_parquet` which halts processing and times out the whole job. However, although this issue is very hard to reproduce in a normal circumstances, it is quite prevalent in production. Hence, a symptomatic fix will be quite useful to avoid entire job failures. We can fail just the task timing out and let ray retry it on a different node. However, today ray remote call does not support timeout argument so we cannot directly timeout a task (https://github.com/ray-project/ray/issues/18916). Hence, this PR will add support a generic `timeout` decorator that works per process and continues execution if the underlying function execution takes longer than expected. The ray task will fail with `TimeoutError`.

### DIFF
--- a/deltacat/compute/compactor_v2/constants.py
+++ b/deltacat/compute/compactor_v2/constants.py
@@ -1,3 +1,5 @@
+from deltacat.utils.common import env_integer
+
 TOTAL_BYTES_IN_SHA1_HASH = 20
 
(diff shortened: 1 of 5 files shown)


DmitriGekhtman (approved):
Cool! Very elegant workaround while we wait for task timeouts make their way into Ray core.

PR #309, opened 2024-05-29
Update boto3 and s3fs dependencies, drop Python 3.7 support

DmitriGekhtman:
This PR updates boto3 and s3fs versions to their latest versions. The current versions conflict with various AWS APIs.

It also drops support for Python 3.7, which is
- Almost a year past EOL
- Not compatible with the latest boto3 version

Also updates daft.

## Note on benchmarks

Updating S3Fs introduced a regression in some of the benchmarks -- the mechanism for this is unknown at the moment, and it's not clear if this would translate to an actual performance regression in production. We're proceeding with caution.

### DIFF
--- a/deltacat/__init__.py
+++ b/deltacat/__init__.py
@@ -44,7 +44,7 @@ from deltacat.types.tables import TableWriteMode
 
 deltacat.logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-__version__ = "1.1.8"
+__version__ = "1.1.7"
 
 
 __all__ = [
--- a/deltacat/__init__.py
+++ b/deltacat/__init__.py
@@ -44,7 +44,7 @@ from deltacat.types.tables import TableWriteMode
 
 deltacat.logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-__version__ = "1.1.6"
+__version__ = "1.1.8"
 
 
 __all__ = [
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 # any changes here should also be reflected in setup.py "install_requires"
 aws-embedded-metrics == 3.2.0
 boto3 ~= 1.34
-getdaft==0.2.23
+getdaft==0.2.25
 numpy == 1.21.5
 pandas == 1.3.5
 pyarrow == 12.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ pydantic == 1.10.4
 pymemcache == 4.0.0
 ray[default] ~= 2.0
 redis == 4.6.0
-s3fs == 2024.5.0
+s3fs == 2022.2.0
 schedule == 1.2.0
 tenacity == 8.1.0
 typing-extensions == 4.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ pydantic == 1.10.4
 pymemcache == 4.0.0
 ray[default] ~= 2.0
 redis == 4.6.0
-s3fs == 2022.2.0
+s3fs == 2024.5.0
 schedule == 1.2.0
 tenacity == 8.1.0
 typing-extensions == 4.4.0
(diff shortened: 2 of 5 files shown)


deltacat/__init__.py
```diff
@@ -44,7 +44,7 @@
 
 deltacat.logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-__version__ = "1.1.7"
+__version__ = "1.1.8"
```
raghumdani, line 47: Let's not bump yet. I have some other changes to go in today.

requirements.txt
```diff
@@ -1,7 +1,7 @@
 # setup.py install_requires
 # any changes here should also be reflected in setup.py "install_requires"
 aws-embedded-metrics == 3.2.0
-boto3 ~= 1.20
+boto3 ~= 1.34
 getdaft==0.2.23
```
raghumdani, line 5: Do you want to upgrade daft to `0.2.25` too?

raghumdani (commented):
Looks good mostly. Please address few comments.

deltacat/__init__.py
DmitriGekhtman: Sg, let's coordinate on the version bump.

requirements.txt
DmitriGekhtman: Will do.

deltacat/__init__.py
DmitriGekhtman: Reverted version bump.

requirements.txt
DmitriGekhtman: Upgraded daft.

DmitriGekhtman:
Test failure after daft bump -- may or may not be related -- will look into it.

DmitriGekhtman:
> [Py3.8 test failure](https://github.com/ray-project/deltacat/actions/runs/9289897697/job/25564867260?pr=309) after daft bump -- may or may not be related -- will look into it.

It succeeded on re-run. Looks like there might be a bit of test flakiness, specifically in `test_resources.py`.

raghumdani:
There seems to be performance regression. Can we know if it's a false positive?

DmitriGekhtman:
> There seems to be performance regression. Can we know if it's a false positive?

Looks like we're re-running the CI to find out.

DmitriGekhtman:
Triggered 3 runs against this branch in
https://github.com/ray-project/deltacat/pull/310
https://github.com/ray-project/deltacat/pull/311
https://github.com/ray-project/deltacat/pull/312

Triggered 3 runs against master in
https://github.com/ray-project/deltacat/pull/313
https://github.com/ray-project/deltacat/pull/314
https://github.com/ray-project/deltacat/pull/315

We can compare results for the test of interest `test_compact_partition_rebase_then_incremental[9-rebase-then-incremental-single-hash-bucket_V2`

DmitriGekhtman:
Already seeing some perf regression alerts from the runs against this branch.

DmitriGekhtman:
Ok...results from round 3 seem to have overwritten the others, clearly I don't know how GitHub Actions works, that's ok. Anyways:

Against this branch:
Py39: 1.7277032199950095 iter/s, stdev 08763625144110297
Py310: 1.6111781171268102 iter/s, stdev 0.096839517459092

Against master
Py39: 3.0060425923086536, 0.06733375464572759
Py310: 3.2338998534936736, 0.08521136159332987

DmitriGekhtman:
...downgrading daft does not appear to fix it

DmitriGekhtman:
that leaves s3fs, let me see if downgrading that fixes it (sad)

DmitriGekhtman:
> that leaves s3fs, let me see if downgrading that fixes it (sad)

looks like to downgrade s3fs, I would also have to downgrade boto3 to avoid a dependency conflict, which negates this PR entirely
let's back up to consider the options

raghumdani:
Let's take the risk and observe real perf diff in prod.

raghumdani (approved):
LGTM

PR #317, opened 2024-06-03
Remove @timeout annotation

raghumdani:
Ray is running multiple tasks on a single process which breaks the functionality as per documentation. Removed the @timeout annotation as well because it's not very useful in large-scale Ray application's context and is a dead code now. 

Note: A file level timeout is added by DAFT team in `getdaft==0.2.25` which will resolve this issue symptomatically. We will also start receiving logs which will allow us to root cause stuck daft reads.


raghumdani:
Performance regression notification can't be related due to the nature of the change.

DmitriGekhtman (approved):

PR #318, opened 2024-06-07
Modify in-place compacted check to include rebase source partition locator

yankevn:
This change adds a check for the rebase source partition locator. This is to prevent erroneously marking tables as in-place compacted during rebase jobs.

Tested by running a rebase job in personal stack.

### DIFF
--- a/deltacat/compute/compactor_v2/compaction_session.py
+++ b/deltacat/compute/compactor_v2/compaction_session.py
@@ -666,6 +666,12 @@ def _execute_compaction(
         f"compacted at: {params.last_stream_position_to_compact},"
     )
     is_inplace_compacted: bool = (
+        params.rebase_source_partition_locator
+        and params.rebase_source_partition_locator.partition_values
+        == params.destination_partition_locator.partition_values
+        and params.rebase_source_partition_locator.stream_id
+        == params.destination_partition_locator.stream_id
+    ) or (
         params.source_partition_locator.partition_values
         == params.destination_partition_locator.partition_values
         and params.source_partition_locator.stream_id
--- a/deltacat/compute/compactor_v2/compaction_session.py
+++ b/deltacat/compute/compactor_v2/compaction_session.py
@@ -123,12 +123,12 @@ def compact_partition(params: CompactPartitionParams, **kwargs) -> Optional[str]
                 f"Committing compacted partition to: {execute_compaction_result.new_compacted_partition.locator} "
                 f"using previous partition: {previous_partition.locator if previous_partition else None}"
             )
-            commited_partition: Partition = params.deltacat_storage.commit_partition(
+            committed_partition: Partition = params.deltacat_storage.commit_partition(
                 execute_compaction_result.new_compacted_partition,
                 previous_partition,
                 **params.deltacat_storage_kwargs,
             )
-            logger.info(f"Committed compacted partition: {commited_partition}")
+            logger.info(f"Committed compacted partition: {committed_partition}")
             round_completion_file_s3_url = rcf.write_round_completion_file(
                 params.compaction_artifact_s3_bucket,
                 execute_compaction_result.new_round_completion_file_partition_locator,
--- a/deltacat/compute/compactor_v2/compaction_session.py
+++ b/deltacat/compute/compactor_v2/compaction_session.py
@@ -662,19 +662,16 @@ def _execute_compaction(
     )
 
     logger.info(
-        f"partition-{params.source_partition_locator.partition_values},"
+        f"Partition-{params.source_partition_locator.partition_values},"
         f"compacted at: {params.last_stream_position_to_compact},"
     )
+    logger.info(
+        f"Checking if partition {rcf_source_partition_locator} is inplace compacted against {params.destination_partition_locator}..."
+    )
     is_inplace_compacted: bool = (
-        params.rebase_source_partition_locator
-        and params.rebase_source_partition_locator.partition_values
+        rcf_source_partition_locator.partition_values
         == params.destination_partition_locator.partition_values
-        and params.rebase_source_partition_locator.stream_id
-        == params.destination_partition_locator.stream_id
-    ) or (
-        params.source_partition_locator.partition_values
-        == params.destination_partition_locator.partition_values
-        and params.source_partition_locator.stream_id
+        and rcf_source_partition_locator.stream_id
         == params.destination_partition_locator.stream_id
     )
     if is_inplace_compacted:
--- a/deltacat/tests/compute/compact_partition_rebase_test_cases.py
+++ b/deltacat/tests/compute/compact_partition_rebase_test_cases.py
@@ -31,7 +31,6 @@ class RebaseCompactionTestCaseParams(BaseCompactorTestCase):
 
     Args:
         * (inherited from CompactorTestCase): see CompactorTestCase docstring for details
-        is_inplace: bool - argument to indicate whether to try compacting an in-place compacted table (the source table is the destination table). Also needed to control whether the destination table is created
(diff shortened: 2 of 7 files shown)


deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -666,6 +666,12 @@ def merge_input_provider(index, item):
         f"compacted at: {params.last_stream_position_to_compact},"
     )
     is_inplace_compacted: bool = (
+        params.rebase_source_partition_locator
```
raghumdani, line 669: You can use `rcf_source_partition_locator`. Also, please add a test case that fails for this scenario.

yankevn: Sure. Does `rcf_source_partition_locator` replace only `params.rebase_source_partition_locator` or does it also replace using `params.source_partition_locator`?

raghumdani: both. See how it's defined above.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -662,13 +662,16 @@ def merge_input_provider(index, item):
...
-        f"partition-{params.source_partition_locator.partition_values},"
+        f"Partition-{params.source_partition_locator.partition_values},"
         f"compacted at: {params.last_stream_position_to_compact},"
     )
+    logger.info(
+        f"Checking if partition {rcf_source_partition_locator} is inplace compacted against {params.destination_partition_locator}..."
+    )
     is_inplace_compacted: bool = (
-        params.source_partition_locator.partition_values
+        rcf_source_partition_locator.partition_values
```
raghumdani, line 672: I don't see how the existing rebase then incremental test case is passing. Please investigate.

yankevn: I believe it's because the rebase-then-incremental test cases do not set `source_partition_locator` and `destination_partition_locator` to the same value, hence no chance of being marked as in-place compacted.

raghumdani: Got it. A separate rebase then incremental where source and destination are same partition locator makes sense?

deltacat/tests/compute/compact_partition_rebase_test_cases.py
```diff
@@ -0,0 +1,89 @@
...
+class RebaseCompactionTestCaseParams(BaseCompactorTestCase):
+    """
+    A pytest parameterized test case for the `compact_partition` function with rebase and incremental compaction.
+
+    Args:
+        * (inherited from CompactorTestCase): see CompactorTestCase docstring for details
+        is_inplace: bool - argument to indicate whether to try compacting an in-place compacted table (the source table is the destination table). Also needed to control whether the destination table is created
+        rebase_expected_compact_partition_result: pa.Table - expected table after rebase compaction runs. An output that is asserted on in Rebase then Incremental unit tests
+    """
+
```
pfaraone, line 37: Needs the `is_inplace` attribute here since the parent class doesn't have it (although we could add it but this would involve a more major rewrite)

deltacat/tests/compute/test_compact_partition_rebase.py
```diff
@@ -0,0 +1,276 @@
...
+            "list_deltas_kwargs": {**ds_mock_kwargs, **{"equivalent_table_types": []}},
+            "object_store": RayPlasmaObjectStore(),
+            "pg_config": pgm,
+            "primary_keys": primary_keys,
+            "read_kwargs_provider": read_kwargs_provider_param,
+            "rebase_source_partition_locator": source_partition.locator,
+            "records_per_compacted_file": records_per_compacted_file_param,
+            "s3_client_kwargs": {},
+            "source_partition_locator": rebased_partition.locator,
+            "sort_keys": sort_keys if sort_keys else None,
```
pfaraone, line 252: I think you can use `sort_keys or None` here

deltacat/tests/compute/compact_partition_rebase_test_cases.py
yankevn: I'll just remove the comment since we don't need it

deltacat/tests/compute/test_compact_partition_rebase.py
yankevn: Would that still work if sort_keys is an empty data structure?

deltacat/tests/compute/compact_partition_rebase_test_cases.py
```diff
@@ -0,0 +1,88 @@
...
+from deltacat.tests.compute.compact_partition_test_cases import (
+    BaseCompactorTestCase,
+    with_compactor_version_func_test_param,
+)
+
+
+@dataclass(frozen=True)
+class RebaseCompactionTestCaseParams(BaseCompactorTestCase):
+    """
+    A pytest parameterized test case for the `compact_partition` function with rebase and incremental compaction.
```
raghumdani, line 30: Change doc as it will only do rebase.

deltacat/tests/compute/compact_partition_rebase_test_cases.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Args:
+        * (inherited from CompactorTestCase): see CompactorTestCase docstring for details
+        rebase_expected_compact_partition_result: pa.Table - expected table after rebase compaction runs. An output that is asserted on in Rebase then Incremental unit tests
```
raghumdani, line 34: Same here regarding doc update

deltacat/tests/compute/compact_partition_rebase_test_cases.py
yankevn: Nice catch, sorry about that!

deltacat/tests/compute/test_util_create_table_deltas_repo.py
```diff
@@ -251,6 +251,7 @@ def create_src_w_deltas_destination_rebase_w_deltas_strategy(
     ds.commit_partition(staged_partition, **ds_mock_kwargs)
 
     # get streams
+    # TODO: Add deltas to destination stream
```
raghumdani, line 254: What is this TODO?

deltacat/tests/local_deltacat_storage/__init__.py
```diff
@@ -820,33 +820,39 @@ def commit_partition(
...
         or []
     )
-    previous_partition_deltas_spos_gt: List[Delta] = [
-        delta
-        for delta in previous_partition_deltas
-        if delta.stream_position > partition_deltas[0].stream_position
-    ]
-    # handle the case if the previous partition deltas have a greater stream position than the partition_delta
-    partition_deltas = previous_partition_deltas_spos_gt + partition_deltas
+
```
raghumdani, line 830: I would still prefer separate private method for merge_and_promote for ease of understanding the code.

deltacat/tests/compute/test_util_create_table_deltas_repo.py
yankevn: In the future, we discussed making the destination stream non-empty to catch some issues with the test cases. I think it's fine if we keep the TODO to keep the task scope smaller? What do you think

raghumdani (approved):
LGTM! Please create a separate private func for merge_and_promote. We spend more than necessary to debug issues due to cyclomatic complexity of code.

deltacat/tests/local_deltacat_storage/__init__.py
yankevn: sure why not

deltacat/tests/compute/test_util_create_table_deltas_repo.py
raghumdani: This mirrors the production behavior anyway.

PR #290, opened 2024-04-08
[Exception] Add initial DeltaCatError support

Zyiqin-Miranda:
This PR add initial DeltaCat exception support. 
With this support:
DeltaCat exception can be easily identified by error code going forward.
Retryable DeltaCat error can be identified at task level and job level.

PR #319, opened 2024-06-13
Allow configuring file read timeout and copy of reference

raghumdani:
- This PR addresses few TODOs from #290. 
- Allow configuring the `file_timeout_ms`.
- Allow configuring whether to enable or disable copy by reference.

### DIFF
--- a/deltacat/compute/compactor_v2/utils/primary_key_index.py
+++ b/deltacat/compute/compactor_v2/utils/primary_key_index.py
@@ -157,7 +157,12 @@ def _optimized_group_record_batches_by_hash_bucket(
 def group_by_pk_hash_bucket(
     table: pa.Table, num_buckets: int, primary_keys: List[str]
 ) -> np.ndarray:
-    table = generate_pk_hash_column([table], primary_keys, requires_sha1=True)[0]
+    new_tables = generate_pk_hash_column([table], primary_keys, requires_hash=True)
+    assert (
+        len(new_tables) == 1
+    ), f"Expected only 1 table in the result but found {len(new_tables)}"
+
+    table = generate_pk_hash_column([table], primary_keys, requires_hash=True)[0]
 
     # group hash bucket record indices
     result = group_record_indices_by_hash_bucket(
@@ -171,7 +176,7 @@ def group_by_pk_hash_bucket(
 def generate_pk_hash_column(
     tables: List[pa.Table],
     primary_keys: Optional[List[str]] = None,
-    requires_sha1: bool = False,
+    requires_hash: bool = False,
 ) -> List[pa.Table]:
     """
     Returns a new table list after generating the primary key hash if desired.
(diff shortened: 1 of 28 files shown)


deltacat/exceptions.py
```diff
@@ -240,85 +259,91 @@ def categorize_compaction_step_exception(e: BaseException, deltacat_storage=None
...
     elif isinstance(e, OutOfMemoryError):
-        raise DependencyRayOutOfMemoryError(
-            msg=f"Ray worker Out Of Memory.",
-        ) from e
+        raise DependencyRayOutOfMemoryError("Ray worker Out Of Memory.") from e
     elif isinstance(e, RayTaskError):
-        # TODO: Further categorize wrapped error
-        raise e.cause from e
+        if e.cause is not None and isinstance(e.cause, Exception):
+            categorize_deltacat_exception(e.cause)
```
Zyiqin-Miranda, line 280: If the cause of `RayTaskError` is not included in the error categorization list, do we want to fail the compaction by raising an `UnclassifiedDeltaCatError`. There could be other transient issues that is not terminating the compaction execution.

deltacat/tests/local_deltacat_storage/__init__.py
```diff
@@ -1184,13 +1183,13 @@ def get_table_version_column_names(
     raise NotImplementedError("Fetching column names is not supported")
 
 
-def can_categorize(e: BaseException) -> bool:
-    if isinstance(e, LocalStorageError):
```
Zyiqin-Miranda, line 1188: Should we keep a high-level LocalStorageError that we can check in one if instead of checking every specifi storage errors?

deltacat/compute/compactor_v2/utils/primary_key_index.py
```diff
@@ -157,7 +157,12 @@ def _optimized_group_record_batches_by_hash_bucket(
 def group_by_pk_hash_bucket(
     table: pa.Table, num_buckets: int, primary_keys: List[str]
 ) -> np.ndarray:
-    table = generate_pk_hash_column([table], primary_keys, requires_sha1=True)[0]
+    new_tables = generate_pk_hash_column([table], primary_keys, requires_hash=True)
+    assert (
```
Zyiqin-Miranda, line 161: What's the context of adding this new assertion?

deltacat/exceptions.py
raghumdani: It is the current expected behavior. Let's me add a test case for it.

deltacat/tests/local_deltacat_storage/__init__.py
raghumdani: Done. Added.

deltacat/compute/compactor_v2/utils/primary_key_index.py
raghumdani: Something I felt should have been there to begin with.

Zyiqin-Miranda (approved):
LGTM! Thanks for addressing the TODOs from exception PR

PR #320, opened 2024-06-15
Extend deltacat interface to support Iceberg bucketing

raghumdani:
This PR contains interface changes and local deltacat support to read/writer delta partition spec. The motivation for this change is explained in a different document. However, to summarize we have a use where Delta must encapsulate Iceberg Manifest but there is no way to represent iceberg partition in DeltaCAT today. This PR adds partition spec model which allows us to specify identity and bucketing partition strategies where latter is what we will also be using with Iceberg. The interface changes are backward compatible.

### DIFF
--- a/deltacat/storage/model/partition_spec.py
+++ b/deltacat/storage/model/partition_spec.py
@@ -1,0 +1,71 @@
+from __future__ import annotations
+from typing import List, Optional, Any
+from deltacat.storage.model.transform import Transform
+
+"""
+An ordered list of partition values determining the values of
+ordered transforms specified in the partition spec.
+"""
+PartitionValues = List[Any]
+
+
+class PartitionFilter(dict):
+    """
+    This class represents a filter for partitions.
+    It is used to filter partitions based on certain criteria.
+    """
+
+    @staticmethod
+    def of(
+        partition_values: Optional[PartitionValues] = None,
+    ) -> PartitionFilter:
+        """
+        Creates a new PartitionFilter instance with the specified partition key and value.
+        """
+        partition_filter = PartitionFilter()
+        partition_filter["partitionValues"] = partition_values
+        return partition_filter
+
+    @property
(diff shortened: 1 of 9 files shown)


deltacat/storage/model/partition_spec.py
```diff
@@ -0,0 +1,71 @@
...
+        partition_filter = PartitionFilter()
+        partition_filter["partitionValues"] = partition_values
+        return partition_filter
+
+    @property
+    def partition_values(self) -> Optional[PartitionValues]:
+        return self.get("partitionValues")
+
+
+class PartitionSpec(dict):
```
pdames, line 34: To better differentiate this from a strict Iceberg PartitionSpec and make for a more straightforward future merge with the similar proposed model changes from https://github.com/pdames/deltacat/blob/iceberg/deltacat/storage/model/partition.py, I'd recommend just calling this `PartitionScheme` and including it in the existing `partition.py` together with other classes related to partitioning here.

I also think we should take the existing model there of attaching the transform to a partition key with a name and distinct ID vs. defining the spec directly as an ordered list of transforms.

pdames (changes requested):
Overall looks like good incremental progress. I think we can merge this as long as we refactor it to not modify the Redshift Manifest model directly. Next, I'll start working on merging the larger set of proposed changes to the internal storage model currently at https://github.com/pdames/deltacat/blob/iceberg/deltacat/storage/model/ with the more mature modeling of transforms in this PR.

pdames:
@raghumdani LMK if you want to merge the changes as-is then incrementally work toward the recommended changes, since that's also a viable option here.

raghumdani:
Sure, I will incrementally work towards this. Created this task: https://github.com/ray-project/deltacat/issues/322

raghumdani:
@pdames Can you approve so I can merge this?

pdames (approved):

PR #323, opened 2024-06-20
Adding metric prefix to get the right method name

raghumdani:
The @metrics annotation uses a method name on which it is annotated. In this case, it annotated over a decorator and hence `categorize_errors` will be used to emit metrics. This PR hardcodes prefix so that we don't have to change dashboards when additional annotations are added.


Zyiqin-Miranda:
LGTM! Thanks for the fix

Zyiqin-Miranda (approved):

PR #324, opened 2024-06-22
Upgrade ray and commit rcf before catalog commit

raghumdani:
This commit fixes two main issues:
1. Upgrade Ray version to >=2.20.0 which has a fix for the issue https://github.com/ray-project/ray/issues/45311. 
2. The new Ray version will also allow us to enable structured logging for ray logs.
3. Removes usage of BlockWritePathProvider as it's no longer supported in ray data as of 2.30.0
4. Fixes an unique issue when committing RCF (scenario explained below)

### Before: RCF read failure scenario

Below are runs in chronological order:

** Run 1**
1. Happy case. Nothing goes wrong. 

** Run 2**
1. Incremental run uses the previous round completion file.
2. Commits a new partition.
3. But fails to write RCF to S3 due to some reason (this is more likely when s3 throttles or the cluster is terminated due to some other issue)

** Run 3 **
1. Tries to read from the stale RCF as we didn't write new RCF in run 2. 
2. Fails to read because the catalog has already deleted the files. 

All subsequent runs fail.

### After: RCF read failure scenario

Assume Run 1 goes through. 

** Run 2 **
1. Incremental run uses the previous round completion file.
2. If RCF could not written, partition is never commit so RCF and partition are consistent. Assume that RCF was written successfully. 
3. Commit to catalog fails for some reason (same as explained above).

** Run 3 **
1. This PR ensures that previous RCF is written based on the current partition locator in the destination. So, correct RCF is read. 
2. Compaction runs successfully even if we saw some intermittent failure.


Zyiqin-Miranda:
>If RCF could not written, partition is never commit so RCF and partition are consistent. Assume that RCF was written successfully.

The order should be reversed based on the changes right? It's partition failed to commit first, then RCF also failed.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -685,9 +679,18 @@ def merge_input_provider(index, item):
             f"and rcf source partition_id of {rcf_source_partition_locator.partition_id}."
         )
         rcf_source_partition_locator = compacted_partition.locator
+
+    round_completion_file_s3_url = rcf.write_round_completion_file(
```
rkenmi, line 683: Can we also include an E2E test case for the compaction suite that covers the RCF read failure scenario?

raghumdani: I can add that.

deltacat/tests/compute/compactor_v2/test_compaction_session.py
```diff
@@ -11,80 +15,241 @@
...
+                }
+            )
+        )
+
+        bucket, backfill_key1, backfill_key2 = rcf_url.strip("s3://").split("/")
+        assert bucket == TEST_S3_RCF_BUCKET_NAME
+
+        # Now delete the RCF at new location and copy it to old location
+        # Copy the RCF from rcf_url to another location
+        s3_resource.Object(TEST_S3_RCF_BUCKET_NAME, f"{backfill_key1}.json").copy_from(
```
Zyiqin-Miranda, line 194: Which scenario will result in this copy behavior in reality? With the new changes, we will be both read/write from s3://source_partition_locator/destination_partition_locator?

deltacat/compute/compactor/utils/round_completion_file.py
```diff
@@ -12,40 +12,66 @@
...
 
-    round_completion_file_url = get_round_completion_file_s3_url(
+    # Note: we read from RCF at two different URI for backward
+    # compatibility reasons.
+    round_completion_file_url_prev = get_round_completion_file_s3_url(
         bucket,
         source_partition_locator,
     )
-    logger.info(f"reading round completion file from: {round_completion_file_url}")
+    all_uris = [round_completion_file_url_prev]
```
Zyiqin-Miranda, line 43: Sounds like the old location of just using source_partition_locator will be storing a RCF from latest successful run?

deltacat/compute/compactor/utils/round_completion_file.py
```diff
@@ -12,40 +12,66 @@
...
-        logger.info(f"read round completion info: {round_completion_info}")
+
+    for rcf_uri in all_uris:
+        logger.info(f"Reading round completion file from: {rcf_uri}")
+        result = s3_utils.download(rcf_uri, False, **s3_client_kwargs)
+        if result:
+            json_str = result["Body"].read().decode("utf-8")
+            round_completion_info = RoundCompletionInfo(json.loads(json_str))
+            logger.info(f"Read round completion info: {round_completion_info}")
+            break
```
Zyiqin-Miranda, line 61: Just based on `compaction_session.py` changes, my understanding is that we are reading/writing to the new source/destination location all the time. This break will return the first RCF we found, which will be at the old location of using old source partition locator?

deltacat/compute/compactor/utils/round_completion_file.py
```diff
@@ -12,40 +12,66 @@
...
+            logger.warn(f"Round completion file not present at {rcf_uri}")
+
     return round_completion_info
 
 
 @metrics
 def write_round_completion_file(
     bucket: Optional[str],
     source_partition_locator: Optional[PartitionLocator],
+    destination_partition_locator: Optional[PartitionLocator],
```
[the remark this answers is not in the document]
raghumdani: It is optional if `completion_url` is passed.

deltacat/tests/compute/compactor_v2/test_compaction_session.py
raghumdani: The copy is not done anywhere. This is to simulate the RCFs were written by previous commits.

deltacat/compute/compactor/utils/round_completion_file.py
raghumdani: Yes, before deployment of this code, RCFs will be stored in `s3://<bucket>/source_locator.json`

deltacat/compute/compactor/utils/round_completion_file.py
rkenmi: Since the new changes will now write the RCF to `s3://<bucket>/<source>/<destination>.json`, shouldn't we be reading RCFs based on the destination locator first, and fall back to the source if not available?

raghumdani: You are right. I had a test for this but not sure how it passed.

raghumdani: nvm, there was no test for this. Now added a test. Please take a look.

rkenmi (approved):

Zyiqin-Miranda:
Thanks for the fix!

Zyiqin-Miranda (approved):

PR #326, opened 2024-06-26
Ensure destination with None locator can be read/written

raghumdani:
- Added a test case which tests rcf with destination with None partition ID can be read/written.


pfaraone (approved):
LGTM

PR #328, opened 2024-06-28
Look for deltacat storage in both args and kwargs

raghumdani:
Earlier we were only looking for deltacat storage in args and top level keys in kwargs. However, there are cases when we have to extract values of kwargs and look for deltacat storage inside them to categorize error correctly.


yankevn (approved):
LGTM!

PR #333, opened 2024-07-03
refactored compaction_session.py

akindu-amazon:
Refactored compaction_session.py with more modular functions that are called within _execute_compaction. These functions are:

_process_merge_results: processes the results of merge and returns merged delta
_merge: produce merge results
_run_local_merge: gets called if hash_bucket_count == 1
_discover_deltas: returns uniform deltas to compact
_hash_bucket: hashes passed in uniform deltas
These functions will allow for easier support for multiple rounds for large tables, while previously compactable tables are compacted the same (all deltacat pytest tests pass).

### DIFF
--- a/deltacat/compute/compactor_v2/compaction_session.py
+++ b/deltacat/compute/compactor_v2/compaction_session.py
@@ -211,48 +211,18 @@ def _execute_compaction(
 
         logger.info(f"Round completion file: {round_completion_info}")
 
+    # discover deltas
     delta_discovery_start = time.monotonic()
-
-    input_deltas: List[Delta] = io.discover_deltas(
-        params.source_partition_locator,
-        params.last_stream_position_to_compact,
-        params.rebase_source_partition_locator,
-        params.rebase_source_partition_high_watermark,
-        high_watermark,
-        params.deltacat_storage,
-        params.deltacat_storage_kwargs,
-        params.list_deltas_kwargs,
-    )
-    if not input_deltas:
+    (
+        uniform_deltas,
+        delete_strategy,
+        delete_file_envelopes,
+        delete_file_size_bytes,
+        no_input_deltas,
+    ) = _discover_deltas(params, high_watermark, compaction_audit)
+    if no_input_deltas:
         logger.info("No input deltas found to compact.")
         return ExecutionCompactionResult(None, None, None, False)
-
-    delete_strategy: Optional[DeleteStrategy] = None
-    delete_file_envelopes: Optional[List[DeleteFileEnvelope]] = None
-    delete_file_size_bytes: int = 0
-    if contains_delete_deltas(input_deltas):
-        input_deltas, delete_file_envelopes, delete_strategy = prepare_deletes(
-            params, input_deltas
-        )
-        for delete_file_envelope in delete_file_envelopes:
-            delete_file_size_bytes += delete_file_envelope.table_size_bytes
-        logger.info(
-            f" Input deltas contain {DeltaType.DELETE}-type deltas. Total delete file size={delete_file_size_bytes}."
-            f" Total length of delete file envelopes={len(delete_file_envelopes)}"
-        )
-    uniform_deltas: List[DeltaAnnotated] = io.create_uniform_input_deltas(
-        input_deltas=input_deltas,
-        hash_bucket_count=params.hash_bucket_count,
-        compaction_audit=compaction_audit,
-        deltacat_storage=params.deltacat_storage,
-        previous_inflation=params.previous_inflation,
-        min_delta_bytes=params.min_delta_bytes_in_batch,
-        min_file_counts=params.min_files_in_batch,
-        # disable input split during rebase as the rebase files are already uniform
-        enable_input_split=params.rebase_source_partition_locator is None,
-        deltacat_storage_kwargs=params.deltacat_storage_kwargs,
-    )
-
     delta_discovery_end = time.monotonic()
 
     compaction_audit.set_uniform_deltas_created(len(uniform_deltas))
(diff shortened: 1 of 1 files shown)


deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -211,48 +211,18 @@ def _execute_compaction(
...
-        params.deltacat_storage_kwargs,
-        params.list_deltas_kwargs,
-    )
-    if not input_deltas:
+    (
+        uniform_deltas,
+        delete_strategy,
+        delete_file_envelopes,
+        delete_file_size_bytes,
+        no_input_deltas,
```
raghumdani, line 221: `no_input_deltas` arg is redundant as you can determine that from `uniform_deltas`.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -211,48 +211,18 @@ def _execute_compaction(
...
+    ) = _discover_deltas(params, high_watermark, compaction_audit)
```
raghumdani, line 222: can we rename this to `_discover_and_build_uniform_deltas`?

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -300,77 +270,30 @@ def _execute_compaction(
     if params.hash_bucket_count == 1:
         logger.info("Hash bucket count set to 1. Running local merge")
         merge_start = time.monotonic()
-        local_merge_input = generate_local_merge_input(
+        merge_results, total_input_records_count = _run_local_merge(
```
raghumdani, line 273: Ideally, I would expect this if statement to be moved inside another function `_run_merge`.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -694,3 +536,274 @@ def merge_input_provider(index, item):
         round_completion_file_s3_url,
         is_inplace_compacted,
     )
+
+
+def _process_merge_results(
```
raghumdani, line 541: reorder the methods in the same order as they are called. This avoid scrolling down reading the code. Even better is if you can move these to separate modules.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -694,3 +536,274 @@ def merge_input_provider(index, item):
...
+                object_store=params.object_store,
+                deltacat_storage=params.deltacat_storage,
+                deltacat_storage_kwargs=params.deltacat_storage_kwargs,
+                memory_logs_enabled=params.memory_logs_enabled,
+            )
+        }
+
+    all_hash_group_idx_to_obj_id = defaultdict(list)
+    all_hash_group_idx_to_size_bytes = defaultdict(int)
+    all_hash_group_idx_to_num_rows = defaultdict(int)
```
raghumdani, line 787: If you are returning these dict from this method, we would expect them to already processed.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -694,3 +536,274 @@ def merge_input_provider(index, item):
...
+        ray_task=mg.merge,
+        max_parallelism=task_max_parallelism,
+        options_provider=merge_options_provider,
+        kwargs_provider=merge_input_provider,
+    )
+    merge_invoke_end = time.monotonic()
+    logger.info(f"Getting {len(merge_tasks_pending)} merge results...")
+    merge_results: List[MergeResult] = ray.get(merge_tasks_pending)
+
+    return merge_results, merge_invoke_end
```
raghumdani, line 663: process merge results as well from each of these methods. That way compaction_session need not process them.

deltacat/compute/compactor_v2/compaction_session.py
raghumdani, line 541: This method can be used from within the `_run_merge`? That way compaction session gets the processed results to work with.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -517,47 +395,11 @@ def merge_input_provider(index, item):
 
     compaction_audit.set_records_deduped(total_dd_record_count.item())
     compaction_audit.set_records_deleted(total_deleted_record_count.item())
```
raghumdani, line 397: As a rule, keep audit setting near to the data. Hence, set audit came out of merge step in the `_run_merge` method.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -694,3 +536,274 @@ def merge_input_provider(index, item):
...
+    merge_resource_options_provider,
+    all_hash_group_idx_to_size_bytes,
+    all_hash_group_idx_to_num_rows,
+    round_completion_info,
+    previous_compacted_delta_manifest,
+    all_hash_group_idx_to_obj_id,
+    task_max_parallelism,
+    compacted_partition,
+    delete_strategy,
+    delete_file_envelopes,
```
raghumdani, line 601: As a rule, avoid redundant arguments in every new method you create. Better to create a separate class that models input whose each arg brings in new information to work with.

deltacat/compute/compactor_v2/compaction_session.py
```diff
@@ -694,3 +536,274 @@ def merge_input_provider(index, item):
...
         is_inplace_compacted,
     )
+
+
+def _process_merge_results(
+    params: CompactPartitionParams, mat_results, merge_results, compaction_audit
+):
+    for merge_result in merge_results:
+        mat_results.extend(merge_result.materialize_results)
+
```
raghumdani, line 546: With the exception of `compaction_audit`, do not modify input arguments. Also, name the `compaction_audit` arg as mutable_compaction_audit everywhere so that it can be potentially modified in different methods it's passed in. Applies everywhere.

raghumdani (changes requested):
The checks have failed. Please investigate.

deltacat/compute/compactor_v2/compaction_session.py
akindu-amazon: I'm a bit confused about this comment. Regarding input arguments, do you mean you don't want arguments that go into helper functions to be modified within the helper functions? I haven't changed any logic so far, so my assumption is that arguments that were modified previously should be modified similarly within the helper functions. And should mutable_compaction_audit be the name of the input argument to my helper functions instead of compaction_audit? Is mutable_compaction_audit supposed to replace compaction_audit everywhere in compaction_session.py? Thank you in advance.

deltacat/compute/compactor_v2/compaction_session.py
akindu-amazon: After going through each helper function call and the params present in CompactPartitionParams, it looks like task_max_parallelism was the redundant argument. This arg gets modified within _execute_compaction as shown below before going into my helper functions. Is there any danger in using the unmodified one present in params instead? Let me know if this question makes sense. 

`
task_max_parallelism: int = params.task_max_parallelism

    if params.pg_config:
        logger.info(
            "pg_config specified. Tasks will be scheduled in a placement group."
        )
        cluster_resources = params.pg_config.resource
        cluster_cpus = cluster_resources["CPU"]
        cluster_memory = cluster_resources["memory"]
        task_max_parallelism = cluster_cpus
        compaction_audit.set_total_cluster_memory_bytes(cluster_memory)
`

deltacat/compute/compactor_v2/compaction_session.py
akindu-amazon: Removed this argument. I started by checking `if not uniform_deltas` but the pytests failed. Seems like there are cases where there are `input_deltas` and `uniform_deltas` is None. Currently I return `uniform_deltas = False` and check `if uniform_deltas == False`, and the pytests pass. The linter seems to be complaining about this and suggests going back to `if not uniform_deltas` instead. Not sure if there is a better way to do this?

raghumdani: The latter? uniform_deltas is not a boolean so comparing it with boolean doesn't make sense.

deltacat/compute/compactor_v2/compaction_session.py
raghumdani: Yes, audit must be updated as soon as the data is calculated.

deltacat/compute/compactor_v2/compaction_session.py
raghumdani: The danger is that we will may use more the number of CPUs as parallelism which will fail the jobs.

deltacat/compute/compactor_v2/compaction_session.py
raghumdani: >Regarding input arguments, do you mean you don't want arguments that go into helper functions to be modified within the helper functions?

Yes

> so my assumption is that arguments that were modified previously should be modified similarly within the helper functions

Earlier they weren't input arguments to any method but now they are. 

> And should mutable_compaction_audit be the name of the input argument to my helper functions instead of compaction_audit?

Yes

> Is mutable_compaction_audit supposed to replace compaction_audit everywhere in compaction_session.py?

No, only if it's passed as method argument to let the reader know that this argument is modified within that function.

PR #334, opened 2024-07-09
Sort by dw_last_updated before materializing 

DmitriGekhtman:
When handling compaction of an upsert delta, sort the table descending by `dw_last_update_time` prior to materializing the result.


deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -161,6 +161,7 @@ def _merge_tables(
 
     final_table = pa.concat_tables(result_table_list)
     final_table = final_table.drop([sc._PK_HASH_STRING_COLUMN_NAME])
+    final_table = final_table.sort_by([("dw_last_updated", "descending")])
```
raghumdani, line 164: Can we only do this if dw_last_updated is present in the schema?

DmitriGekhtman: Sounds good, that will make the change potentially safe to merge.

raghumdani: Yes, can we also add a TODO here to retrieve sort order policy from the table version metadata instead of hard-coding here?

raghumdani: We probably need a test case with dw_last_updated column and ensure it passes too.

DmitriGekhtman: Addressed comments. Added unit test of sort order, which passes.

deltacat/compute/compactor_v2/steps/merge.py
```diff
@@ -162,6 +165,10 @@ def _merge_tables(
     final_table = pa.concat_tables(result_table_list)
     final_table = final_table.drop([sc._PK_HASH_STRING_COLUMN_NAME])
 
+    # TODO: Retrieve sort order policy from the table version metadata instead of hard-coding.
+    if DW_LAST_UPDATED_COLUMN_NAME in final_table.column_names:
+        final_table = final_table.sort_by([(DW_LAST_UPDATED_COLUMN_NAME, "descending")])
```
raghumdani, line 170: Assuming the previously compacted table will already be sorted, we can further optimize by only sorting the incremental table.

deltacat/tests/compute/compact_partition_rebase_test_cases.py
```diff
@@ -65,20 +67,68 @@ class RebaseCompactionTestCaseParams(BaseCompactorTestCase):
...
+                "col_1",
+                DW_LAST_UPDATED_COLUMN_NAME,
             ],
-            names=["pk_col_1", "sk_col_1", "sk_col_2", "col_1"],
         ),
+        expected_terminal_compact_partition_result=pa.Table.from_arrays([]),
         expected_terminal_exception=None,
         expected_terminal_exception_message=None,
         do_create_placement_group=False,
         records_per_compacted_file=DEFAULT_MAX_RECORDS_PER_FILE,
```
raghumdani, line 130: max records per file is 4M but your test case doesn't have more than 4M records. So, it only tests sorting within a single file but not across multiple files.

deltacat/tests/compute/test_compact_partition_rebase.py
```diff
@@ -274,16 +275,24 @@ def test_compact_partition_rebase_same_source_and_destination(
         compacted_delta_locator, storage_type=StorageType.LOCAL, **ds_mock_kwargs
     )
     actual_rebase_compacted_table = pa.concat_tables(tables)
```
raghumdani, line 277: We can add one assertion if the result should yield multiple tables.

raghumdani (commented):
Looks good overall. Please address few minor comments.

deltacat/tests/compute/compact_partition_rebase_test_cases.py
DmitriGekhtman: Would you recommend reducing `DEFAULT_MAX_RECORDS_PER_FILE` to a number smaller than than the `input_deltas` table?

deltacat/compute/compactor_v2/steps/merge.py
DmitriGekhtman: I think I understand. Would you mind linking a code reference?

deltacat/tests/compute/test_compact_partition_rebase.py
DmitriGekhtman: If I understand correctly, the suggestion is to add an assertion on `len(tables)`?

deltacat/tests/compute/compact_partition_rebase_test_cases.py
DmitriGekhtman: (At least, for the new test case?)

deltacat/compute/compactor_v2/steps/merge.py
DmitriGekhtman: Ah, yep, used my eyeballs.
Yes, I see. For the purposes of a POC, perhaps you can do that in a separate PR. Would indeed be good to avoid wasting time sorting a sorted table.

raghumdani: I now realize we can use sort_keys argument to sort the incremental instead of hardcoding here. This was okay for POC but we want to limit this hard coding to our internal packages.

PR #335, opened 2024-07-09
Set default disable copy by reference value to True

yankevn:
### Notes
This change sets the default copy by reference behavior to "disabled".

### Testing
 - job succeeded on personal stack

### DIFF
--- a/deltacat/compute/compactor_v2/constants.py
+++ b/deltacat/compute/compactor_v2/constants.py
@@ -41,7 +41,7 @@ DROP_DUPLICATES = True
 # size in metadata to pyarrow table size.
 PARQUET_TO_PYARROW_INFLATION = 4
 
-# By default, copy by reference is enabled
+# By default, copy by reference is disabled
 DEFAULT_DISABLE_COPY_BY_REFERENCE = True
 
 # Metric Names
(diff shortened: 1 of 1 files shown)


deltacat/compute/compactor_v2/constants.py
```diff
@@ -42,7 +42,7 @@
 PARQUET_TO_PYARROW_INFLATION = 4
 
 # By default, copy by reference is enabled
```
DmitriGekhtman, line 44: Comment would need update.

Not a comment on this CR, and it's likely infeasible to change at this point -- but it's better to have flags and variables that enable rather than negate; i.e. `ENABLE_COPY_BY_REFERENCE` rather than `DISABLE_COPY_BY_REFERENCE`. Else you get confusing double-negatives like `DISABLE_COPY_BY_REFERENCE = False`.

raghumdani:
We don't need this change once you have set it in scaling param.

yankevn:
Why not just make the changes consistent?

raghumdani:
Because we don't have to set default as True long term.

yankevn:
Sure, I guess we can hold off on setting this default then

PR #340, opened 2024-08-01
Now waiting for 10 mins instead of 5 to fetch credentials from STS

raghumdani:
- STS throttle when the new instance is spin up causing `CredentialRetrievalError`. 
- As a symptomatic fix, we are now waiting 10 mins instead of 5 mins earlier.


pfaraone:
LGTM

pfaraone (approved):
LGTM

PR #344, opened 2024-08-09
added init file to private util function directory

akindu-amazon:
Need an init file in the private directory to find compaction util function file.


raghumdani (approved):
LGTM

PR #349, opened 2024-08-29
Adding a test to assert RCF values are calculated correctly

raghumdani:
This PR ensures the RCF stats (inputInflation and averageRecordSizeBytes) are calculated based on the incremental data only.


Zyiqin-Miranda (approved):
LGTM!Thanks for the fix!

PR #350, opened 2024-08-30
Added context to the json formatter

raghumdani:
This PR adds additional context to the json formatter which allows us to log environment variables which can be used to filter right logs.


yankevn:
Do we actually set the context anywhere in this PR?

raghumdani:
> Do we actually set the context_kwargs anywhere in this PR?

It will set in the driver.

PR #352, opened 2024-09-12
Support null as a valid primary key value

raghumdani:
This PR supports tables with null primary keys. We observed one table had null primary key and the schema specifying primary key can be null. This breaks the primary key definition. However, in the short term we will follow spark compaction behavior and in the long term we will rollback to the current behavior where it's not supported.


Zyiqin-Miranda (approved):
Thanks for the fix, the test cases are very thorough.

PR #354, opened 2024-09-23
Remove dead code

raghumdani:
This PR removes dead code which involves reading all the files into memory to calculate stats.


pfaraone (approved):
LGTM!

deltacat/compute/compactor/model/compact_partition_params.py
```diff
@@ -413,6 +413,14 @@ def num_rounds(self) -> int:
     def num_rounds(self, num_rounds: int) -> None:
         self["num_rounds"] = num_rounds
 
+    @property
+    def parquet_to_pyarrow_inflation(self) -> int:
```
pfaraone, line 417: Wouldn't a float be more suitable here if we want to use non-integers?

raghumdani: Addressed in new PR.

PR #361, opened 2024-10-07
Revert daft version

yankevn:


requirements.txt
```diff
@@ -2,7 +2,7 @@
 # any changes here should also be reflected in setup.py "install_requires"
 aws-embedded-metrics == 3.2.0
 boto3 ~= 1.34
-getdaft==0.3.4
+getdaft==0.2.31
```
raghumdani, line 5: Should we roll forward instead?

raghumdani: yeah, it's in v0.3.6

raghumdani (approved):
LGTM

PR #362, opened 2024-10-08
Allow DEFAULT_V2 when avg record size is absent

raghumdani:
This PR allows us to set DEFAULT_V2 option for both parquet and utsv files by handling the null for average record sizes.

### DIFF
--- a/deltacat/compute/resource_estimation/delta.py
+++ b/deltacat/compute/resource_estimation/delta.py
@@ -42,7 +42,11 @@ def _estimate_resources_required_to_process_delta_using_previous_inflation(
     in_memory_size = (
         delta.meta.content_length * estimate_resources_params.previous_inflation
     )
-    num_rows = int(in_memory_size / estimate_resources_params.average_record_size_bytes)
+    num_rows = 0
+    if estimate_resources_params.average_record_size_bytes is not None:
+        num_rows = int(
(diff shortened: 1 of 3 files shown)


deltacat/compute/resource_estimation/delta.py
```diff
@@ -42,7 +42,11 @@ def _estimate_resources_required_to_process_delta_using_previous_inflation(
     in_memory_size = (
         delta.meta.content_length * estimate_resources_params.previous_inflation
     )
-    num_rows = int(in_memory_size / estimate_resources_params.average_record_size_bytes)
+    num_rows = 0
+    if estimate_resources_params.average_record_size_bytes is not None:
```
yankevn, line 46: Are we ok with returning `num_rows = 0` in this case?

raghumdani: It's the current behavior.

yankevn: If `estimate_resources_params.average_record_size_bytes is None`, we return `num_rows = 0` rather than getting an exception?

raghumdani: Yes, that behavior is in driver.

yankevn (approved):

PR #363, opened 2024-10-11
Hotfix for memcached noreply flag + bump version

yankevn:
Here is the relevant documentation showing `noreply` is correct (rather than `no_reply`): https://pymemcache.readthedocs.io/en/latest/apidoc/pymemcache.client.base.html#pymemcache.client.base.Client.delete_many

### DIFF
--- a/deltacat/__init__.py
+++ b/deltacat/__init__.py
@@ -44,7 +44,7 @@ from deltacat.types.tables import TableWriteMode
 
 deltacat.logs.configure_deltacat_logger(logging.getLogger(__name__))
 
-__version__ = "1.1.23"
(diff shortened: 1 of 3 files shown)


raghumdani (approved):
LGTM. Did we not test this before?

yankevn:
The problem is that our unit tests only tested the file object store, since we couldn't use a real memcached client. Not sure if you have any ideas :/

raghumdani:
> The problem is that our unit tests only tested the file object store, since we couldn't use a real memcached client. Not sure if you have any ideas :/

We could test by deploying to our personal stack.

PR #366, opened 2024-10-19
Added additional logging to prepare_deletes

pfaraone:
## Notes
- Added additional context to the AssertionError message thrown for non-monotonically increasing input deltas

### DIFF
--- a/deltacat/compute/compactor_v2/deletes/utils.py
+++ b/deltacat/compute/compactor_v2/deletes/utils.py
@@ -147,10 +147,11 @@ def prepare_deletes(
     """
     if not input_deltas:
         return PrepareDeleteResult(input_deltas, [], None)
+    logger.info(f"Preparing deletes for Partition-{params.source_partition_locator}")
     assert all(
         input_deltas[i].stream_position <= input_deltas[i + 1].stream_position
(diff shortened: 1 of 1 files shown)


deltacat/compute/compactor_v2/deletes/utils.py
```diff
@@ -147,10 +147,11 @@ def prepare_deletes(
     """
     if not input_deltas:
         return PrepareDeleteResult(input_deltas, [], None)
+    logger.info(f"Preparing deletes for Partition-{params.source_partition_locator}")
     assert all(
         input_deltas[i].stream_position <= input_deltas[i + 1].stream_position
         for i in range(len(input_deltas) - 1)
-    ), "Uniform deltas must be in non-decreasing order by stream position"
+    ), f"Uniform deltas of length: {len(input_deltas)} must be in non-decreasing order by stream position"
```
raghumdani, line 154: Is there a way to repro error faster?

pfaraone: I'm attempting reproduction but that is not dependent on this logging change

raghumdani (commented):
I do not think this would be sufficient to debug what you are trying to debug.

PR #438, opened 2024-12-30
Fliver/2.0

flliver:
## Summary

Introduce a new Dataset.from_parquet() that does schema inference using pyarrow. It provides both union and intersect schema inferencing, and adds a bunch of tests to the underlying glob_path and field_group classes to help validate intentions.

## Rationale

It's now super easy to create a rivulet dataset w/ schema from a parquet dataset (or set of parquet files) in one line.

## Changes

new Dataset.from_parquet() and supporting changes.

## Impact

No impact.

## Testing

make test && make lint w/ new tests that validate that single file, union, and intersect schema inference all work properly.

## Regression Risk

N/A

## Checklist

## Additional Notes

I haven't tried doing some of the other constructor types pyarrow provides, like parquet metadata files. It's also possible in the future there might be a better way of wrapping pyarrow datasets (there are alot of constructors for pyarrow datasets), or just take a pyarrow dataset.

### DIFF
--- a/deltacat/storage/rivulet/dataset.py
+++ b/deltacat/storage/rivulet/dataset.py
@@ -53,6 +53,42 @@ class Dataset:
         riv_schema = Schema.from_pyarrow_schema(table.schema, primary_key)
         return cls(base_uri, field_groups=[PydictFieldGroup(data, riv_schema)])
 
+    @classmethod
+    def from_parquet(cls, base_uri: str, primary_key: str, schema_mode: str = 'union'):
+        """
+        Create a Dataset from parquet files.
+
+        Args:
+            base_uri: Base URI for the dataset
+            primary_key: Field to use as primary key
+            schema_mode: Schema combination mode. Options:
+                - 'union': Use unified schema with all columns
+                - 'intersect': Use only common columns across files
+
+        Returns:
+            Dataset: New dataset instance with parquet data
+        """
+        dataset = pa.dataset.dataset(base_uri)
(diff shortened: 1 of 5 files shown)


deltacat/storage/rivulet/dataset.py
```diff
@@ -53,6 +55,50 @@ def from_pydict(cls, base_uri: str, data: Dict[str, List[Any]], primary_key: str
...
+            file_uri: Path to parquet file(s)
+            primary_key: Field to use as primary key
+            schema_mode: Schema combination mode. Options:
+                - 'union': Use unified schema with all columns
+                - 'intersect': Use only common columns across files
+
+        Returns:
+            Dataset: New dataset instance with the schema automatically inferred from the source parquet files
+        """
+        base_uri = base_uri or os.path.join(file_uri, "riv-meta")
```
pdames, line 80: `os.path.join` will produce incorrect file paths due to inclusion of '\' if run on Windows and the URI refers to, say, an S3 object. We may either want to replace this with `posixpath.join` or simply `"/".join` (together with ensuring that all input paths are normalized).

pdames (approved):
LGTM! 

We'll probably need to do a pass through our code now or later to remove all `os.path` references that aren't meant to refer strictly to local filesystems, and to start normalizing all paths that we deal with (e.g., https://github.com/ray-project/ray/blob/master/python/ray/data/datasource/path_util.py#L39).

PR #439, opened 2025-01-03
Add support for decimal256 and allow rounding decimal to scale

raghumdani:
## Summary

Arrow cannot support reading CSV using decimal256. We add support by first reading them as strings and then casting to decimal256. Secondly, we support rounding decimal values to scale specified in the schema if the decimal values in the CSV does not conform to scale. 

## Rationale

The behavior will be same as Spark compaction. 

## Changes

- added tests
- Changed pyarrow.py

## Impact

This change is backward compatible. 

## Testing

Changes were tested in a functional test suite as well reading individual problematic files in beta. 

## Regression Risk

Low

## Checklist

## Additional Notes


deltacat/utils/pyarrow.py
```diff
@@ -64,45 +76,157 @@ def _filter_schema_for_columns(schema: pa.Schema, columns: List[str]) -> pa.Sche
...
+    new_convert_options.column_types = new_schema
+    new_reader_kwargs = {**new_kwargs, "convert_options": new_convert_options}
+    arrow_table = pacsv.read_csv(*reader_args, **new_reader_kwargs)
+
+    for column_index, field in enumerate(schema):
+        if pa.types.is_decimal128(field.type) or pa.types.is_decimal256(field.type):
+            column_array = arrow_table[field.name]
+            # We always cast to decimal256 to accomodate fixed scale of 38
+            cast_to_type = pa.decimal256(
+                DECIMAL256_MAX_PRECISION, DECIMAL128_MAX_PRECISION
```
yankevn, line 151: Shouldn't this be precision and scale rather than 256 precision and 128 precision?

deltacat/utils/pyarrow.py
```diff
@@ -64,45 +76,157 @@ def _filter_schema_for_columns(schema: pa.Schema, columns: List[str]) -> pa.Sche
...
+            )
+            casted_decimal_array = pc.cast(column_array, cast_to_type)
+            # Note that scale can be negative
```
yankevn, line 154: How does the code handle it if scale is negative?

yankevn (commented):
Could you also test on a table that was failing previously?

deltacat/utils/pyarrow.py
raghumdani: we are choosing scale as decimal128 max precision. I can rename if it’s unclear.

deltacat/utils/pyarrow.py
raghumdani: Yes, pc.round can handle negative scale. Refer test cases.

raghumdani:
> Could you also test on a table that was failing previously?

Yes, tested on two tables and the reads are working fine. The resultant tables are same as what spark would produce.

deltacat/utils/pyarrow.py
yankevn: Yeah I do think this would benefit from a renaming

deltacat/utils/pyarrow.py
yankevn: Gotcha so it moves to the left of the decimal place

deltacat/utils/pyarrow.py
raghumdani: done

yankevn (approved):
LGTM

PR #440, opened 2025-01-06
Fliver/2.0 - New Dataset accessors, shift from field_group to schema

flliver:
## Summary

Major refactor of Dataset class, replace field_groups with schemas, major cleanup of Dataset, Schema, Field, etc. accessors/constructors to make them more intuitive.

## Rationale

It's now possible to have multiple overlapping schemas, schemas are separate from fields (i.e. a dataset has one set of fields, while it has many schemas). It's also possible to have multiple merge keys (replacement for primary_key), so that different parts of the dataset can be sorted/joined on different keys (in the future). 

## Changes

* Major refactor of Dataset class
* primary_key -> merge_key
* Lots more test coverage across the board
* Remove field_groups as a concept
* Refactor Schema/Field/Dataset/etc to be simpler @dataclass style classes
* Added more intuitive accessors to make life easy from python CLI

## Impact

Changes the main public interface for users. 

## Testing

Lots of tests, all major classes above 90% coverage now, most above 95% coverage. Reimplemented existing tests using new constructors/accessors and everything still passes. 

## Regression Risk

I don't expect any, it didn't touch the rest of deltacat, and everything still works well (better) from CLI and unit tests. 

## Checklist

  - [X] If this is a bugfix, regression tests have been added


deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
+    def __hash__(self):
+        return hash((frozenset(self._fields.items())))
+
+    def __eq__(self, other):
+        if isinstance(other, Schema):
+            return self._fields == other._fields
+        return False
+
+    # Has a spurious type check problem in @dataclass + asdict(): https://youtrack.jetbrains.com/issue/PY-76059/Incorrect-Type-warning-with-asdict-and-Dataclass
+    def to_dict(self):
```
pdames, line 207: Minor: Stylistically, we should be consistent on indicating return type of None here (i.e., `-> None`) as some methods currently have it while others like this do not.

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
+                self.add_field(field)
+
+    def to_pyarrow(self) -> pa.Schema:
+        """
+        Convert the Schema to a PyArrow schema.
+
+        Returns:
+            pyarrow.schema: A PyArrow schema representation of this Schema.
+        """
+        # TODO: Should we track merge_keys as it goes to/from pyarrow?
```
pdames, line 248: My opinion: yes. You can see how merge keys and other similar schema annotations are already tracked via https://github.com/ray-project/deltacat/blob/2.0/deltacat/storage/model/schema.py#L345 and https://github.com/ray-project/deltacat/blob/2.0/deltacat/storage/model/schema.py#L252. This metadata is also preserved to/from Parquet. Note that `field_id` at https://github.com/ray-project/deltacat/blob/2.0/deltacat/storage/model/schema.py#L18 is an existing metadata field convention already used by Parquet and Iceberg.

pdames (approved):
As a high-level comment, we should also start investigating potential integration points with Ray Datasets, since the core functionality of "read a lot of different file types into a block-based dataset with a common schema" feels very similar to what they have there, but adds a few unique concepts like (1) field groups, (2) SST-based indexing, and (3) a unified multi-format read API.

One option may be to build this by extending Ray Dataset together with Ray Datasource/Datasink. While this would come with a tighter implicit coupling to Ray, it seems like the alternative looks like rewriting, adopting, and/or depending on a lot of the same internal procedures currently used by Ray Datasets just to support a local execution mode without Ray.

deltacat/__init__.py
```diff
@@ -82,6 +83,8 @@
     "ContentType",
     "ContentEncoding",
     "DistributedDataset",
+    "Dataset",
```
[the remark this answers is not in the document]
mcember: I haven't looked through deltacat enough, but when interacting with the dataset why should it need to be local vs distributed? Shouldn't you instantiate a local or distributed reader or writer, and other aspects like which filesystem the dataset uses is configurable?

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
-from deltacat.storage.rivulet.fs.file_location_provider import FileLocationProvider
 from deltacat.storage.rivulet.writer.memtable_dataset_writer import (
     MemtableDatasetWriter,
 )
 
+import pyarrow as pa
+
+
+class FieldsAccessor:
+    """Accessor class used to make it easy to do actions like dataset.fields['name'] to work with fields in the Dataset.
```
mcember, line 25: Is there a reason why we don't cut out the middleman and have dataset __getitem__ return the Field, so you can just do dataset['name']?

deltacat/storage/rivulet/dataset.py
```diff
@@ -91,197 +215,183 @@ def from_parquet(
...
+                if merge_keys is not None:
+                    expected_merge_key_status = name in merge_keys
+                    if processed_field.is_merge_key != expected_merge_key_status:
+                        raise TypeError(
+                            f"Merge key status conflict for field '{name}': "
+                            f"Provided as merge key: {expected_merge_key_status}, "
+                            f"Field's current status: {processed_field.is_merge_key}. "
+                            f"Merge keys should only be defined if raw (name, Datatype) tuples are used."
+                        )
+            else:
```
thesalus, line 312: Super Nitpick: the resultant error message has a weird colon structure (`xxx: yyy: a, zzz: b`). Perhaps `xxx: yyy=a, zzz=b`?

Minor: the error message seems to be mixing 2 separate unexpected conditions (lines 308 & 310 seem to refer to an old failure mode).

deltacat/storage/rivulet/dataset.py
```diff
@@ -91,197 +215,183 @@ def from_parquet(
...
-        if self._schema is None:
-            raise AttributeError("Schema is not set")
-        return self._schema
 
-    def scan(self, query: QueryExpression = QueryExpression()) -> DataScan:
-        dataset_reader = DatasetReader(self._metastore)
-        return DataScan(self.schema, query, dataset_reader)
+        # If this is a new schema it must have a merge key defined.
+        # if schema_name not in self._schemas and merge_keys is None:
+        #    raise ValueError(f"New schemas must have at least one merge_key defined.")
```
thesalus, line 329: Minor: is this guard condition still relevant?

deltacat/storage/rivulet/dataset.py
```diff
@@ -91,197 +215,183 @@ def from_parquet(
...
+            self._schemas[schema_name].add_field(self.schemas["all"][name])
+
+    def add_fields(
         self,
-        field_groups: Union[FieldGroup, Iterable[FieldGroup]] = None,
-        file_format: str | None = None,
-    ) -> DatasetWriter:
-        """Create a new (stateful) writer using the schema at the conjunction of given field groups.
+        fields: Iterable[Tuple[str, Datatype] | Field],
+        schema_name: str = "default",
```
thesalus, line 265: Minor: should we pull `"default"` and `"all"` into (module-level?) constants?

deltacat/storage/rivulet/dataset.py
```diff
@@ -91,197 +215,183 @@ def from_parquet(
...
 
-        # Save data from all field groups
-        for field_group in self.field_groups:
-            field_group.save(path)
+        # Check for any fields that already exist
+        for field in schema.values():
+            if field.name in self.schemas["all"]:
+                existing_field = self.schemas["all"][field.name]
+                if existing_field is not None and field != existing_field:
+                    raise ValueError(
```
thesalus, line 353: Minor: I think there's a bug when: 1) the input `schema` contains multiple fields with the same name and 2) `self.schemas["all"]` does not yet contain that field name. I think we'll need a similar check in the Schema constructor.

deltacat/storage/rivulet/dataset.py
```diff
@@ -91,197 +215,183 @@ def from_parquet(
...
-        riv_schema = Schema.from_pyarrow_schema(dataset_schema, primary_key)
-        return cls(base_uri, field_groups=[GlobPathFieldGroup(file_uri, riv_schema)])
+        dataset_schema = Schema.from_pyarrow(pyarrow_schema, merge_keys)
+        # TODO the file URI never gets stored/saved, do we need to do so?
+        return cls(dataset_name=name, metadata_uri=metadata_uri, schema=dataset_schema)
 
-    def add_field_group(self, field_group: FieldGroup) -> None:
-        """
-        Add a new field group to the dataset
+    def _add_fields_to_schema(
```
thesalus, line 227: Nitpick: while I don't see any obvious adverse effects currently, should this fail if `schema_name = "all"`?

deltacat/storage/rivulet/dataset.py
thesalus: I think there's potential (albeit minor) confusion between wanting the schema (with a given name) and wanting a field (with that same name).

On the other hand, is there anything that discourages a user from mutating fields directly via this accessor and accidentally bypassing the guardrails in Dataset?

deltacat/tests/storage/rivulet/schema/test_schema.py
```diff
@@ -1,116 +1,193 @@
...
+    with pytest.raises(TypeError):
+        schema["test_field"] = "invalid"
+
+
+def test_non_empty_merge_key():
+    with pytest.raises(TypeError):
+        _ = Schema([], merge_keys=["id"])
+
+
+def test_merge_schemas():
```
thesalus, line 73: Minor: I'd include a shared non-conflicting field across both schemas in this (or another) unit test.

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
+                            f"Merge key status conflict for field '{name}': "
+                            f"Provided as merge key: {expected_merge_key_status}, "
+                            f"Field's current status: {processed_field.is_merge_key}. "
+                            f"Merge keys should only be defined if raw (name, Datatype) tuples are used."
+                        )
+            else:
+                raise TypeError(f"Unexpected field type: {type(field)}")
+
+            # Locate a merge key to add, and assemble list of fields
+            if processed_field.is_merge_key:
```
thesalus, line 90: Minor: this should check if `merge_key_field is not None` (or we'll only retain the last merge key).

deltacat/storage/rivulet/__init__.py
```diff
@@ -1,7 +1,6 @@
 from .schema.schema import Schema
 from .schema.schema import Field
+from .dataset import Dataset
+from .schema.schema import Datatype
 
-__all__ = [
-    "Schema",
-    "Field",
-]
+__all__ = ["Schema", "Field", "Dataset", "Datatype"]
```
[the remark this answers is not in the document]
flliver: I was wondering why Black seemed to be bipolar! I'll use the comma trick from now on.

deltacat/__init__.py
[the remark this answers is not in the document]
flliver: I poked at it for an hour and thought about it for a few days, but they are different enough right now that it'd be a significant refactor, and I'd want to discuss it w/ you both, so I just picked the most obvious thing, which is the current name, and went with it for now.

deltacat/storage/rivulet/arrow/serializer.py
```diff
@@ -44,7 +44,7 @@ def _get_min_max_key(self, table: pa.Table) -> (Any, Any):
         """
         Get min and max primary key values from table
```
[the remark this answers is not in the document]
flliver: Fixed.

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
-    FileSystemFieldGroup,
-    GlobPathFieldGroup,
-    PydictFieldGroup,
-)
-from deltacat.storage.rivulet.logical_plan import LogicalPlan
-from deltacat.storage.rivulet.reader.data_scan import DataScan
+from deltacat.storage.rivulet.fs.file_location_provider import FileLocationProvider
 from deltacat.storage.rivulet.reader.dataset_metastore import DatasetMetastore
+from deltacat.storage.rivulet import Schema, Field
+from .schema.schema import Datatype
```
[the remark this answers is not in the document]
flliver: +1, this is the only one that's relative because the compiler complained about a circular dependency and GPT said this was how you fix it. If there's some other way let me know.

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
+        for schema in self.dataset._schemas.values():
+            if field_name in schema:
+                del schema[field_name]
+
+    def __contains__(self, field_name: str) -> bool:
+        """Allows 'field_name in dataset.fields' checks."""
+        return field_name in self.dataset.schemas["all"]
+
+    def __iter__(self):
+        return iter(self.dataset.schemas["all"].items())
```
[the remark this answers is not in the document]
flliver: Yes, but I think that will break in about 10 other places as well, so I'd prefer to do that in a separate commit where we specifically add scale tests w/ 20k+ columns.

deltacat/storage/rivulet/mvp/Table.py
```diff
@@ -60,10 +60,10 @@ def __iter__(self):
             }
             yield row_data
 
-    def to_rows_by_pk(self, pk: str) -> Dict[str, "MvpRow"]:
+    def to_rows_by_key(self, mk: str) -> Dict[str, "MvpRow"]:
         # Find the primary key field in the schema
```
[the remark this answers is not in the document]
flliver: Fixed. Did a couple of searches to find other references as well and replaced them all.

deltacat/storage/rivulet/schema/schema.py
flliver: Ooof, that was alot of work. Is there a way to add that to the linter?

deltacat/__init__.py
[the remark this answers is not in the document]
pdames: Sounds good. 

Regarding users directly instantiating local vs. distributed readers/writers, that is largely the expected end-user behavior today. To use writes as an example, a user is expected to instantiate a local or distributed reader/writer (e.g. Pandas for local or Daft for distributed), populate a dataset/dataframe to write, then provide it as input to https://github.com/ray-project/deltacat/blob/main/deltacat/catalog/delegate.py#L44. 

DeltaCAT then just uses the `LocalDataset` vs. `DistributedDataset` unioned type hints on the backend as a shortcut to branch into either local or distributed read/write handlers. Also, since all "Distributed" datasets remain perfectly functional in "local" mode (i.e., they can all run on a 1-node cluster) the general rule of thumb is that, if your dataset support distribution at all, it should just be added to the "DistributedDataset" type to telegraph that it can make use of a Ray cluster whenever one is available (i.e., `DistributedDataset` would presumably now expand to be a union of Ray Dataset, Daft Dataframe, and Rivulet Dataset).

Reads are a bit more interesting, since a user can either instantiate the desired reader directly, or let DeltaCAT automatically select the "default" or "best-fit" dataset type (i.e., based on dataset type, desired read parallelism, required memory, etc.). I'd like to think that, ideally, Rivulet becomes the preferred/default dataset read type going forward.

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
     MemtableDatasetWriter,
 )
 
+import pyarrow as pa
+
+
+class FieldsAccessor:
+    """Accessor class used to make it easy to do actions like dataset.fields['name'] to work with fields in the Dataset.
+    All field mutation and access should come through this class, or through the public helper functions in the dataset
+    class, e.g. 'add_fields()'. Do not use internal functions or directly access _schemas, as it may result in
```
mcember, line 27: Ultra nit: Good to have this caveat, but I would think it should be clear generally speaking that modifying stateful fields prefixed with "_" or "__" is unsafe. How do we think about when to give this warning or not?

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
+        if field_name not in self.dataset.schemas["all"]:
+            raise KeyError(f"Field '{field_name}' not found in dataset.")
+        return self.dataset.schemas["all"][field_name]
+
+    def __setitem__(self, field_name: str, field: Field):
+        if not isinstance(field, Field):
+            raise TypeError("Value must be a Field object")
+        self.dataset.schemas["all"][field_name] = field
+
+    def __delitem__(self, field_name: str):
```
mcember, line 44: So we just let users drop columns? 

I worry that this interface is too powerful. Have we thought about how this relates to Iceberg schema evolution? My understand is - iceberg data files use field ids, not field names. You can evolve an iceberg schema and do things like rename or re-order columns, and the query engine/catalog connector library in the back-end knows how to reconcile data file fields with the most up to date schema

I think I need a holistic understanding of how we want schema evolution to work in rivulet before grokking if this class is right

deltacat/storage/rivulet/schema/schema.py
pdames: Looks like https://pypi.org/project/flake8-annotations/ would do the trick. We should add it to the linter.

deltacat/storage/rivulet/dataset.py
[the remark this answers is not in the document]
mcember: I'm not sure what you mean by `ListResult` based pagination - wouldn't the standard approach here be to use a generator? It feels really hard though across the board to ensure that all operations that involve schemas have to be streamed end to end - we can make this method a generator but it's going to be collected in memory in other places

As a data point, the p99 schema size for andes service (size of serialized sdl) is 19 KB. The p100 is 15 MB.

deltacat/storage/rivulet/dataset.py
pdames: Agreed - modern IDEs will also throw warnings whenever you use such protected/private methods. I think this is more of a developer warning than an end-user warning, and it would probably be good to keep comments/notes/warnings designed for end-users and internal developers separate (open for suggestions about best ways of doing this - maybe there's a good way to generate separate internal-developer vs. end-user docs using sphinx?).

deltacat/storage/rivulet/dataset.py
pdames: Iceberg will also allow you to set a flag for backwards-incompatible schema evolution, but I don't think there's any corresponding guarantee that any query engine will actually be able to successfully read the resultant dataset if you use it to make backwards incompatible changes. So I think part of the answer to this question also relies on answering "what is your preferred query engine?" and "can it handle, or be made to handle, backwards-incompatible schema evolution gracefully?".

I think the worst case is that backwards-incompatible schema evolution triggers a required compaction operation to put the dataset back in a state where it's universally consumable again (which is something that we should also trigger regularly on backwards-compatible schema evolution as well, simply to reduce the read-side overhead that will build up overtime as schema changes build up).

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
+        """Simple helper to add a field when you don't have a Field object"""
+        self.dataset.add_fields(
+            fields=[(name, datatype)],
+            schema_name=schema_name,
+            merge_keys=[name] if is_merge_key else None,
+        )
+
+
+class SchemasAccessor:
+    """Accessor class used to make it easy to do actions like dataset.schemas['all'] to work with schemas in the Dataset.
```
mcember, line 82: So "all" is a reserved schema name which is a union schema? 

Did you consider whether we can instead have a method like getUnion that returns a union schema, rather than treating it like another user created schema? That feels more explicit to me and minimizes the need for validation logic to ensure users do not for instance delete "all"

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
-        self._metastore: DatasetMetastore = DatasetMetastore(
-            self._location_provider, self._file_store
+    def __init__(
+        self,
+        *,
+        dataset_name: str,
+        metadata_uri: Optional[str] = None,
+        schema: Optional[Schema] = None,
+        schema_name: Optional[str] = None,
+    ):
```
mcember, line 129: IIRC we had a conversation earlier with Patrick and he recommended using kwargs. I forget what exactly the reason, I think it was to allow us to plumb parameters through multiple methods without lots of refactors?

I like in general to have most params be keyword params. Are we aligned on this over kwargs?

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
+        """
+        Create an empty Dataset w/ optional schema. This method is typically only used for small datasets that are manually created.
+        Use the Dataset.from_*() to create a dataset from existing data.
+
+        Args:
+            dataset_name: Unique identifier for the dataset.
```
mcember, line 135: Nit: would be helpful to explain unique in what scope. Presumably once added to a higher level catalog the dataset name is used to fully qualify?

deltacat/storage/rivulet/dataset.py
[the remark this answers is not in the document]
pdames: We have a `ListResult` helper class at https://github.com/ray-project/deltacat/blob/a1b81591012c2b4636578eead5dbc1e14790596d/deltacat/storage/model/list_result.py#L11 to support bounded pagination over unbound collections. You could also certainly wrap the contained results with either a local or distributed generator (which Ray now supports, but didn't at the time we originally made the `ListResult` class).

For schemas in particular, I don't expect them to be the most problematic case (esp. for legacy analytics datasets), but I think things could get a bit more interesting here with ML datasets (e.g., similar to some catalog datasets that currently collapse the vast majority of their 10K+ fields into a single BLOB field to overcome schema limitations on traditional analytics table formats and query engines).

deltacat/storage/rivulet/dataset.py
```diff
@@ -1,83 +1,207 @@
...
+            metadata_uri: The directory to store the _metadata_folder ('.riv-meta-{dataset_name}') containing dataset metadata.
+                          If not provided, we'll use the local directory.
+
+        Private Attributes:
+            _metadata_folder (str):
+                The folder name where metadata for the dataset is kept. It will always be
+                '.riv-meta-{dataset_name}', and be stored under `metadata_uri`.
+            _schemas (dict[str, dict[str, Field]]):
```
mcember, line 143: nit: this is dict[str, Shema] below

deltacat/storage/rivulet/dataset.py
mcember: does this need to validate that the merge key being added is equal to the merge key to other schemas/field groups?

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
-        elif isinstance(field, Field):
-            self.fields[name] = field
-        else:
-            raise TypeError(
-                "The field must be an instance of the Field class or a Datatype."
-            )
+    def __init__(
+        self,
+        fields: Iterable[Tuple[str, Datatype] | Field] = None,
+        merge_keys: Optional[Iterable[str]] = None,
```
mcember, line 54: nit: should this be set?

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
+    ):
+        self._fields: Dict[str, Field] = {}
+        merge_keys = merge_keys or {}
+        if len(fields or []) == 0:
+            if len(merge_keys) > 0:
```
mcember, line 59: you should be able to remove this check and have it subsumed later when you ensure all merge keys are valid fields

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
+                )
+            elif isinstance(field, Field):
+                processed_field = field
+                name = field.name
+                # Check if merge key status conflicts
+                if len(merge_keys) > 0:
+                    expected_merge_key_status = name in merge_keys
+                    if processed_field.is_merge_key != expected_merge_key_status:
+                        raise TypeError(
+                            f"Merge key status conflict for field '{name}': "
```
mcember, line 81: this error message is confusing for end users

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
-    def add_field(self, name: str, field: Field | Datatype):
-        if isinstance(field, Datatype):
-            self.fields[name] = Field(name, field)
-        elif isinstance(field, Field):
-            self.fields[name] = field
-        else:
-            raise TypeError(
-                "The field must be an instance of the Field class or a Datatype."
-            )
+    def __init__(
```
mcember, line 51: Is the code for this method AI generated? It is fairly hard to follow and would benefit from clean up

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
+            else:
+                raise TypeError(f"Unexpected field type: {type(field)}")
+
+            # Locate a merge key to add, and assemble list of fields
+            if processed_field.is_merge_key:
+                merge_key_field = processed_field
+            else:
+                processed_fields.add(processed_field)
+
+        # if merge_key_field is None:
```
mcember, line 95: So we allow empty set of merge keys? If so we can delete this

deltacat/storage/rivulet/dataset.py
flliver: Originally it was because there was both a _schemas and a _fields, but now there's just a _schemas and all the fields are under the schemas, so it might be easier to fold at least the SchemasAccessor into the Dataset class now. The FieldsAccessor still does some additional stuff that can't just move on to the schema class, like when you call __setitem__ it adds the field to the 'all' schema automatically (which the schema class doesn't have access to).

I'm going to leave it as is for now, but playing with it on the CLI I do notice there's some additional rough edges I can smooth out in next commit, i.e. to print the fields you can do dataset.fields , but to get one for a specific schema you need to do dataset.schemas['all']._fields and it's not as pretty. I'll make sure to get that cleaned up nicely and we can dicuss what simplifications we want to the core dataset.fields vs dataset.schema['..'], etc.

deltacat/storage/rivulet/dataset.py
```diff
@@ -91,197 +215,183 @@ def from_parquet(
...
+            if isinstance(field, tuple):
+                name, datatype = field
+                processed_field = Field(
+                    name=name, datatype=datatype, is_merge_key=(name in merge_keys)
+                )
+            elif isinstance(field, Field):
+                processed_field = field
+                name = field.name
+                # Check if merge key status conflicts
+                if merge_keys is not None:
```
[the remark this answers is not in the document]
flliver: I'm not quite understanding this comment.

This check is if someone provides a Field (which already has .is_merge_key), that they don't also provide a list of merge_keys that conflicts with the merge keys defined in the fields.

Later in 319 we also check if merge_keys were provided, that they are all present in the provided fields and/or tuples.

deltacat/storage/rivulet/dataset.py
flliver: The sample error message should look like:

Merge key status conflict for field 'id': Provided as merge key: false, Field's current status: true. Merge keys should only be defined if raw (name, Datatype) tuples are used.

I've updated it to:

Merge key status conflict for field 'id'. merge_key:false, field:true. Remove merge_keys and/or match the merge_keys to the Field.is_merge_key value.

deltacat/storage/rivulet/dataset.py
flliver: no, cleaned up.

deltacat/storage/rivulet/schema/schema.py
```diff
@@ -1,207 +1,261 @@
...
-        Convert the Schema to a PyArrow schema.
-
-        Returns:
-            pyarrow.schema: A PyArrow schema representation of this Schema.
-        """
-        fields = []
-        for name, field in self.fields.items():
-            fields.append(pa.field(name, field.datatype.to_pyarrow()))
-        return pa.schema(fields)
+            for field_data in data["fields"]
```
mcember, line 114: from_dict constructs this from a dict like
{"fields": [ {"name": ..., "datatype"...}, ]

I don't get why you would need a top level key "fields" here

deltacat/tests/storage/rivulet/schema/test_schema.py
```diff
@@ -1,116 +1,193 @@
...
 
-    assert len(schema) == 3
-    assert schema["id"].datatype == Datatype("int32")
-    assert schema.primary_key.name == "id"
+def test_schema_initialization():
+    fields = [("id", Datatype.int64()), ("name", Datatype.string())]
+    schema = Schema(fields, merge_keys=["id"])
+    assert len(schema) == 2
+    assert "id" in schema.keys()
+    assert "name" in schema.keys()
```
mcember, line 18: would be nice for completeness to assert datatypes are correct too

deltacat/tests/storage/rivulet/schema/test_schema.py
```diff
@@ -1,116 +1,193 @@
...
+        ("name", Datatype.string()),
+    ]
+    with pytest.raises(TypeError):
+        Schema(fields, merge_keys=["id"])  # Merge key on here
 
 
-def test_schema_field_operations():
-    """Test adding and removing fields"""
-    schema = Schema({"id": Datatype("int32")}, primary_key="id")
+def test_add_field():
```
mcember, line 30: would be nice to get more coverage. E.g. add field when schema is non-empty. Add field which conflicts with existing field name

deltacat/storage/rivulet/dataset.py
pdames: This method signature is aligned with the recommendation that I had earlier, because the leading `*` forces all subsequent parameters to be specified by the client as keyword arguments. 

This is largely just based on the observation that it's easier to refactor code with modern IDEs when all function/method invocations are made using keyword args vs. positional args, since they remain resilient to refactoring argument order, and since an IDE is better suited to automatically/unambiguously find and update all references to renamed internal parameters but generally can't refactor positional arguments that are moved/deleted.

However, customer-facing APIs just need to go through a much more careful deprecation/update process regardless of whether you use positional or keyword args, so most of these recommendations apply more to project-internal APIs.

deltacat/storage/rivulet/schema/schema.py
flliver: (I did the work)

deltacat/storage/rivulet/dataset.py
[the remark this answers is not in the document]
flliver: k, chatting offline I figured out what you meant, and you're 100% right, I had change the init at some point to make merge_keys never be None, so this was wrong from that point on and I didn't have a test for it. I've added a test and refactored the other tests, so now should be good.

deltacat/storage/rivulet/dataset.py
flliver: Might as well, added.

deltacat/storage/rivulet/dataset.py
flliver: I think it's fine to add stuff to the all schema?

deltacat/tests/storage/rivulet/schema/test_schema.py
flliver: done

deltacat/storage/rivulet/schema/schema.py
flliver: I was able to simplify/remove this, it was left over from an intermediate state where I was trying to enforce a schema only have a single merge key, which was way too complicated and not worth while.

deltacat/storage/rivulet/dataset.py
flliver: Fair, removed since it's obvious.

deltacat/storage/rivulet/dataset.py
flliver: schema evolution will come in later commits, primarily by making schemas immutable and versioned, and tracked by deltacat.

deltacat/storage/rivulet/dataset.py
flliver: i had it that way originally, that the 'all' schema was stored in a separate object, but this keeps everything clean/simple, that the 'all' schema is just like every other schema. We could add an access method specific for the union schema at any point as a convenience, but the 'all' felt pythonic to me.

deltacat/storage/rivulet/dataset.py
flliver: yeah, the name doesn't do anything right now. That will come in a later commit.

deltacat/storage/rivulet/dataset.py
flliver: Fixed.

deltacat/storage/rivulet/schema/schema.py
flliver: I could be wrong, but I left it iterable so it can be any type of iterable?

deltacat/storage/rivulet/schema/schema.py
flliver: Fixed.

deltacat/storage/rivulet/schema/schema.py
flliver: Everything is AI generated then I refactored as needed. I think I'd write the same code from scratch though as long as the function takes an iterable of tuples or Fields. If there's a nicer way of doing:

for field in fields:
    convert tuple to field
    if field(check that there's nothing weirdly defined)
    self.add_field()
    
Let me know. i did clean it up a bit more based on feedback from the PR, so it may be better now. I think two things that may make it clunky are the separation of merge_keys from the tuples (so that every tuple doesn't have to be a thruple unnecessarily), and that it can accept tuples or Fields (trying to be pythonic).

deltacat/tests/storage/rivulet/schema/test_schema.py
flliver: done

deltacat/tests/storage/rivulet/schema/test_schema.py
flliver: Done

deltacat/storage/rivulet/schema/schema.py
mcember: ok yea that's right. It must be logically a set but we should let users pass in list or other iterable and drop duplicates

PR #444, opened 2025-01-10
feat: add demo for rivulet dataset conversion

anshumankomawar:
## Summary

This demo demonstrates creating a dataset using Deltacat, starting with a Parquet file and expanding it with additional columns and records. New data (with expanded schema) is appended without modifying the original Parquet file, enabling efficient updates. Finally, the dataset is exported to feather files.

## Rationale

Showcase current functionality and provide an example for users.

## Changes

Added demo script to deltacat examples folder (i.e. deltacat/examples/rivulet/).

## Impact

No impact to existing code.

## Testing

Unit tests (`make test`).

## Checklist


deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,80 @@
...
+    }
+]
+dataset_writer.write(new_columns_existing_rows)
+
+# Write dataset data/metadata into feather files.
+dataset_writer.flush()
+
+# Step 5 (Optional): Read data from feather file.
+#read_feather = feather.read_feather('./.riv-meta-contacts/data/<replace with generated file>')
+#print(read_feather)
```
flliver, line 80: This isn't quite right, because it's asking you to stick your nose into the private .riv-meta files. You want to either do a dataset.scan() and get back the now merged files and print those, or you want to do a (TBD) dataset.export(file-format=parquet/feather), then read that exported file.

flliver: Ideally, have an example of both.

anshumankomawar: That makes sense hadn't run into the scanner yet. Will update it to use the scanner for now (will update with the dataset export feature in the future).

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,81 @@
...
+    }
+]
+dataset_writer.write(new_columns_existing_rows)
+
+# Write dataset data/metadata into feather files.
+dataset_writer.flush()
+
+# Step 5: Read data from feather file.
+read_records = list(dataset.scan(QueryExpression()).to_arrow())
+[print(record) for record in read_records]
```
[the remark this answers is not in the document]
anshumankomawar: Ah I see what you mean, didn't realize it was being treated as an iterator already. Will update it.

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,80 @@
+import pyarrow as pa
+import pyarrow.parquet as pq
+import deltacat as dc
+
+# Step 1: Create a simple 3x3 Parquet file using pyarrow
+parquet_file_path = "../contacts.parquet"
```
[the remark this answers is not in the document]
anshumankomawar: I think it would be useful for the user to be able to view the generated files if required. But I'll go ahead and add the generated files to the gitignore.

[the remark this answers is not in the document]
anshumankomawar: I have added the configurable args as well. Going to leave it empty for now since I don't think this demo needs any configuration, but this should help stay consistent with other demos.

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -91,8 +91,8 @@ def run(**kwargs):
     dataset_writer.flush()
 
     # Step 5: Read data from feather file.
-    for record in dataset.scan().to_arrow():
-        print(record)
+    for record in dataset.scan().to_pydict():
+        print(record.values())
```
anshumankomawar, line 95: Prints values for the record instead of py_arrow representation.

deltacat/examples/rivulet/parquet_to_feather.py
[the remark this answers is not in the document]
mcember: Following up on anthony's comment - why not use notebooks for demo? It seems to be pretty standard. For example:
https://github.com/Eventual-Inc/Daft/tree/main/tutorials
https://github.com/lancedb/vectordb-recipes/tree/main/tutorials/RAG-with_MatryoshkaEmbed-Llamaindex

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,99 @@
...
+This demo showcases
+1. How to create a dataset from a Parquet file using Deltacat.
+2. How to dynamically modify a dataset schema by adding new columns.
+3. How to append new rows and update existing rows without altering the original data files.
+4. How to query and read data from the updated dataset efficiently.
+"""
+
+def run(**kwargs):
+    # Step 1: Create a simple 3x3 Parquet file using pyarrow
+    parquet_file_path = "./contacts.parquet"
```
mcember, line 21: nitpick: consider using pathlib here like `Path.cwd() / contacts.parquet`

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,99 @@
...
+import pyarrow as pa
+import pyarrow.parquet as pq
+
+from deltacat.examples.common.fixtures import (
+    store_cli_args_in_os_environ,
+)
+
+"""
+This demo showcases
+1. How to create a dataset from a Parquet file using Deltacat.
```
mcember, line 13: the fact that you are breaking it into these modular steps make me think this would be good to refactor as a notebook where each step is a cell

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,99 @@
...
+    }
+    table = pa.Table.from_pydict(data)
+    pq.write_table(table, parquet_file_path)
+
+    # Step 2: Load the Parquet file into a Dataset
+    dataset = dc.Dataset.from_parquet(
+        name="contacts",
+        file_uri=parquet_file_path,
+        metadata_uri=".",
+        merge_keys="id"
```
mcember, line 35: sanity check - merge keys here can either be a string or list of strings?

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,99 @@
...
+        "name": ["Alice", "Bob", "Charlie"],
+        "age": [25, 30, 35]
+    }
+    table = pa.Table.from_pydict(data)
+    pq.write_table(table, parquet_file_path)
+
+    # Step 2: Load the Parquet file into a Dataset
+    dataset = dc.Dataset.from_parquet(
+        name="contacts",
+        file_uri=parquet_file_path,
```
mcember, line 33: Does file path here only accept a single file? It would be more interesting to give an example using a glob path

deltacat/examples/rivulet/parquet_to_feather.py
```diff
@@ -0,0 +1,99 @@
...
+        merge_keys="id"
+    )
+
+    # Step 3: Add two new columns to the Dataset
+    dataset.add_fields([
+        ("email", dc.Datatype.string()),
+        ("is_active", dc.Datatype.bool())
+    ])
+
+    # Step 4: Append two new records, including values for the new columns. The cool thing with deltacat datasets is
```
mcember, line 44: nitpick: consider block comment

deltacat/examples/rivulet/parquet_to_feather.py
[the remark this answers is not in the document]
anshumankomawar: I followed the other existing examples in the deltacat repository, but from the links you shared notebooks look like the standard for user facing demos. Shouldn't take much to update them.

deltacat/examples/rivulet/parquet_to_feather.py
anshumankomawar: yup it can be either.

deltacat/examples/rivulet/parquet_to_feather.py
anshumankomawar: It can take in a path to parquet files as well. I'll update it to use a glob path with multiple parquet files.

anshumankomawar: Having issues with getting the dir reading to work. Will update the demo to support the file dir in a future update.

.gitignore
```diff
@@ -30,6 +30,10 @@ package-lock.json
 *.db
 pyvenv.cfg
 
+# Generated Files
+**/.riv-meta-contacts/
+**/contacts.parquet
+
 # PyInstaller
 #  Usually these files are written by a python script from a template
```
flliver, line 38: just do **/.riv-meta-* and **/*.parquet so we catch all the intermediate files.

deltacat/examples/rivulet/parquet_to_feather.ipynb
```diff
@@ -0,0 +1,180 @@
...
+   "metadata": {},
+   "cell_type": "code",
+   "source": [
+    "dataset = dc.Dataset.from_parquet(\n",
+    "    name=\"contacts\",\n",
+    "    file_uri=parquet_file_path,\n",
+    "    metadata_uri=\".\",   # local directory for writing dataset metadata\n",
+    "    merge_keys=\"id\"     # specify the merge key column\n",
+    ")\n",
+    "print(\"Loaded dataset from Parquet file.\")"
```
flliver, line 71: Don't do comments that just restate the obvious. Either remove the comment or describe what the point of a merge key is.

deltacat/examples/rivulet/parquet_to_feather.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "files alongside the original Parquet file(s) that can be easily joined with the originals."
+   ],
+   "id": "a40499cca02443bf"
+  },
+  {
+   "metadata": {},
+   "cell_type": "code",
+   "source": [
+    "# Open a new writer that will write new data to feather files\n",
+    "dataset_writer = dataset.writer(file_format=\"feather\")\n",
```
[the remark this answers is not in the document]
flliver: Off topic thought, I don't think we actually need a writer as an external class. I think we could just do dataset.write(new_rows). The only reason to have a writer is if you want to modify the output configuration of how data is written to the dataset. In that case I think we want to do something similar to add_schema(), where we add_writer() and that creates a new writer configuration and appends it to the metadata, then you can optionally do dataset.write(records, writer=<named_writer>). Then you can update the default writer for the dataset and it goes with the dataset, which matches our core tenet of the dataset is portable (and as part of that, defines its own writer configurations).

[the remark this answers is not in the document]
flliver: Note, don't do anything I said for that in this commit, just keep it in mind.

flliver (approved):
Do final nit picks and merge it!

PR #445, opened 2025-01-14
added first benchmark testing

89JBWebb:
This pull requests adds benchmark testing that tests operations like loading keys both individually and by quartile as well as traversing the dataset. The metrics for measuring these include time elapsed in milliseconds and the size of data in megabytes.


deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,7 @@
+At the moment, the dependencies for the benchmark tests are not included in the default dev environment.
+
+You can run these tests via:
+
+```shell
+hatch run benchmark:run
```
mcember, line 6: this project doesn't use hatch. Can you add the needed dependencies to dev-requirements, and if not why not?

deltacat/benchmarking/benchmark_engine.py
```diff
@@ -0,0 +1,72 @@
+import sys
```
mcember, line 1: I'm assuming all of this code is copy-pasted from the internal repo. Can you confirm or clarify if anything was modified?

get-pip.py
mcember, line 1: Please remove get-pip.py

deltacat/benchmarking/read/read1.py
```diff
@@ -0,0 +1,59 @@
+import os
```
mcember, line 1: please give this file a better name than read1

deltacat/benchmarking/benchmark_engine.py
89JBWebb: almost everything. I changed the imports to work with the current project and I changed the syntax of the unions. You can not use "|" to denote unions in python 3.9, so I changed them to "Union()"

flliver: instead can you do:

`from __future__ import annotations`

deltacat/benchmarking/benchmark_engine.py
```diff
@@ -0,0 +1,72 @@
...
+from contextlib import contextmanager
+from typing import Generator, Tuple, Iterator
+
+from deltacat.storage.rivulet.dataset import Dataset
+from deltacat.storage.rivulet.reader.query_expression import QueryExpression
+from .benchmark_report import BenchmarkStep, BenchmarkMetric
+
+
+@contextmanager
+def timed_step(description: str) -> Generator[BenchmarkStep, None, None]:
```
flliver, line 12: Switch this .benchmark_report to use full imports, unless you can't due to cyclical imports or something

flliver (changes requested):
Fix changes suggested by matt/myself and ship! make sure to run 'make lint' and 'make test' and that everything passes. Nice first commit!

deltacat/benchmarking/benchmark_engine.py
89JBWebb: I'm not sure what you mean. Would you like the code to be changed to
'''
import deltacat.storage.rivulet.reader.query_expression
import .benchmark_report
'''
in order to import everything from these two files?

deltacat/benchmarking/README.md
89JBWebb: yes I can and I have. Would you like me to delete this from the README file?

deltacat/benchmarking/benchmark_engine.py
89JBWebb: or does full imports mean using the absolute path like this line?
```
from deltacat.benchmarking.benchmark_report import BenchmarkMetric, BenchmarkStep
```

deltacat/benchmarking/read/read1.py
mcember: Ok. In that case, you can include it in the first PR which adds a new benchmark. In general you want to only check in code that is done, and code for future features/iterations can be checked in later

deltacat/benchmarking/benchmark_engine.py
flliver: 2nd one, no relative imports (anything that begins with .<dir>), so everything should begin with deltacat.<..>

deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,7 @@
+All the dependences needed to test are in dev-requirements.txt
+
+You can run these tests via:
+
+```shell
+pytest -m benchmark deltacat/benchmarking
```
mcember, line 6: nitpick: in future change we can change this to add Makefile target

pyproject.toml
```diff
@@ -0,0 +1,161 @@
+[build-system]
```
mcember, line 1: Why are you checking in pyproject.toml? We are using Makefile / setup.py / requirements.txt so we don't need pyproject.toml

deltacat/benchmarking/benchmark_engine.py
mcember: ^ From what I could tell from googling, you can't use future to use "|" syntax. So using unions here is appropriate

deltacat/benchmarking/benchmark_suite.py
```diff
@@ -0,0 +1,11 @@
+from typing import Protocol
+
+from .benchmark_report import BenchmarkRun
```
mcember, line 3: Please replace relative imports with absolute imports per earlier PR comments

mcember (changes requested):
Close but a few things to close out (remove pyproject.toml and modify imports to be absolute imports). After that it should be good to merge. Let me know when you commit those changes and we will merge together on screen share

mcember (approved):

deltacat/benchmarking/test_benchmark_pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+from deltacat.benchmarking.data.random_row_generator import RandomRowGenerator
+from deltacat.benchmarking.data.row_generator import RowGenerator
+
+
+@pytest.fixture
+def schema():
+    return Schema({
+        "id": Field("id", Datatype.int32()),
+        "source": Field("source", Datatype.string()),
+        "media": Field("media", Datatype.image("png"))}, "id")
```
thesalus, line 31: You'll need to tweak the Schema instantiation slightly to match the new constructor.

```python
return Schema(
  [
    ("id", Datatype.int32()),
    ("source", Datatype.string()),
    ("media", Datatype.image("png"),
  ],
  merge_keys=["id"])
)
```

deltacat/benchmarking/test_benchmark_pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+        pass
+
+
+class LoadAndScanSuite(BenchmarkSuite):
+    """Load some number of rows and scan"""
+
+    def __init__(self, dataset: Dataset, schema: Schema, generator, description=None):
+        self.suite = "ReadSuite"
+        self.dataset: Dataset = dataset
+        self.field_group = self.dataset.new_field_group(schema)
```
thesalus, line 50: This will need to be tweaked as well. If you were to transliterate the concepts, it might look something like:

```python
self.schema_name = "initial_load"
self.dataset.add_schema(schema, self.schema_name)
```

And you'd pass the `schema_name` into `load_and_commit()` instead of `field_group`.

thesalus (commented):
The naming and structure look rougher than I remember (through no fault of yours). I look forward to whatever direction you end up taking this.

It looks like this is slated to merge into the `fliver/2.0` branch whereas you'll want to target the [`2.0`](https://github.com/ray-project/deltacat/tree/2.0) branch. However, there were a few changes to the Dataset/Schema models that you'd need to address first (check out the unit test changes in #440 as a convenient reference).

PR #446, opened 2025-01-14
Truncate the table if the size exceeds more than 2GB

raghumdani:
## Summary

This change ensures the resultant record batches size does not exceed 2GB. Pyarrow uses int offsets and hence cannot hold arrays more than 2GB as the offset would overflow. 

## Rationale

The fix had already been applied but missed a corner case. 

## Changes

Address the corner case. 

## Impact

No impact on existing job runs. However, this change will allow onboarding tables with large parquet files.

## Testing

Added functional tests for both cases. 

## Regression Risk

Very Low

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/tests/compute/compactor_v2/utils/test_primary_key_index.py
+++ b/deltacat/tests/compute/compactor_v2/utils/test_primary_key_index.py
@@ -1,0 +1,41 @@
+import pyarrow as pa
+from deltacat.compute.compactor_v2.utils.primary_key_index import (
+    group_by_pk_hash_bucket,
+)
+
+
+class TestGroupByPkHashBucket:
+    def test_sanity(self):
+        record = pa.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+        pk = pa.array(["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"])
+        record_batch = pa.RecordBatch.from_arrays([record, pk], names=["record", "pk"])
+        table = pa.Table.from_batches([record_batch])
+        grouped_array = group_by_pk_hash_bucket(table, 3, ["pk"])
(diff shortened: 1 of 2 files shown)


pfaraone (approved):
LGTM!

Zyiqin-Miranda (approved):
LGTM! Easier fix than I thought!

deltacat/tests/compute/compactor_v2/utils/test_primary_key_index.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_when_record_batches_exceed_int_max_size(self):
+        record = pa.array(["12bytestring" * 90_000_000])
+        record_batch = pa.RecordBatch.from_arrays([record], names=["pk"])
+        table = pa.Table.from_batches([record_batch, record_batch])
+
+        grouped_array = group_by_pk_hash_bucket(table, 3, ["pk"])
+
+        assert len(grouped_array) == 3
+        assert len(grouped_array[2].to_batches()) == 2  # two record batches preserved
```
Zyiqin-Miranda, line 31: Minor: Could you add doc here of why 2 record batches for maintenance? Also, should we also assert other two buckets have how many batches?

raghumdani: same pk, so other two buckets are empty.

Zyiqin-Miranda: Maybe just add a doc string on why two batches + other two buckets are empty just for future reference?

raghumdani: done.

PR #452, opened 2025-01-17
fix block group generation when min key matches global min key

thesalus:
## Summary

Fixes an issue when scanning a rivulet dataset with a query expression where the min key matches the global min key would always return no records.

## Rationale

During testing, @anshumankomawar found that scans involving a query expression where the min key matches the global min key would incorrectly return no records (unless the min key happened to be 0).

## Changes

* Fixed guard conditions that incorrectly handled numeric key values of `0` as "Falsy" values.
* Pinned min key index to 0 when bisecting the interval tree (whereas before it would potentially use an index of `-1`).

## Impact

This should allow rivulet to return expected records even when min key matches the global min key.

## Testing

Added a new functional unit test that fails without the corresponding bugfix

`make test` to ensure no regressions.

## Regression Risk

Low. Existing functional tests passed and changes are backward compatible.

## Checklist

### DIFF
--- a/deltacat/storage/rivulet/metastore/sst_interval_tree.py
+++ b/deltacat/storage/rivulet/metastore/sst_interval_tree.py
@@ -207,8 +207,8 @@ class BlockIntervalTree:
                 f"min_key {min_key} cannot be greater than max_key {max_key}"
             )
 
-        min_key_idx = bisect_left(key_boundaries, min_key) - 1 if min_key else None
-        max_key_idx = bisect_right(key_boundaries, max_key) + 1 if max_key else None
+        min_key_idx = max(0, bisect_left(key_boundaries, min_key) - 1) if min_key is not None else None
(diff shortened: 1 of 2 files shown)


flliver (approved):
Good catch! Any issue w/ max key? other related edge cases?

anshumankomawar (approved):

thesalus:
> Any issue w/ max key? other related edge cases?

I had considered adding a `min()` but it seems like Python gives us the behaviour we want when the max-key index is larger than the size of the list (where it'll gracefully stop at the end of the list).

There was another edge case for interval tree construction that I opted to omit from this commit where the IntervalTree package doesn't permit null intervals. So we can't currently add (and flush) a singular row (or it'll poison the scan):

```
ValueError: IntervalTree: Null Interval objects not allowed in IntervalTree Interval(1, 1, Block(...))
```

mcember (commented):
Ship it!

PR #453, opened 2025-01-17
feat: Add Rivulet Shard Implementation

anshumankomawar:
### Shards

**Reference-Based Shards**
- **Pros:**
  - Minimal overhead when creating new shards.
  - Easy to update shards by regenerating pointers instead of creating new datasets.
  - Simplifies storage and avoids duplicating metadata/dataset attributes.
- **Cons:**
  - Limited functionality, as shards only act as pointers and do not provide direct access to dataset operations. We will need to pass shards into interfaces like the DatasetReader (refer to demo for existing syntax).

**Dataset-Based Shards**
- **Pros:**
  - Might be more intuitive for users, as shards inherit all dataset attributes and functions.
  - Users can directly operate on shards without referencing the parent dataset.
- **Cons:**
  - Higher overhead when creating shards, as they replicate dataset functionality and attributes.
  - Updating shards to reflect changes in the original dataset may require significant effort, including regenerating datasets.

**Implementation Note:**
The current implementation uses a reference-based approach to minimize overhead and simplify shard updates. However, transitioning to a dataset-based approach could improve usability and is worth considering based on future requirements.

## Changes
This pull request introduces a new example for processing and analyzing data with DeltaCat, and includes several updates to the sharding and scanning functionalities in the DeltaCat storage module. The most important changes include the addition of a new Jupyter notebook example, the implementation of an abstract `Shard` class and sharding strategies, and modifications to the dataset and block scanner classes to support sharding and field selection.

### New Example:
* `deltacat/examples/rivulet/shard_demo.ipynb`: Added a new Jupyter notebook demonstrating how to generate, load, shard, process, and aggregate data using DeltaCat, PyTorch, and Ray.

### Sharding Implementation:
* `deltacat/storage/model/shard.py`: Introduced an abstract `Shard` class and a `ShardingStrategyType` enum to define different sharding strategies. Added a `ShardingStrategy` abstract base class for creating shards.
* `deltacat/storage/rivulet/dataset.py`: Added a `shards` method to the `Dataset` class to create shards using a specified sharding strategy. Modified the `scan` method to accept an optional `shard` parameter.

### Scanning Enhancements:
* `deltacat/storage/rivulet/reader/block_scanner.py`: Updated the `BlockScanner` class to accept an optional `fields` parameter in various methods to support field selection during scans.
* `deltacat/storage/rivulet/reader/data_reader.py`: Modified the `DataReader` protocol to include an optional `fields` parameter in the `deserialize_records` and `join_deserialize_records` methods.

### Dataset Reader Updates:
* `deltacat/storage/rivulet/reader/data_scan.py`: Added `shard` and `fields` parameters to the `DataScan` class and updated the `to_arrow`, `to_pydict`, and `to_tensor` methods to use these parameters.
* `deltacat/storage/rivulet/reader/dataset_reader.py`: Updated the `DatasetReader` class to handle sharding and field selection in the `scan` method.

## Testing

`make test`

## Checklist

- [ x] Unit tests covering the changes have been added


deltacat/examples/rivulet/shard_demo.ipynb
```diff
@@ -0,0 +1,177 @@
...
+   "metadata": {},
+   "cell_type": "code",
+   "source": [
+    "parquet_file_path = pathlib.Path.cwd() / \"data.parquet\"\n",
+    "\n",
+    "# for string id try changing char_pool to string.ascii_letters\n",
+    "# this method of generation isn't consistent and you may or may not see a uniform distribution.\n",
+    "def generate_random_id(length=10, char_pool=string.digits):\n",
+    "    return ''.join(random.choices(char_pool, k=length))\n",
+    "\n",
```
flliver, line 51: This reads like a TODO? examples need to be absolute bare minimum w/o distractions and highlight the core point. Remove comments here if they aren't really a TODO.

deltacat/examples/rivulet/shard_demo.ipynb
```diff
@@ -0,0 +1,177 @@
...
+    "fake = Faker()\n",
+    "num_records = 10000\n",
+    "\n",
+    "# Try playing around with the different ID generators below.\n",
+    "# You should observe a difference in the distribution of id's over shards.\n",
+    "data = {\n",
+    "    \"id\": range(1, num_records + 1),\n",
```
flliver, line 58: nit: id's -> ids.

id's is possessive, i.e. the id's numbers are weird. In this case we're just using a regular plural, so it should be ids.

deltacat/examples/rivulet/shard_demo.ipynb
```diff
@@ -0,0 +1,177 @@
...
+    "    # \"id\": [generate_random_id(char_pool=string.ascii_letters) for _ in range(num_records)],\n",
+    "    # \"id\": [fake.uuid4() for _ in range(num_records)],\n",
+    "    \"name\": [fake.name() for _ in range(num_records)],\n",
+    "    \"age\": [random.randint(18, 80) for _ in range(num_records)],\n",
+    "    \"score\": [random.randint(0, 100) for _ in range(num_records)]\n",
+    "}\n",
```
flliver, line 64: On line 55, be more specific that line 59-60 are alternate distributions you can uncomment to try out to see how that affects shard distribution.

deltacat/storage/model/shard.py
```diff
@@ -0,0 +1,56 @@
...
+
+    TODO: Consider a shard to be a dataset in itself...
+    """
+    metastore: DatasetMetastore
+
+    def __init__(self, metastore: DatasetMetastore):
+        self.metastore = metastore
+
+    @abstractmethod
+    def __repr__(self) -> str:
```
mcember, line 24: the fact that there aren't any abstract methods defined besides repr is a red flag that maybe this is premature abstraction. 

I would prefer to just start with a dataclass with some definition of the contents of the shard. Ideally this definition would be portable. Assuming a clustered environment, is it possible to define all the manifests/SSTs in shard without sending over a full metastore? Could you just send something like "this shard includes all rows where field X has range values Y-Z"?

deltacat/examples/rivulet/shard_demo.ipynb
```diff
@@ -0,0 +1,177 @@
...
+   "metadata": {},
+   "cell_type": "code",
+   "source": [
+    "dataset = dc.Dataset.from_parquet(\n",
+    "    name=\"data\",\n",
+    "    file_uri=parquet_file_path,\n",
+    "    metadata_uri=\".\",\n",
+    "    merge_keys=\"id\"\n",
+    ")\n",
+    "print(\"Loaded dataset from Parquet file.\")\n",
```
flliver, line 84: Any reason to put the metadata outside of the dire the parquet file is in? (which is the default?)

anshumankomawar: I believe there's a slight issue with how we derive the metadata_uri
>         metadata_uri = metadata_uri or os.path.join(file_uri, "riv-meta")
The metadata_uri is correctly constructed only if file_uri is a directory and not a singular file. I'll fix this when I implement the FileSpec for rivulet, will also make the from_x methods pluggable.

deltacat/storage/model/shard.py
```diff
@@ -0,0 +1,56 @@
...
+    RANGE = "range"
+
+    def to_class(self, num_shards: int):
+        if self == ShardingStrategyType.RANGE:
+            from deltacat.storage.rivulet.shard.range_shard import RangeShardingStrategy
+            return RangeShardingStrategy(num_shards)
+        else:
+            raise ValueError(f"Unsupported sharding strategy type: {self}")
+
+class ShardingStrategy(ABC):
```
mcember, line 43: consider making this a Protocol instead of an ABC. In general prefer protocols to ABCs, and in this specific case you don't have any shared state/code which benefits from ABC. You will have to refactor shards to accept num_shards as parameter

mcember, line 43: consider making this a Protocol instead of an ABC. In general prefer protocols to ABCs, and in this specific case you don't have any shared state/code which benefits from ABC. You will have to refactor shards to accept num_shards as parameter

deltacat/storage/rivulet/reader/data_reader.py
```diff
@@ -105,6 +107,7 @@ def deserialize_records(
 
         :param records: Input data (generated by generate_records method)
         :param output_type: Type to deserialize into
+        :param fields: Only include these fields in scan result (currently only for to_tensor)
```
mcember, line 110: Fields is exposed in scan, right? So you can use this even if you aren't using the pytorch connector?

deltacat/storage/rivulet/reader/dataset_reader.py
```diff
@@ -57,16 +60,22 @@ def scan(
         # Must zipper if L0 is involved or if manifests span multiple levels
         cannot_avoid_zipper |= 0 in levels or len(levels) > 0
 
+        if shard is not None:
+            # TODO: is there a better way to filter the scan and apply different shards?
+            #       for now keeping it shard specific ideally will need to restructure/rewrite dataset_reader.
+            if isinstance(shard, RangeShard):
+                query = query.intersect_with(shard.query)
```
mcember, line 67: This is a little funky. It would be nice to keep separation between the user query and the subset of data specified by the shard

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,181 @@
...
+T = TypeVar("T", bound=Union[int, float, str])
+
+class RangeShard(Shard, Generic[T]):
+    """ Represents a range-based shard with start and end keys.
+
+    param: query: A QueryExpression object defining the range of the shard.
+    """
+    def __init__(self, metastore: DatasetMetastore, start: T, end: T):
+        super().__init__(metastore)
+        self.query = QueryExpression[T]().with_range(start, end)
```
mcember, line 17: If `query` is the way that Shard represents the data it is filtering, then this should be part of the `Shard` ABC/protocol. 

I don't think we should represent this as a Query

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,181 @@
...
+        returns: A list of RangeShard objects representing the divided range.
+        """
+        manifests = metastore.generate_manifests()
+        global_min, global_max = self.__get_max_interval(set(manifests))
+
+        if global_min is None or global_max is None:
+            return []
+
+        if global_min == global_max:
+            return [RangeShard(metastore, global_min, global_max)]
```
mcember, line 42: Discussion topic to have with broader rivulet channel: for correctness, does the data returned by each shard need to be strictly mutually exclusive? In this case, you would want to return a single shard from global min to global max, and N-1 other shards that are empty

deltacat/storage/rivulet/reader/data_reader.py
anshumankomawar: Yes, it's exposed in scan as well. Not limited to tensor, but only the pytorch connector is making use of it as of now. I can definitely update the other connectors to do something similar as well. The main reason for adding this is cause I wanted to be able to limit the pytorch connector fields to numerical values.

deltacat/storage/rivulet/reader/dataset_reader.py
anshumankomawar: Hmm would it be better to adapt the __load_sst_rows to take in multiple queries and chain them? I'm not sure what that would look like, might require refactoring query expression itself?

deltacat/storage/rivulet/shard/range_shard.py
anshumankomawar: I can see that, might be better to not mix queries and shards. Does having a defined start and end key work?

I think adding it to a shard protocol should work. Realistically if we want a uniform distribution across manifests the merge key will need to be involved and most likely it it will be in the form of a range. I'm okay with doing this and adapting if we introduce new ShardingStrategies.

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # determine range generation type
+        range_generator = RangeGeneratorType.from_key(key=global_min)
+
+        shards = []
+        local_min = global_min
+        for idx in range(self.num_shards):
```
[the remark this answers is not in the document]
anshumankomawar: Sure I can do that.

The id generator also has a random key gen method which (can) highlight this issue.

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,181 @@
...
+            # Compute the max key for the current shard
+            is_last_shard = idx == self.num_shards - 1
+            local_max = global_max if is_last_shard else range_generator.interpolate(global_min, global_max, idx + 1, self.num_shards)
+
```
flliver, line 53: What's with these extra layers of abstraction to have RangeGeneratorType? in python you use your builtins, just calculate the range, and make sure whatever class you're calculating range from has the appropriate function implemented.

Like this whole function should be three lines:
```
min = metastore.get_min_value("field")
max = metastore.get_max_value("field")
shards: List[Shard] = RangeShard.split(min, max)
return shards
```

metastore should have min/max value functions, range shard should have its own functions for common operations, and stop passing state all over god's green earth. 

I think you're issue is you're trying to do all this without modifying or adding functionality to the base classes you are working with. Instead, delete everything you have, write the absolute bare minimum human undestandable pseudo code you need to calculate what you are trying to do, which is to split an interval into a set of smaller intervals, then add the functionality to do that to the most obvious class where it belongs, without adding any additional classes. Every time you add a function, ask yourself, 'is this the responsibilty of this class?' and if it's not, move it to where it belongs.  Only when you find that there's functionality that clearly does not belong in any class do you add a new class. 

The goal should be to add the fewest number of new abstractions, fewest lines, and most easily understood functions you can to get the job done.

flliver (changes requested):
Good start, needs work! It's too complicated for what it's trying to accomplish, make better use of your base classes and remove all unnecessary abstraction.

deltacat/storage/rivulet/reader/data_reader.py
```diff
@@ -105,6 +107,7 @@ def deserialize_records(
...
         :returns: A generator yielding records of the specified type.
         """
         ...
```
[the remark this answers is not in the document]
anshumankomawar: I see what you mean. I added fields as an optional field so that users can specify what columns they want to use for the tensor generation. But they should instead be creating a schema with those fields and using that instead. I'll go ahead and remove fields it isn't correct.

deltacat/storage/rivulet/shard/range_shard.py
anshumankomawar: > I think you're issue is you're trying to do all this without modifying or adding functionality to the base classes you are working with.

I think this is why I ended up having to create these abstractions. I tried to have shard do all of the work but maybe the goal should have been to implement some of these features in the existing classes. I'll refactor it to make it simpler.

anshumankomawar: I wanted to treat RangeShard as a single unit and so didn't want to include the generation logic there. Thought it would be better to have a range generator that returns a list of ranges. I thought it could be easier to expand in the future if we wanted to support additional types. But yes having the splitting logic in the Shard itself would reduce a lot of the additional code.

deltacat/storage/rivulet/reader/pyarrow_data_reader.py
```diff
@@ -56,11 +60,21 @@ def deserialize_records(
             if row_idx == 0:
                 yield batch
 
+        elif output_type == Tensor:
+            selected_fields = fields if fields is not None else batch.column_names
+            row_values = [
+                tensor(batch.column(column_idx)[row_idx].as_py())
+                for column_idx, column in enumerate(batch.column_names)
+                if column in selected_fields
```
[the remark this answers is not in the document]
anshumankomawar: Looked into this again, looks like the existing implementation wasn't working, and so I tried to work around it with fields. I put in a fix to make sure schema fields are respected.

deltacat/storage/rivulet/reader/data_scan.py
```diff
@@ -36,21 +38,34 @@ def __init__(
...
+        return self.dataset_reader.scan(self.dataset_schema, pa.RecordBatch, self.query, self.shard, self.fields)
 
     def to_pydict(self) -> Generator[Dict, None, None]:
         """
         Generates scan results as a Dict for each row
         """
-        return self.dataset_reader.scan(self.dataset_schema, Dict, self.query)
+        return self.dataset_reader.scan(self.dataset_schema, Dict, self.query, self.shard, self.fields)
+
+    def to_tensor(self) -> Generator[
```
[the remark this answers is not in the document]
anshumankomawar: That looks great! I'll separate this pr so that we can take a look at the pytorch integration independently.

PR #455, opened 2025-01-18
Basic Transaction Snapshot Isolation Support & Storage Namespace Interface Implementation

pdames:
## Summary

Basic concurrent transaction conflict detection via MVCC-based snapshot isolation with corresponding passing unit tests. Added a main `metastore` storage interface implementation w/ namespace creation/retrieval support.

## Checklist


deltacat/tests/storage/model/test_metafile_io.py
```diff
@@ -297,10 +304,186 @@ def _commit_single_delta_table(temp_dir: str) -> List[Tuple[Metafile, Metafile,
...
+                    dest_metafile=new_delta,
+                )
+            )
+        transaction = Transaction.of(
+            txn_type=TransactionType.APPEND,
+            txn_operations=txn_operations,
+        )
+        # when the transaction is committed
+        write_paths, txn_log_path = transaction.commit(temp_dir)
+        # expect 3 new metafiles to be written
```
pdames, line 477:
```suggestion
        # expect all new deltas to be successfully written
```

deltacat/tests/storage/model/test_metafile_io.py
```diff
@@ -296,11 +304,250 @@ def _commit_single_delta_table(temp_dir: str) -> List[Tuple[Metafile, Metafile,
...
+        # expect all but one concurrent transaction to succeed each round
+        exception_count = 0
+        success_results = []
+        for result in results:
+            if isinstance(result, RuntimeError) or isinstance(result, ValueError):
+                exception_count += 1
+            else:
+                success_results.append(result)
+        assert exception_count == concurrent_commit_count * rounds - rounds
+        assert len(success_results) == rounds
```
pdames, line 363: Ideally the exceptions and successes here should be mapped directly back to each round index to more strictly assert that we got 1 success from that round specifically (e.g., the current assert would incorrectly pass 2 successes from concurrent round 1 and 0 successes from concurrent round 2).

deltacat/storage/model/metafile.py
```diff
@@ -293,115 +634,211 @@ def current(
...
 
-        mci.next_txn_id = txn_id
-        mci.next_txn_operation_type = txn_operation_type
-        metafile_name = TXN_ID_SEPARATOR.join(
-            [
-                f"{mci.next_revision:020}",
-                txn_operation_type.value,
-                f"{txn_id}{extension}",
-            ]
+        sorted_metafile_paths = MetafileRevisionInfo._sorted_file_paths(
```
pdames, line 770: I think we can improve performance here by writing a file like `_latest.mpk` in each rev directory populated with the latest metafile revision, and then putting each revision file in its own revision number directory (such that this method only needs to list the contents of the latest revision number directory to check for conflicts, and that metafile reads don't need to list revision contents to find the latest revision (they just go directly to reading `_latest.mpk`).

deltacat/storage/model/metafile.py
```diff
@@ -207,83 +402,229 @@ def operations(self) -> TransactionOperationList:
...
+        """
+        Returns the start time of the transaction in milliseconds since the
+        epoch.
+        """
+        return self.get("start_time")
+
+    def _mark_start_time(self) -> None:
+        """
+        Sets the start time of the transaction as current milliseconds since
+        the epoch. Raises a value error if the transaction start time has
```
pdames, line 416:
```suggestion
        the epoch. Raises a RuntimeError if the transaction start time has
```

deltacat/storage/model/metafile.py
```diff
@@ -22,13 +27,153 @@
...
+) -> Tuple[str, pyarrow.fs.FileSystem]:
+    """
+    Normalizes the input path and resolves a corresponding file system.
+    :param path: A file or directory path.
+    :param filesystem: File system to use for path IO.
+    :return: Normalized path and resolved file system for that path.
+    """
+    # TODO(pdames): resolve and cache filesystem at catalog root level
+    #   ensure returned paths are normalized as posix paths
+    paths, filesystem = _resolve_paths_and_filesystem(
```
mcember, line 48: nit: we shouldn't be using internal utilities from other libraries

deltacat/storage/model/metafile.py
```diff
@@ -22,13 +27,153 @@
...
+            # so only quote single file path here.
+            paths = f'"{paths}"'
+        raise OSError(
+            (
+                f"Failing to read AWS S3 file(s): {paths}. "
+                "Please check that file exists and has properly configured access. "
+                "You can also run AWS CLI command to get more detailed error message "
+                "(e.g., aws s3 ls <file-name>). "
+                "See https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/index.html "  # noqa
+                "and https://docs.ray.io/en/latest/data/creating-datasets.html#reading-from-remote-storage "  # noqa
```
mcember, line 86: link is broken, should be https://docs.ray.io/en/latest/data/loading-data.html#reading-files-from-cloud-storage

Why are we returning ray data documentation? unecessary coupling

deltacat/storage/model/metafile.py
pdames: +1 - Have a TODO to pull this internal and modify as required.

deltacat/storage/model/metafile.py
pdames: Method was pulled from Ray Data - will remove link (and also propose a quick fix regarding the broken link back to Ray Data). We should ideally change this error message to point to relevant AWS docs, and our related internal docs when ready.

deltacat/storage/model/metafile.py
```diff
@@ -22,13 +27,153 @@
...
+
+    return file_infos
+
+
+def _get_file_info(
+    path: str,
+    filesystem: pyarrow.fs.FileSystem,
+    ignore_missing_path: bool = False,
+) -> Union[FileInfo, List[FileInfo]]:
+    """Get the file info or list of file infos for the provided path."""
```
thesalus, line 166: Would this ever return a list of file infos (if only a single `path` str is provided)?

deltacat/storage/model/metafile.py
```diff
@@ -174,16 +318,67 @@ def of(
...
+    ) -> Transaction:
+        """
+        Read a Transaction file and return the deserialized object.
+        :param path: Transaction file path to read.
+        :param filesystem: File system to use for reading the Transaction file.
+        :return: Deserialized object from the Transaction file.
+        """
+        if not filesystem:
+            path, fs = _filesystem(path, filesystem)
+        with fs.open_input_stream(path) as file:
```
thesalus, line 367: Should `_filesystem` reassign back to `path, filesystem`?

```suggestion
        if not filesystem:
            path, filesystem = _filesystem(path, filesystem)
        with filesystem.open_input_stream(path) as file:
```

deltacat/storage/model/metafile.py
```diff
@@ -207,83 +402,229 @@ def operations(self) -> TransactionOperationList:
...
+        except ValueError:
+            raise OSError(
+                f"Transaction log file `{txn_log_file_path}` does not "
+                f"contain a valid UUID string."
+            )
+        end_time = Transaction.end_time(
+            path=txn_log_file_path,
+            filesystem=filesystem,
+        )
+        if end_time - start_time < 0:
```
thesalus, line 490: Nitpick: (for my own edification) why not compare the values directly?

```suggestion
        if end_time < start_time:
```

deltacat/storage/model/metafile.py
```diff
@@ -207,83 +402,229 @@ def operations(self) -> TransactionOperationList:
...
+            finally:
+                raise OSError(
+                    f"Transaction end time {end_time} is earlier than start "
+                    f"time {start_time}! This may indicate a problem "
+                    f"with either the system clock on the host serving the "
+                    f"transaction, or a loss of precision in the filesystem "
+                    f"recording the completed transaction file timestamp (at "
+                    f"least millisecond precision is required). To preserve "
+                    f"catalog integrity, the corresponding completed "
+                    f"transaction log at `{txn_log_file_path}` has been removed."
```
thesalus, line 506: It looks like we *only* raise an Exception if the other conditions are violated. Why don't we remove the transaction logs in those cases as well?

Is this the only irrecoverable scenario?

thesalus (commented):
Leaving some early comments (partway through metafile.py)

deltacat/tests/storage/model/test_metafile_io.py
```diff
@@ -303,7 +304,70 @@ def _commit_single_delta_table(temp_dir: str) -> List[Tuple[Metafile, Metafile,
...
+                table.locator.table_name = f"{base_table_name}_{round}"
+                futures = [
+                    pool.apply_async(
+                        _commit_concurrent_transaction, (temp_dir, transaction)
+                    )
+                    for _ in range(concurrent_commit_count)
+                ]
+                results.extend([future.get() for future in futures])
+
+        # expect all but one concurrent transaction to succeed each round
```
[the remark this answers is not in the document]
pdames: Yeah - I wrote a related comment to change this check to ensure that exactly one succeeds each round, instead of ensuring that the sum equals the number of rounds (which could produce false positives).

deltacat/storage/model/metafile.py
```diff
@@ -22,13 +27,153 @@
...
+            excluded from the returned file set. Default excluded prefixes are
+            "." and "_".
+
+    Returns:
+        An iterator of (file_path, file_size) tuples.
+    """
+    if exclude_prefixes is None:
+        exclude_prefixes = [".", "_"]
+
+    from pyarrow.fs import FileSelector
```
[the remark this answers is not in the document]
pdames: This is typically done in python projects where you don't want your package to require users to pull in an optional dependency (together with listing associated typing for that dependency inside of strings like `fs: "pyarrow.fs.FileSelector"` in this case) but, in this case, it's not needed since DeltaCAT always requires pyarrow (although it's worth noting that if we ever wanted to make pyarrow optional, we'd have to change A LOT of code to only pull in the optional pyarrow dependency on demand).

deltacat/storage/model/metafile.py
pdames: Good catch - this was left behind during refactoring out of `_get_file_infos()`, and _shouldn't_ ever return a list now. 
```suggestion
    """Get the file info for the provided path."""
```

deltacat/storage/model/metafile.py
pdames: In this case, I use the reassignment to try to help prevent errors where the original unresolved/not-normalized filesystem and path objects are used again later in the method. Basically, once they've been resolved, we need to ensure that we _always_ use the normalized path and filesystem returned from this method. Open for other suggestions about how to more cleanly ensure this though.

deltacat/storage/model/metafile.py
pdames: No good reason! Your suggestion is cleaner - thanks!

deltacat/storage/model/metafile.py
pdames: We should - good catch! Any of the above failed checks risk breaking subsequent Transaction Snapshots that attempts to operate on revisions of files written by the transaction with corrupt properties. 

One other thing I've been thinking is that we should ideally add the same checks in-line during other transactions that are getting info about the transaction that committed a revision, since right now a race condition could mean that another transaction is already reading the corrupt log before we clean it up (but that race condition should imply that they will also fail the same checks we have here, and could thus make their own decision to ignore that transaction until it's cleaned up).

deltacat/storage/model/metafile.py
mcember: Ok sounds good. This sort of function also can be moved to utils

deltacat/storage/model/metafile.py
```diff
@@ -22,13 +27,153 @@
...
+                "See https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/index.html "  # noqa
+                "and https://docs.ray.io/en/latest/data/creating-datasets.html#reading-from-remote-storage "  # noqa
+                "for more information."
+            )
+        )
+    else:
+        raise error
+
+
+def _expand_directory(
```
mcember, line 94: can we move this to utils? perhaps filesystem utils

pdames: Yeah - was planning on at least (1) factoring these filesystem methods out into utils, (2) also moving Transaction classes out into separate model files.

deltacat/storage/model/metafile.py
thesalus: Sorry, I was a bit unclear. The comment was moreso about a potential bug from the difference in variable names.

`fs` will be None if `filesystem` *is* present (unless we overwrite it or instantiate it separately).

deltacat/storage/model/metafile.py
```diff
@@ -293,115 +634,211 @@ def current(
...
+                # the current transaction can only build on top of the snapshot
+                # of commits from transactions that completed before it started
+                txn_end_time = Transaction.end_time(
+                    path=posixpath.join(txn_log_dir, mri.txn_id),
                     filesystem=filesystem,
-                    ignore_missing_path=True,
                 )
-                if file_paths_and_sizes:
-                    latest_committed_metafile_path = latest_metafile_path
+                if txn_end_time is not None and txn_end_time < current_txn_start_time:
```
thesalus, line 659: Minor: will this fail if `current_txn_start_time is None`? Should the containing `else` instead be `else if current_txn_start_time is not None`?

deltacat/storage/model/metafile.py
pdames: Oh, yes - your suggested change is exactly what it should be! Thanks!

deltacat/storage/model/metafile.py
pdames: Yep - good catch! This would be a problem if looking for the latest revision outside the context of a transaction.

deltacat/catalog/main/impl.py
```diff
@@ -0,0 +1,213 @@
...
+    DistributedDataset,
+    LifecycleState,
+    LocalDataset,
+    LocalTable,
+)
+from deltacat.types.media import ContentType
+from deltacat.types.tables import TableWriteMode
+
+# TODO(pdames): Create internal DeltaCAT port
+from ray.data.datasource.path_util import _resolve_paths_and_filesystem
```
pdames, line 22: use `deltacat.utils.filesystem` version instead

deltacat/catalog/main/impl.py
```diff
@@ -0,0 +1,213 @@
...
+        return self._root
+
+    @property
+    def filesystem(self) -> Optional[pyarrow.fs.FileSystem]:
+        return self._filesystem
+
+
+# table functions
+def write_to_table(
+    data: Union[LocalTable, LocalDataset, DistributedDataset],
```
mcember, line 48: We're going to have to reconcile dataset.py and some functions here. I'm wondering if it makes sense to spin out a separate catalog impl just for table [version?]. Worth taking a look at methods in rivulet [dataset.py](https://github.com/ray-project/deltacat/blob/2.0/deltacat/storage/rivulet/dataset.py)  at methods like `from_parquet`, `export`, `add_schema`, and thinking where they fit in to deltacat

deltacat/catalog/main/impl.py
```diff
@@ -0,0 +1,213 @@
...
+    When creating a table, all `create_table` parameters may be optionally
+    specified as additional keyword arguments. When appending to, or replacing,
+    an existing table, all `alter_table` parameters may be optionally specified
+    as additional keyword arguments."""
+    raise NotImplementedError("write_to_table not implemented")
+
+
+def read_table(
+    table: str, namespace: Optional[str] = None, *args, **kwargs
+) -> DistributedDataset:
```
mcember, line 68: How does the user signal if they want to read as a ray dataset or daft dataframe? 

We should compare this with the [rivulet dataset scan](https://github.com/ray-project/deltacat/blob/2.0/deltacat/storage/rivulet/dataset.py#L519) and think about how to integrate that into this interface

deltacat/catalog/main/impl.py
```diff
@@ -0,0 +1,213 @@
...
+    """Read a table into a distributed dataset."""
+    raise NotImplementedError("read_table not implemented")
+
+
+def alter_table(
```
mcember, line 73: Can you explain how table versions and streams are layered in here? My understanding was that a user can explicitly manage multiple versions of a table, and that within a table version they can explicitly manage multiple different streams (representing "variants" of the same data).

deltacat/storage/main/impl.py
```diff
@@ -0,0 +1,733 @@
...
+
+def list_namespaces(*args, **kwargs) -> ListResult[Namespace]:
+    """
+    Lists a page of table namespaces. Namespaces are returned as list result
+    items.
+    """
+    return _list_metafiles(
+        *args,
+        metafile=Namespace.of(NamespaceLocator.of("placeholder")),
+        txn_op_type=TransactionOperationType.READ_SIBLINGS,
```
mcember, line 130: why is READ_SIBLINGS a transaction type? 

A transaction type like READ_LATEST makes sense to me, because you are essentially locking to a transaction start time and reading with respect to that snapshot. Does READ_SIBLING just list sibling directories, and in that case does it even have to do with transactions?

This may be a smell that the scope of transaction management / commits is too large, if you need to go through transaction management to basically do any metastore reads

deltacat/storage/main/impl.py
```diff
@@ -0,0 +1,733 @@
...
+    """
+    raise NotImplementedError("list_stream_partitions not implemented")
+
+
+def list_deltas(
+    namespace: str,
+    table_name: str,
+    partition_values: Optional[PartitionValues] = None,
+    table_version: Optional[str] = None,
+    first_stream_position: Optional[int] = None,
```
mcember, line 182: Are stream positions comparable across partitions? Or is this only relevant within a single partition?

deltacat/storage/main/impl.py
```diff
@@ -0,0 +1,733 @@
...
+    last_stream_position: Optional[int] = None,
+    ascending_order: Optional[bool] = None,
+    include_manifest: bool = False,
+    partition_scheme_id: Optional[str] = None,
+    *args,
+    **kwargs,
+) -> ListResult[Delta]:
```
mcember, line 189: So having paginated listing everywhere means we need a custom implementation of pyarrow filesystems? We should bring this up with Anshuman who is refactor filesystem interfaces

deltacat/storage/main/impl.py
```diff
@@ -0,0 +1,733 @@
...
+    given table version or partition does not exist.
+
+    To conserve memory, the delta returned does not include a manifest by
+    default. The manifest can either be optionally retrieved as part of this
+    call or lazily loaded via a subsequent call to `get_delta_manifest`.
+    """
+    raise NotImplementedError("get_latest_delta not implemented")
+
+
+def download_delta(
```
mcember, line 277: Does this belong in the storage interface? Playing devil's advocate - the storage interface should be just a dumb wrapper for manipulating the metastore. Once you get into actually using metadata files to download data, that feels more like the catalog interface or some level of abstraction between catalog and storage

deltacat/catalog/main/impl.py
```diff
@@ -0,0 +1,213 @@
...
+    *args,
+    **kwargs
+) -> None:
+    """Alter table definition."""
+    raise NotImplementedError("alter_table not implemented")
+
+
+def create_table(
+    table: str,
+    namespace: Optional[str] = None,
```
mcember, line 91: How does creating a table work when namespace isn't specified? Does it create a default namespace?

deltacat/storage/main/impl.py
```diff
@@ -0,0 +1,733 @@
...
+    """
+    Gets the most recently committed partition for the given stream locator and
+    partition key values. Returns None if no partition has been committed for
+    the given table version and/or partition key values. Partition values
+    should not be specified for unpartitioned tables.
+    """
+    raise NotImplementedError("get_partition not implemented")
+
+
+def stage_delta(
```
mcember, line 567: I've been meaning to ask: where in the deltacat metastore are data files stored? Presumably staging a delta will choose where to write data files?

mcember, line 567: Given the large number of arguments, should we do what you've done in other parts of deltacat and make most parameters keyword only?

pdames: This would break backwards compatibility with 1.X so, while I think this is probably the right approach eventually, I'd prefer to do it in a subsequent iteration after we've merged 2.0 back to main.

pdames: Right now the data file storage location is auto-resolved by the catalog. This may be an area where we need some additional arguments if we'd like to support staging deltas to multiple disparate locations with rivulet though.

deltacat/catalog/main/impl.py
pdames: Yes - tables land in a global system namespace if the namespace isn't specified, so the namespace effectively becomes implicit.

deltacat/storage/main/impl.py
pdames: For compactors and similar system maintenance jobs, the answer is "yes", since they're arguably one of few job types that need to be delta aware. The assumption here is that the, average end-user shouldn't need to understand the CDC log of deltas that are ultimately merged into the final table/dataset they read/write, and should instead just read/write at the table/dataset abstraction level instead of at the delta level.

deltacat/storage/main/impl.py
pdames: Perhaps, but it may not be the only solution to the problem (e.g., we could also entertain creating a "delta page" metafile that stores N delta locators inside of it).

Either way, I definitely agree that the default pyarrow filesystems will start becoming vastly inefficient (esp. for S3 and other cloud storage systems) with the current metastore layout as the number of objects starts growing into the thousands and beyond, since pyarrow fs will attempt to eagerly list all all child partitions of a table stream, all child deltas of a stream partition, etc. Even if we can paginate the loading of the full Table/Partition/Delta metafile contents, just the exhaustive listing itself here is undesirable.

deltacat/storage/main/impl.py
pdames: Stream positions are not currently assumed to be comparable across partitions.

deltacat/storage/main/impl.py
pdames: You don't strictly need to go through a transaction manager for any metastore reads, but the point of MVCC Snapshot Isolation is to use read transactions whenever you want to "freeze" the database in a consistent state during a read, such that you don't interleave partial state from multiple in-progress transactions into your read.

So reads outside the context of a transaction are perfectly valid, but you'll see the evolving "streaming" state of the metastore instead of the "frozen" state of the metastore captured at the beginning of the transaction, and associated ACID compliance guarantees will be lost.

However, I think it could be argued that READ_SIBLINGS in particular could be considered potentially unnecessary when you already have READ_CHILDREN (e.g., we could treat all namespaces as children of the root catalog, which presumably has no other catalog-level siblings).

deltacat/catalog/main/impl.py
pdames: This is just an old/out-of-date interface. The dataset type they want to read the table as needs to be exposed as an explicit argument here.

deltacat/catalog/main/impl.py
pdames: The current thought is that there is only a fixed number of streams per table version, where each stream is uniquely identified by the metastore type it's associated with (either Iceberg, Hudi, Delta Lake, or the main DeltaCAT straem). Non-DeltaCAT native stream types will have metadata writes to the DeltaCAT main stream translated and synchronized to them, either directly (e.g., Iceberg) or transitively (e.g., from Iceberg to Hudi/DeltaLake via XTable).

The TableVersion then holds properties that are intended to be common to all streams, like the latest Arrow Schema  associated with the Stream that the Iceberg/Hudi/Delta schema has ultimately been derived from (with no derivation needed for the main DeltaCAT stream, since it uses Arrow Schemas natively).

I think TableVersion creation will largely be a backend decision (e.g., by default, we should put a backwards-incompatible schema change into a new table version), but could also be an optional, explicit decision from the end-user (e.g., the user effectively wants to create a new evolutionary "branch" of their table, and hold that branch in a new table version). 

By Default, table reads will resolve to the latest active table version, but a user can specify if they'd like to explicitly read a different table version.

deltacat/catalog/main/impl.py
pdames: We could always propose a separate catalog interface just to see what it looks like. Today, in DeltaCAT 1.X, only the storage interface is used extensively, so we're pretty free to evolve or propose a new catalog interface in 2.0.

deltacat/storage/main/impl.py
pdames: So, in short, I don't disagree with the purity of keeping the storage interface restricted metadata manipulation only, but we would need to find a way to bridge the gap somehow between exposing system-level API parameters for delta-level read/write to backend procedures like compaction w/o also unnecessarily exposing the internals of the metastore format to end users.

deltacat/storage/main/impl.py
pdames: In reading your comment again, I should also be clear that `READ_SIBLINGS` reads the latest revision of all siblings for the current transaction's snapshot (i.e., this will read the latest committed version of every namespace in the catalog for the current transaction snapshot, and not just return whatever namespace revisions it happens to find on disk).

deltacat/storage/main/impl.py
mcember: Ok that makes sense, I'm aligned with this after reading your comment

PR #456, opened 2025-01-19
fix: respect schema fields in file reader

anshumankomawar:
## Summary

Found an issue with schema fields not being respected during dataset scan. If new schemas are created and explicitly mentioned in a DataScan the returned fields should be limited to that of the schema.

## Changes

Updated the `FileReader implementations to select columns during batch read. This functionality is already provided by pyarrow through methods like batch.select which have an optional column field.

## Testing

- `make test` to run existing tests. 
- verified changes with manual testing of DataScan ensured outputs are as expected
- added a end to end test for scan with multiple schemas to highlight the fix

## Checklist

### DIFF
--- a/deltacat/storage/rivulet/feather/file_reader.py
+++ b/deltacat/storage/rivulet/feather/file_reader.py
@@ -13,6 +13,7 @@ from deltacat.storage.rivulet.reader.data_reader import (
     FILE_FORMAT,
 )
 from deltacat.storage.rivulet.reader.pyarrow_data_reader import RecordBatchRowIndex
+from deltacat.storage.rivulet.schema.schema import Schema
 
 
 class FeatherFileReader(FileReader[RecordBatchRowIndex]):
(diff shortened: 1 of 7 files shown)


deltacat/storage/rivulet/feather/file_reader.py
```diff
@@ -23,13 +24,15 @@ class FeatherFileReader(FileReader[RecordBatchRowIndex]):
     TODO can consider abstracting code between this and ParquetFileReader
     """
 
-    def __init__(self, sst_row: SSTableRow, file_store: FileStore, primary_key: str):
+    def __init__(self, sst_row: SSTableRow, file_store: FileStore, primary_key: str, schema: Schema):
         self.sst_row = sst_row
         self.input = file_store.new_input_file(self.sst_row.uri)
 
         self.key = primary_key
```
flliver, line 31: I can't decide if I like forcing every reader to explicitly provide a schema, but can't think of an alternative and we would certainly get there with push down filters anyways, so LGTM at the moment!

flliver (approved):

deltacat/storage/rivulet/feather/file_reader.py
anshumankomawar: Sounds good! Push down filters will should definitely be the long term solution.

PR #457, opened 2025-01-21
fix: clear data and sst files after manifest flush

anshumankomawar:
## Summary
Manifests currently include data from each write and do not remove data from previous writes. The issue is in the dataset writer flush method. Clearing sst_files and data_files after each flush ensures the manifest contains data from only that flush.

## Testing

- `make test` for existing tests. 
- added a test for `MemtableDatasetWriter` to highlight this bug.

## Checklist


flliver (approved):
Good catch, ship it! Longer term we'll need to think about what locking/idempotency looks like around the flush method, but we can save that for the integrate Matt's doing w/ the DC manifest committer.

PR #459, opened 2025-01-21
Port benchmarks from rivulet

mcember:
## Summary

Port benchmarks. Modify imports and make refactors for changes to schema and other interfaces when we ported internal repo to deltacat

NOTE: currently this diff is pulling in changes from linter, per https://github.com/ray-project/deltacat/pull/458. We should merge in 458 first then rebase this.


deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,7 @@
+All the dependences needed to test are in dev-requirements.txt
+
+You can run these tests via:
+
+```shell
+pytest -m benchmark deltacat/benchmarking
```
pdames, line 6: Should we consolidate the instructions required to run these benchmarks inside of make to avoid the risk of changing the command to run them in one place but not the other?

```suggestion
make benchmark
```

deltacat/benchmarking/benchmark_engine.py
```diff
@@ -0,0 +1,82 @@
...
+    """Convenience for computing elapsed time of a block of code as a metric.
+
+    :param description: description of the step
+    :return: a benchmark operation populated with the elapsed time
+    """
+    metric = BenchmarkStep(description)
+    start_time = time.time()
+    yield metric
+    end_time = time.time()
+    metric.add(BenchmarkMetric("elapsed_time", 1000 * (end_time - start_time), "ms"))
```
pdames, line 22: Is there an opportunity to replace this with the existing `invoke_with_perf_counter` https://github.com/ray-project/deltacat/blob/2.0/deltacat/utils/performance.py#L8-L17?

deltacat/benchmarking/test_benchmark_pipeline.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@contextmanager
+def make_tmpdir():
+    tmpdir = tempfile.mkdtemp()
+    try:
+        yield tmpdir
+    finally:
+        shutil.rmtree(tmpdir)
+        pass
```
pdames, line 39: Nit: This is also duplicated at: https://github.com/ray-project/deltacat/blob/2.0/deltacat/tests/storage/conftest.py#L9-L16. Can we make this a fixture, and share it between benchmarks and unit tests from one common location?

deltacat/benchmarking/benchmark_engine.py
mcember: In general I would like to refactor the whole benchmark framework. I would like to be able to write the tests more like normal client code without wrapping things in BenchmarkEngine/Run/Step objects.

Having said that this CR is more or less a copy-pasta, I don't want to take on scope for refactoring. `invoke_with_perf_counter` can't be just used as-is because it runs the passed in func whereas benchmark metric is more like a dataclass which gets wrapped in benchmark step

deltacat/benchmarking/test_benchmark_pipeline.py
mcember: My understanding talking to an LLM is that it is an anti pattern to import fixtures explicitly, and multiple conftest.py files can only be automatically imported hierarchically, not from sibling directories. So unless we moved /benchmarks under /tests, we can't share fixtures

As a compromise I moved this to test utils and imported that from the benchmarks and test fixture

deltacat/benchmarking/README.md
mcember: fixed thanks

pdames (approved):

deltacat/storage/rivulet/dataset.py
```diff
@@ -28,6 +28,7 @@
 
 # These are the hardcoded default schema names
 ALL = "all"
+ALL = "all"
```
thesalus, line 31: Nitpick: looks like this got duplicated.

deltacat/benchmarking/README.md
```diff
@@ -0,0 +1,7 @@
+All the dependences needed to test are in dev-requirements.txt
```
thesalus, line 1: Typo: This is going to bug me otherwise.

```suggestion
All the dependencies needed to test are in dev-requirements.txt
```

thesalus (approved):

thesalus (approved):

PR #460, opened 2025-01-21
feat: add range shard implementation

anshumankomawar:
# Shards

## Reference-Based Shards

**Pros**  
- Minimal overhead when creating new shards  
- Easy to update shards by regenerating pointers instead of creating new datasets  
- Simplifies storage and avoids duplicating metadata/dataset attributes  

**Cons**  
- Limited functionality: shards only act as pointers and do not provide direct access to dataset operations  
- Requires passing shards into interfaces like `DatasetReader` (refer to demo for existing syntax)  

## Dataset-Based Shards

**Pros**  
- Potentially more intuitive: shards inherit all dataset attributes and functions  
- Users can directly operate on shards without referencing the parent dataset  

**Cons**  
- Higher overhead when creating shards, as they replicate dataset functionality and attributes  
- Updating shards to reflect changes in the original dataset may require significant regeneration of datasets  

**Implementation Note:**  
The current implementation uses a reference-based approach to minimize overhead and simplify shard updates. Transitioning to a dataset-based approach could improve usability and is worth considering based on future requirements.

---

## Changes

This pull request adds a new example for processing and analyzing data with DeltaCat, and includes several updates to the sharding and scanning functionalities in the DeltaCat storage module.

### Sharding Implementation

- **`deltacat/storage/model/shard.py`**  
  - Introduced an abstract `Shard` protocol. 
  - Added a `ShardingStrategy` abstract base class for creating shards.
- **`deltacat/storage/rivulet/dataset.py`**  
  Added a `shards` method to the `Dataset` class to create shards using a specified sharding strategy. Modified the `scan` method to accept an optional `shard` parameter.

### Dataset Reader Updates

- **`deltacat/storage/rivulet/reader/data_scan.py`**  
  Added `shard` parameters to the `DataScan` class and updated the `to_arrow`, `to_pydict` methods accordingly.
- **`deltacat/storage/rivulet/reader/dataset_reader.py`**  
  Updated the `DatasetReader` class to construct a merged query using the shard and user query. Ensures we filter before read.

---

## Testing

```bash
make test
```

## Checklist


deltacat/storage/rivulet/reader/data_scan.py
```diff
@@ -36,21 +37,28 @@ def __init__(
...
+        return self.dataset_reader.scan(
+            self.dataset_schema, pa.RecordBatch, self.query, shard=self.shard
+        )
 
     def to_pydict(self) -> Generator[Dict, None, None]:
         """
         Generates scan results as a Dict for each row
         """
-        return self.dataset_reader.scan(self.dataset_schema, Dict, self.query)
+        print(self.shard)
```
anshumankomawar, line 61: Missed this print statement will remove it.

deltacat/storage/rivulet/reader/query_expression.py
```diff
@@ -44,6 +46,29 @@ def with_range(self, bound1: T, bound2: T) -> "QueryExpression":
         self.key_range = tuple(sorted([bound1, bound2]))
         return self
 
+    @staticmethod
+    def with_shard(query: Optional["QueryExpression"], shard: Shard):
```
mcember, line 50: you can avoid quoting types hints referencing self if you add: `from __future__ import annotations`

deltacat/storage/rivulet/reader/query_expression.py
```diff
@@ -44,6 +46,29 @@ def with_range(self, bound1: T, bound2: T) -> "QueryExpression":
...
+        Generate a query expression that accounts for the shard boundaries.
+        Shard boundaries are inclusive and mark the outer bounds of the query.
+        """
+        if shard is None:
+            return query
+
+        if query.key_range is None:
+            return QueryExpression().with_range(shard.min_key, shard.max_key)
+
+        min_key = shard.min_key
```
mcember, line 61: consider simplifying lines 61 -70 like:

`return QueryExpression().with_range(min(shard.min_key, query.key_range.min_key), max(shard.max_key, query.key_range.max_key)`

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,101 @@
+from __future__ import annotations
+from typing import Generic, List, TypeVar, Union, Iterable
+from deltacat.storage.model.shard import Shard, ShardingStrategy
+from deltacat.storage.rivulet.reader.dataset_metastore import DatasetMetastore
+
+
+T = TypeVar("T", bound=Union[int, float, str])
```
mcember, line 7: this typevar is important because it dictates what data types primary keys must adhere to in order to be shard-able

Can you add some validation code in `dataset`/`schema` to ensure this is true, or add a TODO for it? We should also declare this type var once and share it via imports rather than declare it in multiple places

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,101 @@
...
+
+        def interpolate_numeric(
+            start: int, end: int, step: int, total_steps: int
+        ) -> int:
+            """
+            Integer interpolation using integer (floor) division.
+            """
+            return start + (end - start) * step // total_steps
+
+        def interpolate_str(start: str, end: str, step: int, total_steps: int) -> str:
```
mcember, line 49: Can you update this to interpolate in a way that respects lexicographical ordering? That way we have some nice guarantees like data from different shards will always be easily sortable 

see: https://chatgpt.com/share/6793e5db-c600-8012-a88e-a3b9de3baafc

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,101 @@
...
+            max_len = max(len(start), len(end))
+            start_bytes = start.ljust(max_len, "\0").encode("utf-8")
+            end_bytes = end.ljust(max_len, "\0").encode("utf-8")
+            interpolated = bytes(
+                s + (e - s) * step // total_steps
+                for s, e in zip(start_bytes, end_bytes)
+            )
+            return interpolated.decode("utf-8").strip("\0")
+
+        # Determine which interpolation function to use based on the type of min/max
```
mcember, line 63: looks like we don't support float, so you should remove that from the type union

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,101 @@
...
+
+        :param global_min: The minimum key for the entire range (int or str).
+        :param global_max: The maximum key for the entire range (int or str).
+        :param num_shards: The number of shards to create.
+        :return: A list of RangeShard objects.
+        """
+        if global_min == global_max or num_shards <= 1:
+            return [RangeShard(global_min, global_max)]
+
+        def interpolate_numeric(
```
mcember, line 41: So these are nested functions? That feels a little excessive...I would just make them private functions (i.e. `_interpolate_numeric`), that way you could unit test them independently too

deltacat/storage/rivulet/shard/range_shard.py
```diff
@@ -0,0 +1,101 @@
...
+            raise ValueError(
+                "Unsupported combination of types for global_min and global_max."
+            )
+
+        shards: List[RangeShard] = []
+        for i in range(num_shards):
+            start = interpolate(global_min, global_max, i, num_shards)
+            end = interpolate(global_min, global_max, i + 1, num_shards)
+            if i > 0 and isinstance(global_min, int):
+                start += 1
```
mcember, line 78: What about if we're using string values? Don't you need similar logic to ensure that sequential ranges don't overlap?

deltacat/tests/storage/model/test_shard.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+from deltacat.storage.model.shard import ShardingStrategy
+
+
+def test_sharding_strategy_from_string_range():
```
mcember, line 6: general unit testing comment: these tests are fine and good, but I feel that we shouldn't feel the need to test super simple code. If the shard strategy selection of "range" is broken, other higher level tests should fail.

deltacat/tests/storage/rivulet/shard/test_range_shard.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def test_range_shard_split_integers_single_shard():
+    shards = RangeShard.split(global_min=1, global_max=10, num_shards=1)
+    assert len(shards) == 1
+    assert shards[0].min_key == 1
+    assert shards[0].max_key == 10
+
+
+def test_range_shard_split_integers_same_value():
```
mcember, line 80: hmmm if this is actually desired behavior, can we document in `dataset.shards` that we will return the num_shards or fewer, and in what instances we return fewer than user provided num_shards? 

If this is about super small key ranges then we probably will never see it in practice

mcember (approved):
Approved pending you address minor comments. Doesn't need re-review

deltacat/storage/rivulet/reader/query_expression.py
anshumankomawar: ah didn't know about that, much cleaner.

deltacat/storage/rivulet/shard/range_shard.py
anshumankomawar: That's fair I'll move it to the shard model, and add some TODO to add validation.

deltacat/tests/storage/rivulet/shard/test_range_shard.py
anshumankomawar: It's more so targeted towards smaller key ranges I doubt we would see this in practice but thought we could include a test incase anyone does encounter it. I'll document this behavior in dataset.shards.

deltacat/storage/rivulet/shard/range_shard.py
anshumankomawar: You're right I missed that will add something to handle that case.

deltacat/storage/rivulet/reader/query_expression.py
anshumankomawar: This might not work, I was aiming to treat a shard boundary as the outer bounds. i.e. If query min was lower than shard min I would still want to use the shard min.

deltacat/storage/rivulet/shard/range_shard.py
anshumankomawar: That would be nice to count on. I'll update this to a lexicographically ordered sort

deltacat/tests/storage/model/test_shard.py
anshumankomawar: sounds good will keep that in mind in the future.

PR #462, opened 2025-01-23
Allow overriding content encoding

raghumdani:
## Summary

This change adds a new kwarg that can be used to override the content encoding in the metadata. This applies to every parquet file. The reason we are adding this is to gradually dial up all parquet files to read via identity encoding. 

## Rationale

The internal catalog metadata incorrectly sets content type of parquet files as gzip even though full file isn't gzip compressed.

## Changes

Add a new kwarg which can be used to override the content encoding for parquet

## Impact

Unblocks compaction of new tables.

## Testing

Functional tests

## Regression Risk

Very low as it doesn't affect the current codepath unless the kwarg is set. 

## Checklist

## Additional Notes


yankevn (approved):
LGTM

PR #464, opened 2025-01-24
Refactor rivulet manifest files to use deltacat 

mcember:
## Summary
See milestone https://github.com/ray-project/deltacat/milestone/5

* Remove json manifest format in favor of deltacat delta (which wraps manifest)
* Begin refactoring some internal classes

Note that this PR uses a metastore format like
 $root/ 
   $metadata/
      <SST files>
   $data/ 
      <data files>
   <delta_id_n>/
      /rev
       <delta revisions>
   <txn>

This is NOT the final structure for using the deltacat metastore. In part this is waiting on the storage interface from https://github.com/ray-project/deltacat/pull/455/

Discussion points:
1. How is a deltcat writer expected to assign stream position
2. does deltacat have opinions about where data files are written?

## Testing

rivulet unit tests


deltacat/storage/rivulet/fs/fs_utils.py
```diff
@@ -0,0 +1,150 @@
+import re
```
mcember, line 1: NOTE all of the methods are copy-pasta from https://github.com/ray-project/deltacat/pull/455/

Makefile
```diff
@@ -35,6 +35,9 @@ lint: install
 test: install
 	venv/bin/pytest -m "not integration"
 
+unit-test: install
```
mcember, line 38: I just added this to run tests more rapidly. Note that some of the unit tests for compaction take a while so we should consider moving them to benchmarks

deltacat/storage/model/manifest.py
```diff
@@ -135,7 +135,7 @@ def of(
                         f"'{entry_content_type}'"
                     )
                     raise ValueError(msg)
-                entry_content_encoding = meta["content_encoding"]
+                entry_content_encoding = meta.get("content_encoding", None)
```
mcember, line 138: This was raising a ValueError for manifests with no content encoding (like in Rivulet...). We don't populate lots of fields currently for rivulet deltas/manifests so we will have to update code to populate them (see: https://github.com/ray-project/deltacat/issues/476)

deltacat/storage/rivulet/metastore/delta.py
```diff
@@ -0,0 +1,184 @@
...
+        delta_locator = DeltaLocator.at(
+            namespace=None,
+            table_name=None,
+            table_version=None,
+            stream_id=None,
+            stream_format=None,
+            partition_values=None,
+            partition_id=None,
+            # Using microsecond precision timestamp as stream position
+            # TODO discuss how to assign stream position
```
mcember, line 148: We had initial discussion that timestamp was safe to use as stream position. Keeping this TODO in code base since we still need to decide e.g. if we should use ms or nanosecond stream position, whether storage interface should default it if not provided

deltacat/storage/rivulet/reader/dataset_metastore.py
```diff
@@ -37,29 +43,67 @@ def generate_sstables(self) -> Generator[SSTable, None, None]:
...
+    def _get_delta(self, delta_dir: str, filesystem: FileSystem) -> RivuletDelta:
+        """
+        Given a DeltaCat delta directory, find latest delta file
+
+        This will be replaced by deltacat storage API.
+        Current implementation does not respect open/closed transactions, it just
+            looks for the latest revision
+        """
+        rev_directory = os.path.join(delta_dir, "rev")
+        revisions = filesystem.get_file_info(fs.FileSelector(rev_directory))
```
anshumankomawar, line 77: nit: should we use the fs utils method here? I believe it adds some error handling around get_file_info.

I can include it as part of my changes as well.

deltacat/storage/rivulet/reader/dataset_metastore.py
```diff
@@ -37,29 +43,67 @@ def generate_sstables(self) -> Generator[SSTable, None, None]:
...
+        # Rivulet data and SST files written to /data and /metadata
+        # Deltacat transactions written to /txn
+        excluded_dir_names = ["data", "metadata", "txn"]
+        root_path, filesystem = resolve_path_and_filesystem(self.delta_root_uri)
+        root_children = filesystem.get_file_info(fs.FileSelector(root_path))
+        delta_directories = [
+            child
+            for child in root_children
+            if not child.is_file and child.base_name not in excluded_dir_names
+        ]
```
anshumankomawar, line 102: Will have to change how we do this once I introduce the namespace -> table -> table_version -> stream -> partition hierarchy. But their probably isn't a better solution until we start filling out the storage interface.

anshumankomawar (approved):

deltacat/storage/rivulet/reader/dataset_metastore.py
```diff
@@ -37,29 +43,67 @@ def generate_sstables(self) -> Generator[SSTable, None, None]:
...
+        NOTE: this will be replaced by deltacat storage API.
 
         :return: a generator of accessors into the Manifests
         """
-        for uri in self.file_provider.generate_manifest_uris():
-            manifest = self.manifest_io.read(uri)
-            yield ManifestAccessor(manifest, self.file_provider, self.sst_reader)
+        # Rivulet data and SST files written to /data and /metadata
+        # Deltacat transactions written to /txn
+        excluded_dir_names = ["data", "metadata", "txn"]
```
thesalus, line 95: Minor: wouldn't it be more resilient to only target the manifest directory (until the deltacat storage API is available)?

deltacat/storage/rivulet/reader/dataset_metastore.py
```diff
@@ -37,29 +43,67 @@ def generate_sstables(self) -> Generator[SSTable, None, None]:
...
         self._max_key = None
+        self.delta_root_uri = delta_root_uri
         self.file_provider = file_provider
-        self.manifest_io = manifest_io or JsonManifestIO()
+        self.manifest_io = manifest_io or DeltacatManifestIO(delta_root_uri)
         self.sst_reader = sst_reader or JsonSstReader()
 
+    def _get_delta(self, delta_dir: str, filesystem: FileSystem) -> RivuletDelta:
+        """
+        Given a DeltaCat delta directory, find latest delta file
```
thesalus, line 70: Minor: might I suggest `_get_latest_delta`?

deltacat/storage/rivulet/reader/dataset_metastore.py
```diff
@@ -37,29 +43,67 @@ def generate_sstables(self) -> Generator[SSTable, None, None]:
...
+        Current implementation does not respect open/closed transactions, it just
+            looks for the latest revision
+        """
+        rev_directory = os.path.join(delta_dir, "rev")
+        revisions = filesystem.get_file_info(fs.FileSelector(rev_directory))
+        # Take lexicographical max
+        latest_revision = None
+        for revision in revisions:
+            latest_revision = (
+                revision if not latest_revision else max(latest_revision, revision.path)
```
thesalus, line 82: Shouldn't this be

```suggestion
                revision.path if not latest_revision else max(latest_revision, revision.path)
```

Otherwise you wouldn't be comparing the paths themselves.

thesalus (approved):
> How is a deltcat writer expected to assign stream position

Should the writer be providing a stream position in all scenarios? Is it valid for the writer to ask to append a transaction (and simply take whatever (valid) stream position that gets generated)?

deltacat/tests/storage/rivulet/writer/test_dataset_writer.py
```diff
@@ -10,8 +10,8 @@
     write_mvp_table,
     mvp_table_to_record_batches,
     validate_with_full_scan,
-    assert_data_file_extension,
     create_dataset_for_method,
+    assert_data_file_extension,
```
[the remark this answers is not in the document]
mcember: hmm not sure how this happened, if it was the linter or IDE or what

deltacat/storage/rivulet/metastore/delta.py
```diff
@@ -0,0 +1,189 @@
...
+        Ticket: https://github.com/ray-project/deltacat/issues/469
+        """
+        for sst_uri in sst_files:
+            entry_list.append(
+                ManifestEntry.of(
+                    url=sst_uri,
+                    # TODO have rivulet writer populate these values
+                    # see: https://github.com/ray-project/deltacat/issues/476
+                    meta=ManifestMeta.of(
+                        record_count=None,  # or known
```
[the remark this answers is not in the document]
mcember: it means that e.g. when writing data rivulet writer will know the record count and propagate that to manifests. Not updating code since I think the comment/linked ticket are sufficient

deltacat/storage/rivulet/metastore/delta.py
```diff
@@ -0,0 +1,189 @@
...
+            namespace=None,
+            table_name=None,
+            table_version=None,
+            stream_id=None,
+            stream_format=None,
+            partition_values=None,
+            partition_id=None,
+            # Using microsecond precision timestamp as stream position
+            # TODO consider having storage interface auto assign stream position
+            stream_position=time.time_ns(),
```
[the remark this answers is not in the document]
mcember: Yea. Not updating for now since will move automatic stream position generation behind storage/catalog interface

deltacat/storage/rivulet/metastore/delta.py
```diff
@@ -0,0 +1,189 @@
...
+                        operation_type=TransactionOperationType.CREATE,
+                        dest_metafile=delta,
+                    )
+                ]
+            ),
+        ).commit(self.root)
+        paths = tx_results[0]
+        assert (
+            len(paths) == 1
+        ), "expected delta commit transaction to write exactly 1 metafile"
```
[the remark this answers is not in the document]
mcember: This is more like a sanity check type exception - we only added one metafile and if the transaction committed more than one than something is seriously wrong

We are abstracting the low level Tx handling behind storage interface in near future

deltacat/storage/rivulet/reader/dataset_metastore.py
mcember: revision in this case is a path since it is generated by filesystem list

PR #479, opened 2025-02-02
Initial Test Setup for Converter using Spark and Pyiceberg

Zyiqin-Miranda:
## Summary
This PR adds Initial test setup for converter using Spark and Pyiceberg. It reuse the MakeFile and docker setup in DeltaCAT. 

## Testing

make test-converter

## Additional Notes

Will follow up with test cases that use Pyiceberg to generate equality deletes and Spark to read the position deletes we produced to mimic the end-to-end converter workflow and validate the results. 
For read, since Spark can't produce equality deletes currently, we should explore to generate equality deletes using Pyiceberg directly. 
For write, this test setup could also validate the position delete converter write in sense of position delete file link, naming, table properties convertion, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,14 @@ test-integration: install
 	docker-compose -f dev/iceberg-integration/docker-compose-integration.yml exec -T spark-iceberg ipython ./provision.py
 	venv/bin/python -m pytest deltacat/tests/integ -v -m integration
 
+test-converter:
+	docker-compose -f dev/iceberg-integration/docker-compose-integration.yml kill
+	docker-compose -f dev/iceberg-integration/docker-compose-integration.yml rm -f
+	docker-compose -f dev/iceberg-integration/docker-compose-integration.yml up -d
+	sleep 3
+	docker-compose -f dev/iceberg-integration/docker-compose-integration.yml exec -T spark-iceberg ipython ./provision.py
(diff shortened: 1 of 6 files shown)


pdames (approved):
LGTM - thanks!

Makefile
```diff
@@ -43,6 +43,14 @@ test-integration: install
 	docker-compose -f dev/iceberg-integration/docker-compose-integration.yml exec -T spark-iceberg ipython ./provision.py
 	venv/bin/python -m pytest deltacat/tests/integ -v -m integration
 
+test-converter:
```
pdames, line 46: This is fine for now (doesn't conflict with any other test names), but we'll probably want to change its name or roll it up within an existing suite (like integ tests) later.

Zyiqin-Miranda:
Thanks @pdames for the review, created sub-issue here: https://github.com/ray-project/deltacat/issues/486. Existing integration test is breaking, added description in the issue to fix it first.

pdames:
> Thanks @pdames for the review, created sub-issue here: #486. Existing integration test is breaking, added description in the issue to fix it first.

Great, thanks for looking into that! Would be good to see the Iceberg integration tests working again. :)

pdames:
Feel free to merge this PR when you're ready @Zyiqin-Miranda

PR #483, opened 2025-02-04
chore: support json transaction files for debugging

anshumankomawar:
## Summary

This PR introduces support for specifying the `METAFILE_FORMAT` using an environment variable.

### Usage

The added `metafile-format` can be used like
   ```
$ METAFILE_FORMAT=json python ./my-script.py

supported values: json, msgpack[default]
```

## Testing

`make test`

## Regression Risk

## Checklist

### CODE REVIEW

File: deltacat/constants.py [hunk 1]
```diff
@@ -37,6 +38,15 @@
     False,
 )
 
+# CLI Args
+METAFILE_FORMAT_KEY = "metafile-format"
```
pdames, line 42: Prefer to use standard environment variable casing here:
```suggestion
METAFILE_FORMAT_KEY = "METAFILE_FORMAT"
```

deltacat/utils/common.py
```diff
@@ -1,25 +1,46 @@
...
+    # allow hyphens in keys, normalize as underscores
+    normalized_key = key.replace("-", "_")
+
+    parser.add_argument(f"--{key}", metavar=normalized_key, type=str)
+    args, _ = parser.parse_known_args(sys.argv[1:])  # Allow unknown args
+    return getattr(args, normalized_key, None)
+
+
+def env_bool(key: str, default: bool) -> bool:
+    cli_value = get_cli_arg(key)
```
pdames, line 21: Hmm... since you can also set environment variables within the context of a single bash command, can we just let users do that and not overload the environment variables here to either look for CLI args or environment variables?

(i.e., `$ METAFILE_FORMAT=json python ./my-script.py`)

We can then better create a more formal/unified deltacat CLI experience within the context of a dedicated CLI application, similar to `dev/deploy/aws/scripts/runner.py`.

pdames (approved):

PR #484, opened 2025-02-06
WIP storage interface

mcember:
## Summary

Work in progress storage interface implementation

Currently have namespace/table/tv working. Stream is in progress

Opening PR to share progress and coordinate with other developers


deltacat/storage/model/metafile.py
```diff
@@ -128,13 +137,13 @@ def latest_revision(
...
     @staticmethod
     def new_revision(
-        revision_dir_path: str,
-        current_txn_op_type: deltacat.storage.model.transaction.TransactionOperationType,
-        current_txn_start_time: int,
-        current_txn_id: str,
-        filesystem: pyarrow.fs.FileSystem,
-        extension: Optional[str] = METAFILE_EXT,
-        success_txn_log_dir: Optional[str] = None,
+            revision_dir_path: str,
```
mcember, line 140: note: must have accidentally triggered IDE auto formatting...maybe linter will fix this?

deltacat/storage/model/metafile.py
```diff
@@ -371,8 +380,11 @@ def path(self) -> Optional[str]:
             else None
         )
 
+    def is_undefined(self) -> bool:
+        return self.get("undefined", False)
 
-class Metafile(dict):
+
+class Metafile(ABC, dict):
```
mcember, line 387: I added ABC more for clarity and to explicitly mark which abstract methods were expected for all metafiles. Only marked to_serializable currently because I noticed not all metafiles implemented it, and would prefer to have explicit implementations in child classes where possible

deltacat/catalog/model/table_definition.py
```diff
@@ -19,7 +19,7 @@ def of(
                 "table": table,
                 "tableVersion": table_version,
                 "stream": stream,
-                "nativeObject": native_object,
+                "": native_object,
```
mcember, line 22: this was accidental...will revert

deltacat/storage/main/impl.py
```diff
@@ -66,113 +75,102 @@
...
-    list_results = _list_metafiles(
-        *args,
-        metafile=metafile,
-        txn_op_type=txn_op_type,
-        **kwargs,
-    )
-    results = list_results.all_items()
-    return results[0] if results else None
-
+METAFILE = TypeVar("METAFILE", bound=Metafile)
```
mcember, line 78: using this generic didn't end up working out...will remove

deltacat/storage/main/impl.py
```diff
@@ -66,113 +75,102 @@
 )
 from deltacat.utils.common import ReadKwargsProvider
 
-
-def _get_catalog(**kwargs) -> PropertyCatalog:
```
mcember, line 70: moved helpers to bottom of file

deltacat/storage/model/metafile.py
```diff
@@ -106,13 +109,19 @@ def list_revisions(
...
+            current_txn_start_time: Optional[int] = None,
+            current_txn_id: Optional[str] = None,
+            ignore_missing_revision: bool = False,
+    ) -> Optional[MetafileRevisionInfo]:
+        """
+        Fetch latest revision of a metafile, or return None if no revisions exist
+        :param revision_dir_path: root path of directory for metafile
+        :param ignore_missing_revision: if True, will return MetafileRevisionInfo.undefined() on no revisions
+        :raises ValueError if no revisions are found AND ignore_missing_revision=False
+        """
```
flliver, line 124: having it return optional while also still having an 'ignore_missing_revision' feels like we're giving too many ways to do the same thing. I'd get rid of the flag and just return the optional.

deltacat/storage/model/metafile.py
```diff
@@ -742,15 +777,25 @@ def revisions(
...
+                    filesystem=filesystem,
+                    txn_start_time=current_txn_start_time,
+                    txn_id=current_txn_id)
+            except ValueError:
+                # This indicates the metafile has been deleted. In this case, return no results
+                return ListResult.empty()
+            if not immutable_id:
+                # This indicates the metafile does not exist. In this case, return no results
+                return ListResult.empty()
+
```
flliver, line 798: @pdames, this TODO was removed but it doesn't look like it was implemented?

mcember: I moved the TODO into id() getter in metafile.py. Partially addressed it by adding metafile.generate_new_id and id_exists. 

I assume that the TODO was referring to the fact that calling metafile.id will lazily assign it if not present, and the code was working around this by fetching the dict value directly. 

There were dozens of calls to .id, and when I tested removing the lazy assignment it was taking too much time to fully refactor out lazy assignment

deltacat/storage/main/impl.py
```diff
@@ -1,6 +1,15 @@
+"""
+Implementation of storage interface
+
+NOTE: THIS CURRENTLY DIVERGES FROM storage/interface.py!
+
+After this implementation is done, we will make storage/interface_v2.py. We then have to migrate
+  storage/iceberg/impl from storage/interface_v1.py to storage/interface_v2.py
```
pdames, line 7: If you're only diverging by adding parameters, don't worry about revving the interface into an old and new version, since all of these APIs always had `*args` and `**kwargs` available, and the additional parameters should thus continue to flow through the old interface and just be unused by existing implementations.

Also, don't worry about keeping the iceberg implementation working - this will be discarded and turned into a metadata synchronization implementation.

deltacat/storage/model/metafile.py
pdames: In other words, leave it to the caller to decide whether or not they need to raise an exception? That may create a bit more code duplication overall, but is probably fine (this is just following the convention carried over from Ray Data right now).

deltacat/storage/model/metafile.py
pdames: Hmm.. I prefer to avoid using ABC due to performance overhead (e.g., https://discuss.python.org/t/speedup-abc-issubclass/21080) and imposing these types of strict expectations on sub-classes. When you're just reading to/from disk the overhead may not be appreciable, but these types of changes have had to be reverted in the past due to how slow they start to make working with millions/billions of metafiles in a distributed execution environment like compaction on Ray.

deltacat/storage/model/locator.py
```diff
@@ -108,12 +109,25 @@ def hexdigest(self) -> str:
         for use in equality (i.e. two locators are equal if they have the same
         hexdigest) and inclusion in URLs.
         """
-        return sha1_hexdigest(self.canonical_string().encode("utf-8"))
+        # TODO uncomment for PR
+        # return sha1_hexdigest(self.canonical_string().encode("utf-8"))
+        return self._normalize_for_filesystem(self.canonical_string())
```
pdames, line 114: This might be handy to enable/disable in general via a "developer mode" or "human readable" flag.

deltacat/storage/model/metafile_utils.py
```diff
@@ -0,0 +1,26 @@
+
+
+def get_class(serialized_dict: dict):
```
pdames, line 3: I'd recommend just putting this directly in `metafile.py` to continue to maintain the 1:1 mapping between all `model` child files and the model type that the underlying code applies to.

deltacat/storage/model/metafile.py
```diff
@@ -371,8 +380,11 @@ def path(self) -> Optional[str]:
             else None
         )
 
+    def is_undefined(self) -> bool:
+        return self.get("undefined", False)
```
pdames, line 384: Instead of introducing a new "undefined" parameter that is stored/maintained separately, what do you think about just tying this back to `return bool(self.revision)` to ensure that we enforce valid/defined revision numbers as any non-zero value?

deltacat/storage/model/metafile_utils.py
mcember: putting it in metafile.py leads to a circular dependency

deltacat/storage/main/impl.py
```diff
@@ -229,13 +220,15 @@ def list_partition_deltas(
...
-    raise NotImplementedError("list_partition_deltas not implemented")
+    if isinstance(partition_like, Partition):
+        part_obj = partition_like
+    else:
+        part_obj = Partition.of(locator=partition_like, schema=None, content_types=None)
+
+    return _list(part_obj, TransactionOperationType.READ_CHILDREN, *args, **kwargs)
 
 
 def get_delta(
```
mcember, line 234: Note that partition and delta operations are AI generated. DO NOT REVIEW

deltacat/tests/_io/__init__.py
```diff
@@ -0,0 +1 @@
+# NOTE - this test module is named _io because naming it io will lead to pytest start up failing
```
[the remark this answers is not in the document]
pdames: Hmm... we may need to circle back to this and document the fix, I recall having this problem with PyCharm too a while back and fixing it with a config change found by looking up the error message.

deltacat/storage/model/metafile_utils.py
pdames: ah, yes - there are a few ways to fix that, but I usually just get rid of the import statements and reference each class by its fully qualified path (e.g. `return deltacat.storage.model.table.Table` instead of `return Table`)

PR #487, opened 2025-02-10
feat: add schema map to deltacat tableversion

anshumankomawar:
## Summary

As part of the deltacat-rivulet integration rivulet requires a way to associate names with schemas. Prior to this change Deltacat TableVersions had a SchemaList (List[Schema]). This PR updates this to a SchemaMap allowing for rivulets requirements.

## Changes

Addition of `SchemaMap`
Update TableVersion to use `SchemaMap`
## Impact

Should not have any impact on existing functionality based on existing tests.

## Testing

Added unit tests to verify `SchemaMap` functionality.
Ran `make test` to verify existing tests.

## Checklist

## Additional Notes

This is the first step in multi-schema deltacat-rivulet support.

### DIFF
--- a/deltacat/storage/main/impl.py
+++ b/deltacat/storage/main/impl.py
@@ -255,6 +255,7 @@ def _resolve_latest_table_version_id(
     return table.latest_table_version
 
 
+
 def list_namespaces(*args, **kwargs) -> ListResult[Namespace]:
     """
     Lists a page of table namespaces. Namespaces are returned as list result
@@ -402,6 +403,7 @@ def list_stream_partitions(stream: Stream, *args, **kwargs) -> ListResult[Partit
     )
 
 
+
 def list_deltas(
     namespace: str,
     table_name: str,
@@ -519,6 +521,7 @@ def list_partition_deltas(
     return filtered_deltas
 
 
+
 def get_delta(
     namespace: str,
     table_name: str,
@@ -568,6 +571,7 @@ def get_delta(
         metafile=delta,
         **kwargs,
     )
+
 
 
 def get_latest_delta(
@@ -621,6 +625,7 @@ def get_latest_delta(
     )
 
 
+
 def download_delta(
     delta_like: Union[Delta, DeltaLocator],
     table_type: TableType = TableType.PYARROW,
@@ -643,6 +648,7 @@ def download_delta(
     raise NotImplementedError("download_delta not implemented")
 
 
+
 def download_delta_manifest_entry(
     delta_like: Union[Delta, DeltaLocator],
     entry_index: int,
@@ -661,6 +667,7 @@ def download_delta_manifest_entry(
     NOTE: The entry will be downloaded in the current node's memory.
     """
     raise NotImplementedError("download_delta_manifest_entry not implemented")
+
 
 
 def get_delta_manifest(
@@ -707,6 +714,7 @@ def create_namespace(
     return namespace
 
 
+
 def update_namespace(
     namespace: str,
     properties: Optional[NamespaceProperties] = None,
@@ -743,6 +751,7 @@ def update_namespace(
         filesystem=catalog.filesystem,
     )
     return namespace
+
 
 
 def create_table_version(
@@ -862,6 +871,7 @@ def create_table_version(
     return stream
 
 
+
 def update_table(
     namespace: str,
     table_name: str,
@@ -904,6 +914,7 @@ def update_table(
         catalog_root_dir=catalog.root,
         filesystem=catalog.filesystem,
     )
+
 
 
 def update_table_version(
@@ -1033,6 +1044,7 @@ def update_table_version(
         catalog_root_dir=catalog.root,
         filesystem=catalog.filesystem,
     )
+
 
 
 def stage_stream(
@@ -1281,6 +1293,7 @@ def get_staged_stream(
     )
 
 
+
 def get_stream(
     namespace: str,
     table_name: str,
@@ -1519,6 +1532,7 @@ def commit_partition(
     return partition
 
 
+
 def delete_partition(
     stream_locator: StreamLocator,
     partition_values: Optional[PartitionValues] = None,
@@ -1588,6 +1602,7 @@ def get_staged_partition(
     )
 
 
+
 def get_partition(
     stream_locator: StreamLocator,
     partition_values: Optional[PartitionValues] = None,
@@ -1634,6 +1649,7 @@ def get_partition(
     )
 
 
+
 def stage_delta(
     data: Union[LocalTable, LocalDataset, DistributedDataset, Manifest],
     partition: Partition,
@@ -1662,6 +1678,7 @@ def stage_delta(
     raise NotImplementedError("stage_delta not implemented")
 
 
+
 def commit_delta(delta: Delta, *args, **kwargs) -> Delta:
     """
     Registers a new delta with its associated target table version and
@@ -1672,6 +1689,7 @@ def commit_delta(delta: Delta, *args, **kwargs) -> Delta:
     position in the target partition.
     """
     raise NotImplementedError("commit_delta not implemented")
+
 
 
 def get_namespace(namespace: str, *args, **kwargs) -> Optional[Namespace]:
@@ -1697,6 +1715,7 @@ def namespace_exists(namespace: str, *args, **kwargs) -> bool:
     )
 
 
+
 def get_table(namespace: str, table_name: str, *args, **kwargs) -> Optional[Table]:
     """
     Gets table metadata for the specified table. Returns None if the given
@@ -1720,6 +1739,7 @@ def table_exists(namespace: str, table_name: str, *args, **kwargs) -> bool:
         metafile=Table.of(locator=locator),
         **kwargs,
     )
+
 
 
 def get_table_version(
@@ -1749,6 +1769,7 @@ def get_table_version(
     )
 
 
+
 def get_latest_table_version(
     namespace: str, table_name: str, *args, **kwargs
 ) -> Optional[TableVersion]:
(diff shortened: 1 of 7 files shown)


deltacat/storage/model/metafile.py
anshumankomawar, line 1: These are changes proposed by the linter.

deltacat/storage/model/schema.py
```diff
@@ -634,18 +641,140 @@ def _populate_fields(
...
+                    schema_data
+                    if isinstance(schema_data, Schema)
+                    else Schema(schema_data)
+                )
+                mapping.insert(None, schema_obj)
+        else:
+            raise ValueError(f"Cannot create SchemaMap from {item}")
+        return mapping
+
+    def _generate_default_name(self, schema: Schema) -> str:
```
anshumankomawar, line 686: Maybe we don't need this but added this to support the existing SchemaList implementation. Should allow us to pass in a 
`schemas=[schema1, schema2, schema3]` without an associated name.

deltacat/tests/storage/model/test_schema_map.py
```diff
@@ -0,0 +1,192 @@
...
+        SchemaMap.of(42)
+
+
+def test_insert_default_name(schema_a):
+    smap = SchemaMap()
+    smap.insert(None, schema_a)
+    # Expect a default key to be generated.
+    key = list(smap.keys())[0]
+    assert key.startswith("default_schema_")
+    assert smap[key].equivalent_to(schema_a)
```
thesalus, line 78: A couple other test cases:

1. What should happen if you insert the same schema twice (using `None` as the key) where the schema has a name?
2. What should happen if you insert the same schema twice (using `None` as the key) where the schema does not have a name?

thesalus (commented):
The preservation of the list of schemas and the auto-generation of the schema names doesn't smell right to me. But I think this needs some input from someone with a bit more visibility into the future and purpose of the schema list.

If most interaction with schema requires a reference to a schema name then those need to be predictable and/or easily-determinable. Right now if I were to call `SchemaMap.of(schemaList)` then I won't know what the name is for any given schema in that list (aside from iterating through the SchemaMap and doing an equivalency check). Similarly, if I were to call `schemaMap.insert(None, schema)`, I would not know what the name is.

deltacat/storage/model/metafile.py
anshumankomawar: I think these will be included as part of matt's pr, I pulled in the unfinished storage impl.

anshumankomawar:
> The preservation of the list of schemas and the auto-generation of the schema names doesn't smell right to me. But I think this needs some input from someone with a bit more visibility into the future and purpose of the schema list.
> 
> If most interaction with schema requires a reference to a schema name then those need to be predictable and/or easily-determinable. Right now if I were to call `SchemaMap.of(schemaList)` then I won't know what the name is for any given schema in that list (aside from iterating through the SchemaMap and doing an equivalency check). Similarly, if I were to call `schemaMap.insert(None, schema)`, I would not know what the name is.

I agree we definitely need more insight into if `ordering` even needs to be maintained. If this isn't a concern we can get rid of the name generation and rely entirely on provided names. In this implementation the assumption is that if you create a SchemaMap with a list you are ignoring the key, value approach entirely and you can call the `get_schemas` function to return a simple list of schemas (maybe this can be renamed). This is to align with the older SchemaList approach. 

Ideally we should decide if ordering is required and go from there..

deltacat/tests/storage/model/test_schema_map.py
anshumankomawar: Will add tests to cover those.

deltacat/storage/model/table_version.py
```diff
@@ -244,9 +244,13 @@ def to_serializable(self) -> TableVersion:
...
         serializable.schemas = (
-            [_.serialize().to_pybytes() for _ in serializable.schemas]
-            if serializable.schemas
+            {
+                key: schema.serialize().to_pybytes()
+                for key, schema in self.schemas.items()
+            }
+            if self.schemas
             else None
         )
```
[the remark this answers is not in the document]
anshumankomawar: From what I can tell serializing/deserializing does preserve the order and key names. I was able to run the existing tests, specifically the `test_metafile_io.py` test which checks table version equivalency before and after serialization.

deltacat/storage/model/schema.py
```diff
@@ -634,18 +641,140 @@ def _populate_fields(
...
+        raises: None.
+        """
+        return list(self.values())
+
+    def equivalent_to(self, other: Any) -> bool:
+        """
+        Equivalence check for SchemaMap objects.
+
+        param: other (Any): Another mapping (dict or SchemaMap) to compare against.
+        returns: bool indicating whether both mappings have the same keys in the same order and equivalent Schema objects.
```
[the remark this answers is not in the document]
anshumankomawar: Left a high level comment.

deltacat/storage/model/schema.py
```diff
@@ -620,7 +626,8 @@ def _populate_fields(
         max_field_id = (
             visitor_dict["maxFieldId"] + len(field_ids_to_fields)
         ) % MAX_FIELD_ID_EXCLUSIVE
-        field_id = Field.of(field).id or max_field_id
+        _field = Field.of(field)
+        field_id = _field.id if _field is not None else max_field_id
```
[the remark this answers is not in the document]
anshumankomawar: I don't believe so. I guess this is one of the benefits of a dynamically typed language. To avoid this we would probably have to provide types for variables as well but at that point it would be better to just use a staticly typed language.

deltacat/storage/model/table_version.py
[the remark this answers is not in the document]
thesalus: The reason I asked is because this code is converting the OrderedDict to a regular dict (which I assumed would not preserve ordering).

However, it looks like dicts started preserving ordering since Python 3.7.

deltacat/storage/main/impl.py
```diff
@@ -27,7 +27,8 @@
     LocalDataset,
```
mcember, line 27: Note: Patrick merged storage impl into 2.0 so you rebase your branch from 2.0 and make a new revision of this PR

deltacat/storage/model/schema.py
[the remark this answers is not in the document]
mcember: I don't think you can check this from a type safety perspective but I bet you could set a linting rule on it

deltacat/storage/model/schema.py
```diff
@@ -634,18 +641,122 @@ def _populate_fields(
         field_ids_to_fields[field_id] = field
 
 
-class SchemaList(List[Schema]):
+class SchemaMap(OrderedDict):
```
mcember, line 644: Does order here matter?

deltacat/storage/model/schema.py
```diff
@@ -634,18 +641,122 @@ def _populate_fields(
...
-    def __getitem__(self, item):
-        val = super().__getitem__(item)
-        if val is not None and not isinstance(val, Schema):
-            self[item] = val = Schema(val)
-        return val
+    def of(item: Union[Dict[str, Any], List[Any]]) -> SchemaMap:
+        """
+        Create a SchemaMap from a dictionary or a list of schema data.
+
+        param: item (Union[Dict[str, Any], List[Any]]): A dict mapping names to schema data or a list of schema data.
```
mcember, line 661: I'm assuming the types here are Any because we are flexible and accept multiple types (for a schema, or for a field)?

This type signature and doc doesn't give me a lot of clarity for how to construct schemas

deltacat/storage/model/schema.py
mcember: I think we need a broader discussion on what it means to have multiple schemas and how users are going to use the different schemas. 

I think it is reasonable though to just assign names based on ordinal/index so that if users want they can treat it like a list instead of a dict

deltacat/storage/model/schema.py
```diff
@@ -634,18 +641,122 @@ def _populate_fields(
...
+
+    def get_schemas(self) -> List[Schema]:
+        """
+        Retrieve all stored schemas as a list.
+
+        param: None.
+        returns: List[Schema] containing all schemas in insertion order.
+        raises: None.
+        """
+        return list(self.values())
```
mcember, line 741: sanity check - if values() is empty, does this return an empty list, or a list whose first element is empty list?

mcember (approved):
In general this LGTM. Please rebase from 2.0 and address comments. Also discuss any backwards compatibility concerns as we talked about on slack

deltacat/storage/model/schema.py
anshumankomawar: Don't have an answer for this as of yet, but since python 3.9 a Dict preserves insertion order by default. OrderedDict just makes it more explicit.

deltacat/storage/model/schema.py
anshumankomawar: Returns an empty list.

deltacat/storage/main/impl.py
```diff
@@ -77,19 +78,18 @@
...
     Lists a page of table namespaces. Namespaces are returned as list result
     items.
     """
     placeholder_ns = Namespace.of(NamespaceLocator.of("placeholder"))
     return _list_metafiles(
-        placeholder_ns,
-        TransactionOperationType.READ_SIBLINGS,
-        *args,
-        **kwargs
+        placeholder_ns, TransactionOperationType.READ_SIBLINGS, *args, **kwargs
```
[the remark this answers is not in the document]
anshumankomawar: Will keep that in mind.
>  but if you put a comma after **kwargs it would keep all args on separate lines
Updating the comma resolved the issue.

PR #494, opened 2025-03-05
[Converter][Test]Add working end-to-end converter compute test with Minio S3 endpoint

Zyiqin-Miranda:
## Summary
This PR adds working end-to-end converter find deletes compute happy case test. 
With 1. Minio S3 endpoint replaced in S3 file system 2. Daft download parquet file mocked, 
We are able to test the compute part end-to-end. 
Future dev work adding new features can use the same setup to test correctness.

## Rationale

## Changes

## Impact

## Testing

## Regression Risk

## Checklist

## Additional Notes

1. More test cases pending: null primary key, multiple primary key concatenating.
2. We'd also want to add using Pyspark to read position delete we produced and assert same record count.

### DIFF
--- a/deltacat/compute/converter/steps/convert.py
+++ b/deltacat/compute/converter/steps/convert.py
@@ -17,14 +17,14 @@ logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 def convert(convert_input: ConvertInput):
     files_for_each_bucket = convert_input.files_for_each_bucket
     convert_task_index = convert_input.convert_task_index
-    iceberg_warehouse_bucket_name = convert_input.iceberg_warehouse_bucket_name
+    iceberg_table_warehouse_prefix = convert_input.iceberg_table_warehouse_prefix
     identifier_fields = convert_input.identifier_fields
     compact_small_files = convert_input.compact_small_files
     position_delete_for_multiple_data_files = (
         convert_input.position_delete_for_multiple_data_files
     )
     max_parallel_data_file_download = convert_input.max_parallel_data_file_download
-
+    s3_file_system = convert_input.s3_file_system
     if not position_delete_for_multiple_data_files:
         raise NotImplementedError(
             f"Distributed file level position delete compute is not supported yet"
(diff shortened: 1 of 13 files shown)


deltacat/compute/converter/steps/convert.py
```diff
@@ -34,16 +34,27 @@ def convert(convert_input: ConvertInput):
 
     logger.info(f"Starting convert task index: {convert_task_index}")
     data_files, equality_delete_files, position_delete_files = files_for_each_bucket[1]
+    # Get string representation of partition value out of Record[partition_value]
+    partition_value_str = (
+        files_for_each_bucket[0].__repr__().split("[", 1)[1].split("]")[0]
```
pdames, line 39: The code to extract partition values looks pretty unfortunately opaque (but I understand this happens semi-regularly, esp. when trying to access details like this that Iceberg prefers to hide from external users). Maybe we can at least put these one-liners inside of appropriately named helper methods?

deltacat/compute/converter/utils/iceberg_columns.py
```diff
@@ -3,14 +3,17 @@
 
 
 def _get_iceberg_col_name(suffix):
-    return f"{suffix}"
+    return suffix
 
 
 _ORDERED_RECORD_IDX_COLUMN_NAME = _get_iceberg_col_name("pos")
 _ORDERED_RECORD_IDX_COLUMN_TYPE = pa.int64()
+_ORDERED_RECORD_IDX_FIELD_METADATA = {b"PARQUET:field_id": "2147483545"}
```
pdames, line 11: Can we (1) leave a link back to the relevant part of the Iceberg spec like https://iceberg.apache.org/spec/#reserved-field-ids so that we know where this magic number came from going forward and (2) store the field ID in a global constant to reuse here and other places where we reuse it?

deltacat/compute/converter/utils/iceberg_columns.py
```diff
@@ -32,7 +35,10 @@ def append_record_idx_col(table: pa.Table, ordered_record_indices) -> pa.Table:
 
 _FILE_PATH_COLUMN_NAME = _get_iceberg_col_name("file_path")
 _FILE_PATH_COLUMN_TYPE = pa.string()
+_FILE_PATH_FIELD_METADATA = {b"PARQUET:field_id": "2147483546"}
```
pdames, line 38: Same comment here - can we leave a link back to Iceberg spec reserved field IDs and store the field ID in a global constant (e.g., either in this file or in `constants.py`)?

pdames (approved):

Zyiqin-Miranda:
Not touching compactor code, so performance alert is unrelated, addressed all comments, merging.

PR #496, opened 2025-03-06
fix: add base64 encoding before schema serialization

anshumankomawar:
## Summary

Ensure binary schema data is serialized as base64.

## Testing

Ran related test..
```
======================================================================================= test session starts ========================================================================================
platform darwin -- Python 3.9.6, pytest-7.2.0, pluggy-1.5.0 -- /Volumes/workplace/multimodal/deltacat/venv/bin/python
cachedir: .pytest_cache
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Volumes/workplace/multimodal/deltacat, configfile: pytest.ini
plugins: Faker-33.1.0, requests-mock-1.11.0, cov-4.0.0, mock-3.14.0, benchmark-4.0.0
collected 606 items / 605 deselected / 1 selected

deltacat/tests/storage/model/test_metafile_io.py::TestMetafileIO::test_replace_partition PASSED

=========================================================================== 1 passed, 605 
```

Also ran sample script to verify schema was readable in metafile.

## Regression Risk

Need to update existing metafile test cases to run through a JSON and msgpack variant of each test.

## Checklist


pdames (approved):
LGTM - thanks for fixing this! Feel free to merge once we get all checks passing, and when we've restricted this behavior to just JSON metafiles.

pdames:
Ah - before merging, can you make sure that the base64 encoding/decoding is only enabled when the format is set to JSON before merging (since we don't need or want to do base64 encoding for the default `msgpack `path)? Also, we should double-check whether or not that's the root cause of the current test suite failures.

PR #498, opened 2025-03-07
Initial Working Prototype of DeltaCAT FS API w/ Limited Catalog & Namespace Get/Put/Copy Functionality

pdames:
## Summary

Initial WIP draft of DeltaCAT FS API w/ limited functionality for creating/getting/copying catalogs and namespaces (see `test_deltacat_api.py` for an example). Also adds bug fixes and refactoring of catalog properties to support multi-process-safe named catalog registration in local or distributed Ray apps/clusters.

## Checklist

### DIFF
--- a/deltacat/api.py
+++ b/deltacat/api.py
@@ -1,0 +1,165 @@
+from typing import Any
+
+
+import deltacat as dc
+
+
+def copy(source, destination):
+    src_parts = source.split("/")
+    src_parts = [part for part in src_parts if part]
+    dst_parts = destination.split("/")
+    dst_parts = [part for part in dst_parts if part]
+    if not dc.is_initialized():
+        raise ValueError("Catalog not initialized.")
+    if len(src_parts) != len(dst_parts) and len(src_parts) != len(dst_parts) + 1:
+        # TODO(pdames): Better error message.
+        raise ValueError(
+            f"Cannot copy {source} to {destination}. "
+            f"Source and destination must share the same type."
+        )
+    src_obj = get(source)
+    if len(src_parts) == 1:
+        # copy the given catalog
+        raise NotImplementedError
+    elif len(src_parts) == 2:
+        # TODO(pdames): Make catalog specification optional if there is only
+        #  one catalog (e.g., auto-retrieve src_parts[0]/dst_parts[0])
+        # copy the given namespace
+        src_namespace_name = src_parts[1]
+        dst_catalog_name = dst_parts[0]
+        dst_namespace_name = dst_parts[1] if len(dst_parts) >= 2 else src_namespace_name
+        new_namespace = dc.create_namespace(
+            namespace=dst_namespace_name,
+            properties=src_obj.properties,
+            catalog=dst_catalog_name,
+        )
+        return new_namespace
+    elif len(src_parts) == 3:
+        # copy the given table
+        raise NotImplementedError
+    elif len(src_parts) == 4:
+        # copy the given table version
+        raise NotImplementedError
+    elif len(src_parts) == 5:
+        # copy the given stream
+        raise NotImplementedError
+    elif len(src_parts) == 6:
+        # copy the given partition
+        raise NotImplementedError
+    elif len(src_parts) == 7:
+        # copy the given partition delta
+        raise NotImplementedError
+    raise ValueError(f"Invalid path: {src_parts}")
+
+
+def concat(source, destination):
+    pass
+
+
+def delete(source):
+    pass
+
+
+def move(source, destination):
+    pass
+
+
+def list(path):
+    pass
+
+
+def get(path) -> Any:
+    parts = path.split("/")
+    parts = [part for part in parts if part]
+    if not dc.is_initialized():
+        # TODO(pdames): Re-initialize DeltaCAT with all catalogs from the
+        #  last session.
+        raise ValueError("Catalog not initialized.")
+    if len(parts) == 1:
+        # TODO(pdames): Save all catalogs registered from the last session on
+        #  disk so that users don't need to re-initialize them every time.
+        # get the given catalog
+        catalog_name = parts[0]
+        return dc.get_catalog(catalog_name)
+    elif len(parts) == 2:
+        # get the given namespace
+        catalog_name = parts[0]
+        namespace_name = parts[1]
+        return dc.get_namespace(
+            namespace=namespace_name,
+            catalog=catalog_name,
+        )
+    elif len(parts) == 3:
+        # get the given table
+        raise NotImplementedError
+    elif len(parts) == 4:
+        # get the given table version
+        raise NotImplementedError
+    elif len(parts) == 5:
+        # get the given stream
+        raise NotImplementedError
+    elif len(parts) == 6:
+        # get the given partition
+        raise NotImplementedError
+    elif len(parts) == 7:
+        # get the given partition delta
+        raise NotImplementedError
+    raise ValueError(f"Invalid path: {path}")
+
+
+def put(path, *args, **kwargs) -> Any:
+    parts = path.split("/")
+    parts = [part for part in parts if part]
+    if len(parts) == 1:
+        # TODO(pdames): Save all catalogs registered from the last session on
+        #  disk so that users don't need to re-initialize them every time.
+        # register the given catalog
+        catalog_name = parts[0]
+        return dc.put_catalog(catalog_name, *args, **kwargs)
+    elif len(parts) == 2:
+        # register the given namespace
+        catalog_name = parts[0]
+        namespace_name = parts[1]
+        if not dc.is_initialized():
+            # TODO(pdames): Re-initialize DeltaCAT with all catalogs from the
+            #  last session.
+            raise ValueError("Catalog not initialized.")
+        new_namespace = dc.create_namespace(
+            namespace=namespace_name,
+            catalog=catalog_name,
+            *args,
+            **kwargs,
+        )
+        return new_namespace
+    elif len(parts) == 3:
+        # register the given table
+        raise NotImplementedError
+    elif len(parts) == 4:
+        # register the given table version
+        raise NotImplementedError
+    elif len(parts) == 5:
+        # register the given stream
+        raise NotImplementedError
+    elif len(parts) == 6:
+        # register the given partition
+        raise NotImplementedError
+    elif len(parts) == 7:
+        # register the given partition delta
+        raise NotImplementedError
+    raise ValueError(f"Invalid path: {path}")
+
+
+def exists(path):
+    pass
+
+
+def query(path, expression):
+    pass
+
+
+def tail(path):
+    pass
+
+
+def head(path):
+    pass
--- a/deltacat/api.py
+++ b/deltacat/api.py
@@ -53,19 +53,19 @@ def copy(source, destination):
 
 
 def concat(source, destination):
-    pass
+    raise NotImplementedError
 
 
 def delete(source):
-    pass
+    raise NotImplementedError
 
 
 def move(source, destination):
-    pass
+    raise NotImplementedError
 
 
 def list(path):
-    pass
+    raise NotImplementedError
 
 
 def get(path) -> Any:
@@ -150,16 +150,16 @@ def put(path, *args, **kwargs) -> Any:
 
 
 def exists(path):
-    pass
+    raise NotImplementedError
 
 
 def query(path, expression):
-    pass
+    raise NotImplementedError
 
 
 def tail(path):
-    pass
+    raise NotImplementedError
 
 
 def head(path):
-    pass
+    raise NotImplementedError
--- a/deltacat/catalog/model/catalog.py
+++ b/deltacat/catalog/model/catalog.py
@@ -8,7 +8,7 @@ from functools import partial
 import ray
 
 from deltacat import logs
-from deltacat.catalog import interface as catalog_interface
+from deltacat.catalog.main import impl as deltacat_catalog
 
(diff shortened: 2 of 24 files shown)


deltacat/api.py
```diff
@@ -0,0 +1,165 @@
+from typing import Any
+
+
+import deltacat as dc
+
+
+def copy(source, destination):
```
mcember, line 7: What is the motivation for this? I don't have a clear mental model for the breadth of what this higher-level-than-catalog API is doing

Would be good to add some pydoc explaining it (both for end user and for reviewability)

mcember: What better name can we give this than API? Users may see this and assume that this is what the deltacat library does, not realizing this is basically just a sort of router to configure and fetch different catalog implementations

deltacat/tests/test_deltacat_api.py
```diff
@@ -0,0 +1,39 @@
...
+        # Initialize DeltaCAT with two local catalogs.
+        dc.put("test_catalog_1", root=cls.temp_dir_1)
+        dc.put("test_catalog_2", root=cls.temp_dir_2)
+
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree(cls.temp_dir_1)
+        shutil.rmtree(cls.temp_dir_2)
+
+    def test_cross_catalog_namespace_copy(self):
```
mcember, line 20: In the case of namespace copy here, you're initializing a new namespace in the destination catalog. But it's not a deep copy

How will this look for table, table version, etc? For those objects you would have to initialize their parents too. You may get into some weird edge cases where e.g. the metafile type that you're cloning is not compatible with the parents (e.g. you clone a partition but it uses different keys than the destination). 

I'm not grokking what the utility of copy is and how you will implement it across all metafile types

deltacat/api.py
```diff
@@ -0,0 +1,165 @@
...
+def list(path):
+    raise NotImplementedError
+
+
+def get(path) -> Any:
+    parts = path.split("/")
+    parts = [part for part in parts if part]
+    if not dc.is_initialized():
+        # TODO(pdames): Re-initialize DeltaCAT with all catalogs from the
+        #  last session.
```
mcember, line 76: what is the last session? Where would it be persisted?

deltacat/api.py
```diff
@@ -0,0 +1,165 @@
...
+
+def move(source, destination):
+    raise NotImplementedError
+
+
+def list(path):
+    raise NotImplementedError
+
+
+def get(path) -> Any:
```
mcember, line 71: Are get and put here meant to configure catalogs, or metafile objects within the catalogs? I'm not sure how this relates to catalog/model/catalog.py

deltacat/api.py
```diff
@@ -0,0 +1,165 @@
...
+        # get the given partition delta
+        raise NotImplementedError
+    raise ValueError(f"Invalid path: {path}")
+
+
+def put(path, *args, **kwargs) -> Any:
+    parts = path.split("/")
+    parts = [part for part in parts if part]
+    if len(parts) == 1:
+        # TODO(pdames): Save all catalogs registered from the last session on
```
mcember, line 114: When we developed rivulet we created local catalogs by default in ${working directory}/.riv-meta-{dataset}. I ported a similar thing to the default catalog root (if none provided) to initialize in ${working directory}/

We should discuss what all directories are used by default. E.g. do we want to by default create all metadata in ~/.deltacat, including child catalogs or a top level config of all catalogs

deltacat/catalog/catalog_properties.py
```diff
@@ -1,153 +0,0 @@
-from __future__ import annotations
-from typing import Optional
-
-import pyarrow
-from deltacat.constants import DELTACAT_CATALOG_PROPERTY_ROOT
-
-from deltacat.utils.filesystem import resolve_path_and_filesystem
-
-"""
-Property catalog is configured globally (at the interpreter level)
```
mcember, line 10: Do all of the capabilities here have an analog in the new approach?

I'm happy to get rid of the global static configuration. I need to keep reading code but I'm not sure if you ported over the defaulting logic and the specific env variables

deltacat/catalog/model/catalog.py
```diff
@@ -44,6 +44,19 @@ def __reduce__(self):
...
+        if self._native_object:
+            string_rep += f"native_object={self._native_object})"
+        return string_rep
+
+    def __repr__(self):
+        return self.__str__()
+
 
 @ray.remote
 class Catalogs:
```
mcember, line 62: Does the initialization logic in Catalogs work even if the user is running deltacat locally / without Ray? If the idea is that deltacat/api.py uses catalogs to initialize and persist catalog implementations, then Catalogs initializes Ray, doesn't that mean the user is essentially always going to initialize ray?

mcember (changes requested):
I would like to discuss a few topics before we commit to this approach

deltacat/api.py
pdames: A session starts when deltacat is initialized, and ends when it is terminated. Zero to N catalogs are associated with each session. The session catalogs and their config should be persisted to a file on disk so that it can be restored into the `all_catalogs` Ray Actor the next time DeltaCAT is initialized without explicit overrides of registered ctalogs (e.g., all of a user's registered catalogs and their `CatalogProperties` could be saved in a location like `~/.dc/config`).

deltacat/tests/test_deltacat_api.py
pdames: Intention to support this case clearly/correctly is to provide a `recursive` flag, similar to the Linux `cp -r` command.

mcember: In that case we would look at completed transactions and only copy files pertaining to completed transactions? The sort of immutability guarantees here are really nice

Zyiqin-Miranda (approved):

deltacat/api.py
pdames: Motivation is to support a lightweight, user-facing DeltaCAT client API/CLI driven by familiar Linux FS-style commands. Ideally, this should be lower-profile and much more intuitive to a new user than the current `Catalog` API (which provides roughly equivalent functionality and signatures to Iceberg's `Catalog` API, and I thus expect to be better suited for similar use-cases as Iceberg's Catalog API like technical integration, say, compute framework and catalog service owners). Also, to be clear, the `Catalog` API itself is also intended to support routing between different named catalogs and implementations via `delegate.py` (the act of doing so is just more verbose, as illustrated in `examples/iceberg`).

The intention is to make this the primary API that DeltaCAT end users interacts with, to avoid the unnecessary verbosity of `Catalog` APIs (unless they prefer that model), and even greater verbosity/understanding required to interact with the `Storage` APIs. 

The simplifications that this API provides can be demonstrated by trying to rewrite the tests in `test_deltacat_api.py` directly against the Catalog API instead. Some open questions remain around whether or not the current set of APIs is appropriate for (1) writing data to a table and/or rivulet dataset (2) crafting the UX around inference of default names as part of this process vs. forcing a user to always provide fully qualified paths to all of these APIs (one idea to resolve this problem is to make the level of the path in these APIs clear via URI schemes like `dc put 'table://my_table', {data_or_paths}` vs. `dc put 'catalog://my_catalog', {catalog_init_args}`)

Here are some high-level cases that I'd like to document in an associated issue (while keeping in mind that `source` and `dest` also need to be extended to support proper URIs vs. the current implicit DeltaCAT catalog paths only handled in this PR):

 1. `dc cp {source} {dest}`
   a. Stage changes to a local catalog (e.g., on your laptop), and then copy those same changes into a remote catalog (use should specify a force flag like `-f` to replace an existing object at `dest`).
   b. Import an Iceberg/Hudi/Delta catalog at `source` to a DeltaCAT catalog at `dest` (ideally this only translates metadata).
   c. Export the DeltaCAT catalog/object at `source` to the Iceberg/Hudi/Delta dataset at `dest` (assumes the artifacts exist at `dest` to either identify its format type, or additional args will be required to telegraph the desired `dest` type).
   d. Copy a HuggingFace dataset at `source` to a DeltaCAT dataset at `dest`.
   e. Register all files under the S3 URI at `source` to a DeltaCAT dataset at `dest`.
 2. `dc get {path}`
   a. Fetch the unique DeltaCAT metadata identified by the fully-qualified human readable path (e.g., a DeltaCAT Catalog/Namespace/Table/TableVersion/Stream/Partition/Delta)
 3. `dc put {path} {arguments}`
   a. Create the DeltaCAT metadata at the fully-qualified human readable path (e.g., a DeltaCAT Catalog/Namespace/Table/TableVersion/Stream/Partition/Delta). Arguments are mapped to the associated `create*` command for the target object (use should specify a force flag like `-f` to replace an existing object at `path`).
4. `dc cat {source} {dest}`
   a. Replay the transactions from the DeltaCAT catalog at `source` to the catalog at `dest`. (e.g. to support staging changes to a local catalog, and then merging those changes into a remote catalog (while ensuring that there are no transaction conflicts).
   b. Append namespace tables, table partitions, or partition deltas from the DeltaCAT object at `source` to the object at `dest`.
5. `dc exists {path}`
   a. Check if the object at the given at the given path exists (e.g., for DeltaCAT catalogs, calls into the underlying `exists*` catalog/storage APIs)
6. `dc list {path}`
   a. List all objects at the given path (e.g., for DeltaCAT catalogs, calls into the underlying `list*` catalog/storage APIs)
7. `dc delete {path}`
   a. Delete object at the given path (e.g., for DeltaCAT catalogs, calls into the underlying `delete*` catalog/storage APIs).
8. `dc query {expression}`
   a. Evaluate the given SQL query expression (e.g., delegate to Daft, Spark, DuckDB, etc. and rely on their deltacat scan operators for query fulfillment).
9. `dc tail {path}`
   a. List the last N elements at the given path (perhaps redundant with `dc ls` command?)
   b. Use `-f` flag to track changes to the given `path`  in near-real-time (e.g., equivalent to streaming a live listing of objects appended to the given path).
10. `dc head {path}`
   a. List the first N elements at the given path (perhaps redundant with `dc ls` command?)

deltacat/api.py
pdames: Both, but I also expect `get`, `put`, and other APIs here to evolve to use fully qualified URIs, with the scheme indicating either the top-level DeltaCAT object type that they intend to operate on like `catalog://` or `namespace://`. Right now in this PR, all paths are implicitly scheme-less fully qualified `catalog://` URIs.

deltacat/api.py
pdames: I'm currently thinking to save session metadata in `~/.dc/config` by default for now (this also allows a parallel deltacat catalog to exist in a separate/default `~/.deltacat` directory).

deltacat/api.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def copy(source, destination):
+    src_parts = source.split("/")
+    src_parts = [part for part in src_parts if part]
+    dst_parts = destination.split("/")
+    dst_parts = [part for part in dst_parts if part]
+    if not dc.is_initialized():
+        raise ValueError("Catalog not initialized.")
+    if len(src_parts) != len(dst_parts) and len(src_parts) != len(dst_parts) + 1:
```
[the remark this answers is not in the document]
pdames: Good question - I'll leave a comment clarifying this, but the `+ 1` is to support running copy commands like `dc cp 'my_catalog_1/my_namespace_1' 'my_catalog_2'` where the namespace copied to `my_catalog_2` is implicitly `my_namespace_1` (i.e. a rename is not desired as part of the copy, and it's redundant to state that the destination name remains the same as the source).

deltacat/catalog/model/catalog.py
pdames: Right now Ray will always be initialized either locally on your laptop or on a distributed cluster as an easy way to (1) provide serializable isolation on concurrent updates to all registered catalogs via an actor, (2) make the same code functional within parallel local processes, or an arbitrarily large number of distributed parallel remote processes.

deltacat/tests/test_deltacat_api.py
pdames: Exactly! And, as long as everything we do ultimately trickles down to the storage APIs (which in turn trickles down to our metafile and transaction APIs), then we should automatically honor completely transaction boundaries. Replaying transactions at a full catalog level from one catalog to another is something that we haven't yet implemented, but also seems simple enough to do in theory.

deltacat/api.py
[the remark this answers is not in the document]
Zyiqin-Miranda: Thanks for the explanation, not directly related to this length check, but I was thinking that whether we want to support more lower level coping like cp specific partition/delta? Like users could either choose certain `partition/delta` or `all partitions/deltas`, but they probably don't want to specify `all partition_values/delta_stream_positions`?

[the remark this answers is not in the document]
pdames: Yep - I think this is similar to the Q&A with @mcember above related to using the a `dc cp -r` flag to telegraph that a user would like to copy all deltas in a partition to another destination (e.g., `dc cp -r 'catalog://my_src_catalog/my_namespace/my_table/v.1/my_partition' 'catalog://my_dst_catalog') should also by just auto-populating the dst namespace/table/table-version/partition and all partition deltas from the source (which I think implies that, going forward, the general pattern is that the `dest` for a _recursive_ copy just needs to share a prefix with the source).

pdames:
Created a high-level tracking issue at `https://github.com/ray-project/deltacat/issues/510` to track the sub-issues required to consider the work started here "complete", which will involve a series of larger PRs on this topic. Will resolve conflicts and merge what we have here to persist a common starting point for myself and others to work from.

PR #499, opened 2025-03-07
Prototype of Dev/catalog interface

anshumankomawar:
## Summary

This is a WIP progress implementation of the v2 catalog interface table functions. 

## Checklist

## Additional Notes


deltacat/catalog/v2/catalog_impl.py
```diff
@@ -80,15 +104,86 @@ def create_table(
...
+        partition_scheme=partition_scheme,
+        sort_keys=sort_keys,
+        table_version_description=description,
+        table_description=description,
+        table_properties=table_properties,
+        lifecycle_state=lifecycle_state or LifecycleState.ACTIVE,
+        catalog=catalog,
+    )
+
+    # What do we do with this dataset?
```
anshumankomawar, line 143: Noticed this in @mattcember's initial impl. Do we want to create a rivulet dataset here? Also is this dataset maintained somewhere, or how does the user interact with it. I'm missing some context with how the catalog interface is responsible for managing the rivulet dataset instance.

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -101,17 +196,38 @@ def list_tables(
 ) -> ListResult[TableDefinition]:
     """List a page of table definitions. Raises an error if the given namespace
     does not exist."""
-    raise NotImplementedError()
+    catalog = kwargs.get("catalog")
+    if not isinstance(catalog, CatalogProperties):
+        raise ValueError("Catalog must be a CatalogProperties instance")
+
+    namespace = namespace or default_namespace()
+    storage_impl.list_tables(namespace=namespace, catalog=catalog)
```
anshumankomawar, line 204: Need to add a helper method to go from Tables to TableDefinitions.

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -101,17 +196,38 @@ def list_tables(
...
+        raise ValueError("Catalog must be a CatalogProperties instance")
+
+    namespace = namespace or default_namespace()
+    table = storage_impl.get_table(
+        table_name=table, namespace=namespace, catalog=catalog
+    )
+
+    table_version = storage_impl.get_table_version(namespace, table, table.latest_table_version, catalog=catalog)
+
+    return TableDefinition.of(
```
anshumankomawar, line 225: Need to implement get_stream used placeholder for now. Is this what constructing a TableDefinition should look like?

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -190,13 +321,34 @@ def alter_namespace(
...
     optionally purging them."""
-    raise NotImplementedError("drop_namespace not implemented")
+    catalog = kwargs.get("catalog")
+    if not isinstance(catalog, CatalogProperties):
+        raise ValueError("Catalog must be a CatalogProperties instance")
+    
+    if not namespace_exists(namespace, catalog=catalog):
+        raise ValueError(f"Namespace {namespace} does not exist")
+    
+    storage_impl.delete_namespace(
```
anshumankomawar, line 347: will update to drop_namespace

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -57,11 +61,31 @@ def alter_table(
...
+    catalog = kwargs.get("catalog")
+    if not isinstance(catalog, CatalogProperties):
+        raise ValueError("Catalog must be a CatalogProperties instance")
+
+    namespace = namespace or default_namespace()
+
+    if not table_exists(table, namespace, catalog=catalog):
+        raise ValueError(f"Table {namespace}.{table} does not exist")
+
+    storage_impl.update_table(
```
anshumankomawar, line 73: Doesn't work as intended. When I update table do I also need to create a new table version? Currently a get_table call after update_table results in a None table version.

deltacat/storage/main/impl.py
```diff
@@ -1302,6 +1303,80 @@ def delete_stream(
...
+        *args,
+        namespace=namespace,
+        table_name=table_name,
+        **kwargs,
+    )
+
+    if not table:
+        raise ValueError(f"Table `{namespace}.{table_name}` does not exist.")
+
+    transaction = Transaction.of(
```
anshumankomawar, line 1330: Does deleting table cascade the delete? Should I be deleting underlying metafiles as well.

deltacat/catalog/v2/catalog_impl.py
pdames: There is no update_table with this signature - I think you're more looking to `update_table_version` (and calls from this catalog to `create_table` should also call into `create_table_version` at the storage level.

deltacat/catalog/v2/catalog_impl.py
pdames: I wouldn't think we'd need the Rivulet dataset to be created at this point, since this is just an empty table version (this seems like something that's more relevant as part of a `write_to_table` implementation).

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -101,17 +196,67 @@ def list_tables(
...
+
+    catalog = kwargs.get("catalog")
+    if not isinstance(catalog, CatalogProperties):
+        raise ValueError("Catalog must be a CatalogProperties instance")
+
+    namespace = namespace or default_namespace()
+    return _get_table_definition(table, namespace, *args, **kwargs)
+
+
+def _get_table_definition(
```
anshumankomawar, line 228: Is this what constructing a table definition looks like? Also should we raise errors if table version/stream does not exist.

deltacat/storage/main/impl.py
pdames: Don't worry about deleting the underlying metafiles for now (they should all become inaccessible once the table is deleted... this probably becomes more of a relevant question for when we get into garbage collector behavior, but a garbage collector should also be smart enough to realize that all metadata under a deleted table is implicitly deleted, whether it's been marked as deleted or not)

deltacat/catalog/v2/catalog_impl.py
pdames: Yes - this looks about right. IMO, we should raise errors if no table version or default DeltaCAT stream exists (since these are also required to be created alongside any table).

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -80,15 +104,86 @@ def create_table(
...
+    dataset = Dataset(
+        dataset_name=name,
+        metadata_uri=catalog.root,
+        namespace=namespace,
+        filesystem=catalog.filesystem,
+    )
+
+    # Add schema if provided
+    if schema and schema.arrow:
+        from deltacat.storage.rivulet import Schema as RivuletSchema
```
mcember, line 153: We should be deprecating RivuletSchema in favor of using deltacat schema

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -101,17 +196,67 @@ def list_tables(
...
+    table_definitions = [
+        _get_table_definition(table.table_name, namespace, *args, **kwargs)
+        for table in tables.all_items()
+    ]
+    
+    return ListResult(items=table_definitions)
 
 
 def get_table(
     table: str, namespace: Optional[str] = None, *args, **kwargs
```
mcember, line 214: Should we also optionally accept a version?

Should namespace here go after *args? It feels weird to have positional arguments in the order [table, namespace], since people will think of it as the hierarchy from left to right [namespace, table]. At least in create table I thought to demote it from positional argument into keyword only argument for this reason

deltacat/tests/catalog/main/test_catalog_impl_table_operations.py
```diff
@@ -0,0 +1,418 @@
...
+        """Setup for each test - create a test namespace"""
+        self.test_namespace = "test_table_namespace"
+        if not catalog.namespace_exists(self.test_namespace, catalog=self.catalog_properties):
+            catalog.create_namespace(
+                namespace=self.test_namespace, 
+                properties={"description": "Test Table Namespace"}, 
+                catalog=self.catalog_properties
+            )
+
+    def create_sample_arrow_schema(self):
```
mcember, line 38: any reason to not use fixtures here?

deltacat/tests/catalog/main/test_catalog_impl_table_operations.py
```diff
@@ -0,0 +1,418 @@
...
+        with pytest.raises(ValueError) as excinfo:
+            catalog.create_table(
+                name=table_name,
+                namespace=self.test_namespace,
+                description="Second creation attempt",
+                catalog=self.catalog_properties
+            )
+        
+        # Verify the error message
+        assert f"Table {self.test_namespace}.{table_name} already exists" in str(excinfo.value)
```
mcember, line 129: I thought you could put the expected error string in the `with pytest.raises` line?

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -80,15 +104,86 @@ def create_table(
...
+    if table_exists(name, namespace, **kwargs):
+        if fail_if_exists:
+            raise ValueError(f"Table {namespace}.{name} already exists")
+        return get_table(name, namespace, catalog=catalog)
+
+    # Create namespace if it doesn't exist
+    if not namespace_exists(namespace, catalog=catalog):
+        create_namespace(
+            namespace=namespace, properties=namespace_properties, catalog=catalog
+        )
```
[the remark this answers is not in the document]
mcember: +1. Also - are we expecting the user to pass in namespace properties just so that if we auto-create it automatically, we can create the default namespace with namespace properties?

I would consider removing this on the grounds to keep the signature simple - if they want custom namespace properties they can create the namespace themselves

deltacat/storage/main/impl.py
```diff
@@ -1302,6 +1303,80 @@ def delete_stream(
         filesystem=catalog_properties.filesystem,
     )
 
+def delete_table(
+    namespace: str,
+    table_name: str,
+    purge: bool = False,
```
mcember, line 1309: Do we want to expose a purge parameter if we don't support it yet? It could be confusing to end users (they may see the parameter but not have read our pydoc saying not implemented). If we do have it in the signature, we should to raise a not implemented exception.

deltacat/catalog/v2/catalog_impl.py
anshumankomawar: That's fair, will move it to after *args. If you already have table and table_version wouldn't the user be expected to call get_stream?

deltacat/catalog/v2/catalog_impl.py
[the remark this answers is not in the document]
pdames: I think I agree to keep the focus of `create_table` on the table creation itself, but I don't care too much either way about keeping/removing namespace properties here, since the important thing in my mind is keeping _required_ argument count low (I'm OK with having a large number of optional args, since they're hidden from the user by default - so we definitely shouldn't ever require the specification of `namespace_properties` as part of `create_table`).

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -101,17 +196,67 @@ def list_tables(
...
+        raise ValueError("Catalog must be a CatalogProperties instance")
+
+    namespace = namespace or default_namespace()
+    table = storage_impl.get_table(
+        table_name=table_name, namespace=namespace, catalog=catalog
+    )
+
+    table_version = storage_impl.get_table_version(namespace, table_name, table.latest_table_version, catalog=catalog)
+
+    stream = storage_impl.get_stream(
```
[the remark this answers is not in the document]
pdames: Yeah, for example, I'm imagining they'll explicitly need to ask for the `iceberg` stream if they want to query the Iceberg variant, so it seems like we'll need a way for them to explicitly pass in at least `StreamFormat`.

deltacat/tests/catalog/main/test_catalog_impl_table_operations.py
pdames: Also, we have a lot of better custom error types in `exceptions.py` that we should start making better use of (and adding new exception types as needed) to avoid any cases that otherwise require string inspection like this (we probably need a corresponding high-level issue to comb through all of the high-level base Python errors we're throwing in the native DeltaCAT catalog/storage impl, and turn them into more appropriately typed DeltaCAT exceptions).

deltacat/catalog/v2/catalog_impl.py
anshumankomawar: Your right. Updating this to call update_table and update_table_version with the correct properties.

deltacat/storage/main/impl.py
anshumankomawar: I'll throw an error for now so that they're aware it isn't implemented yet.

deltacat/storage/main/impl.py
```diff
@@ -1302,6 +1303,80 @@ def delete_stream(
         filesystem=catalog_properties.filesystem,
     )
 
+def delete_table(
+    namespace: str,
```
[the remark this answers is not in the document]
anshumankomawar: Will do. Hadn't looked too closely for this initial review was just hoping to get high level questions answered.

deltacat/catalog/v2/catalog_impl.py
anshumankomawar: Agreed can continue this task in a future PR. Will remove and rivulet dataset code in create_table for now.

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -80,15 +104,86 @@ def create_table(
...
+    if not table_exists(table, namespace, catalog=catalog):
+        raise ValueError(f"Table {namespace}.{table} does not exist")
+
+    # Call storage implementation to drop the table
+    storage_impl.delete_table(
+        table_name=table, 
+        namespace=namespace, 
+        purge=purge, 
+        catalog=catalog
+    )
```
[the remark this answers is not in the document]
anshumankomawar: @mattcember  suggested something similar as well. I think it makes sense to add an optional table_version argument and a delete_table_version as well.

Will add the TODO

deltacat/catalog/v2/catalog_impl.py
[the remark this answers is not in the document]
anshumankomawar: I'm going to leave this out for my PR but will add a TODO to address this.

deltacat/storage/main/impl.py
```diff
@@ -1302,6 +1303,80 @@ def delete_stream(
...
+    """
+    table: Optional[Table] = get_table(
+        *args,
+        namespace=namespace,
+        table_name=table_name,
+        **kwargs,
+    )
+
+    if not table:
+        raise ValueError(f"Table `{namespace}.{table_name}` does not exist.")
```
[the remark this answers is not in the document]
anshumankomawar: Looks like we don't throw an error message in the storage layer. We create a new table version using the existing table_version if an explicit table version isn't provided. 
```
        if not table_version:
            # generate the next table version ID
            table_version = TableVersion.next_version(prev_table_version)
        else:
            # ensure that the given table version number matches expectations
            expected_table_version = TableVersion.next_version(prev_table_version)
            _, version_number = TableVersion.parse_table_version(
                table_version,
            )
            _, expected_version_number = TableVersion.parse_table_version(
                expected_table_version,
            )
            if version_number != expected_version_number:
                raise ValueError(
                    f"Expected to create table version "
                    f"{expected_version_number} but found {version_number}.",
                )

```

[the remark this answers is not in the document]
anshumankomawar: Going to leave the extra call in for now but as we discussed maybe we can consider adding optional table version numbers to the catalog APIs.

deltacat/storage/main/impl.py
```diff
@@ -1302,6 +1303,80 @@ def delete_stream(
...
+    given namespace does not exist.
+    """
+    namespace: Optional[Namespace] = get_namespace(
+        *args,
+        namespace=namespace,
+        **kwargs,
+    )
+
+    if not namespace:
+        raise ValueError(f"Namespace `{namespace}` does not exist.")
```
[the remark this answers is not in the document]
anshumankomawar: Looks like the transaction expects a metafile so we would need to do a get_namespace using the id to get the metafile.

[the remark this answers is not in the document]
pdames: Hmm... can we just create the namespace to delete, since it already contains the unique locator the transaction should need to perform the delete?


```python
namespace = Namespace.of(NamespaceLocator.of("namespace"))
```

deltacat/catalog/v2/catalog_impl.py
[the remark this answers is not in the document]
pdames: Do we need a separate `delete_table_version` API or should we just infer behavior based on whether or not an optional table version has been specified in the existing `delete_table` API?

deltacat/tests/catalog/main/test_catalog_impl_table_operations.py
anshumankomawar: I've updated the ones directly included in my changes but there are still some errors thrown in the storage level that need to be updated. Don't want to dilute this cr will track a separate ticket for it.

deltacat/catalog/v2/catalog_impl.py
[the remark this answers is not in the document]
anshumankomawar: Will update the docs to note that the namespace will be created for you if it doesn't already exist.

deltacat/storage/main/impl.py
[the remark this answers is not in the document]
anshumankomawar: Missed this comment, going to see if I can get it to work.

deltacat/storage/main/impl.py
```diff
@@ -1303,6 +1305,84 @@ def delete_stream(
...
+    """
+    table: Optional[Table] = get_table(
+        *args,
+        namespace=namespace,
+        table_name=name,
+        **kwargs,
+    )
+
+    if not table:
+        raise TableNotFoundError(f"Table `{namespace}.{name}` does not exist.")
```
pdames, line 1330: Similar to `delete_namespace` below, can we remove this `get_table` call and just catch/reraise a more clear ValueError from the commit itself?

pdames (approved):
LGTM! Remaining comments are relatively low-priority, but would be great if they can also be addressed prior to merge (or, if not, feel free to submit a small fast-follow PR). Also, let's just ensure that all failing approval workflows have succeeded prior to merge.

PR #503, opened 2025-03-11
Prototype of write interfaces / data access layer

mcember:
## Summary

Prototype of low level ("storage") write interface, and implementations for iceberg and deltacat 

This PR only includes the low level ("storage") write interface, which Daft or other query engines can integrate with, not the higher level ("catalog") user-facing interface. 

Subsequent PRs will prototype Daft integration and flesh out higher level user-facing interfaces. See README in data access layer module for more details 

## Changes
* Fix minor bugs in rivulet and deltacat schema class
* Introduce dataset writer interface and implement for rivulet and iceberg 

## Testing
Unit tests


pdames:
Closing stale PR.

PR #504, opened 2025-03-12
feat: implement catalog table functions

anshumankomawar:
## Summary

This PR continues work on the catalog interface layer for Deltacat. It primarily focuses on implementing table related functions.

Functions Implemented:
- alter_table
- create_table
- drop_table
- list_tables
- get_table
- rename_table

Also includes minor bugfixes and better exception handling in the storage interface layer.

## Testing

- ran existing tests using `make test`
- added new tests for catalog table functions

## Checklist

## Additional Notes

There are some follow up tasks that have been added as TODO's.

### DIFF
--- a/README-development.md
+++ b/README-development.md
@@ -82,12 +82,12 @@ make test
 Runs all unit tests.
 
 > [!NOTE]
-> To run an individual unit test where `my_deltacat_test` exists in either the test file, class, 
+> To run an individual unit test where `my_deltacat_test` exists in either the test file, class,
 > or function/method name, run a command of the form:
 > ```shell
 > python -m pytest -k "my_deltacat_test" -s -vv
 > ```
-> Note that the `-s` flag disables output capturing so that you can see stdout from `print` 
+> Note that the `-s` flag disables output capturing so that you can see stdout from `print`
 > and other statements in real time, and `-vv` let's you see each test version and its input
 > parameters to ease debugging.
 
--- a/deltacat/catalog/v2/catalog_impl.py
+++ b/deltacat/catalog/v2/catalog_impl.py
@@ -1,21 +1,25 @@
 from typing import Any, Dict, List, Optional, Union
 
-from deltacat.catalog.catalog_properties import (
-    CatalogProperties,
+from deltacat.exceptions import (
+    NamespaceAlreadyExistsError,
+    StreamNotFoundError,
+    TableAlreadyExistsError,
+    TableVersionNotFoundError,
 )
-
 from deltacat.storage.model.partition import PartitionScheme
 from deltacat.catalog.model.table_definition import TableDefinition
 from deltacat.storage.model.sort_key import SortScheme
 from deltacat.storage.model.list_result import ListResult
 from deltacat.storage.model.namespace import Namespace, NamespaceProperties
 from deltacat.storage.model.schema import Schema
-from deltacat.storage.model.table import TableProperties
+from deltacat.storage.model.table import Table, TableProperties
+from deltacat.storage.model.table_version import TableVersion
 from deltacat.storage.model.types import (
     DistributedDataset,
     LifecycleState,
     LocalDataset,
     LocalTable,
+    StreamFormat,
 )
 from deltacat.types.media import ContentType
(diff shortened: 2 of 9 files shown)


README-development.md
```diff
@@ -82,12 +82,12 @@ make test
 Runs all unit tests.
 
 > [!NOTE]
-> To run an individual unit test where `my_deltacat_test` exists in either the test file, class, 
+> To run an individual unit test where `my_deltacat_test` exists in either the test file, class,
```
anshumankomawar, line 85: linter errors from previous commit.

deltacat/storage/main/impl.py
```diff
@@ -899,7 +901,7 @@ def update_table(
         **kwargs,
     )
     if not old_table:
-        raise ValueError(f"Table `{namespace}.{table_name}` does not exist.")
+        raise TableNotFoundError(f"Table `{namespace}.{table_name}` does not exist.")
```
anshumankomawar, line 904: As pointed out by @pdames update the rest of the storage implementation to use Deltacat exceptions instead of throwing ValueErrors.

deltacat/storage/main/impl.py
```diff
@@ -1315,6 +1317,84 @@ def delete_stream(
...
+    **kwargs,
+) -> None:
+    """
+    Drops the given table and all its contents (table versions, streams, partitions,
+    and deltas). If purge is True, also removes all data files associated with the table.
+    Raises an error if the given table does not exist.
+
+    TODO: Honor purge once garbage collection is implemented.
+    """
+    table: Optional[Table] = get_table(
```
anshumankomawar, line 1334: As mentioned by @pdames
> Can we remove this get_namespace call and just catch/reraise a more clear ValueError from the commit itself?

anshumankomawar: Will try to create an in memory Table/TableLocator instead of calling get_table.

deltacat/storage/main/impl.py
```diff
@@ -1315,6 +1317,84 @@ def delete_stream(
...
+    namespace: str,
+    purge: bool = False,
+    *args,
+    **kwargs,
+) -> None:
+    """
+    Drops the given table namespace and all its contents. Raises an error if the
+    given namespace does not exist.
+    """
+    namespace: Optional[Namespace] = get_namespace(
```
anshumankomawar, line 1373: Same as previous `get_table` call. Try to use an in memory solution.

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -76,129 +143,360 @@ def create_table(
...
+
+    return ListResult(items=table_definitions)
 
 
 def get_table(
-    table: str, namespace: Optional[str] = None, *args, **kwargs
+    name: str,
+    *args,
+    namespace: Optional[str] = None,
+    table_version: Optional[str] = None,
```
anshumankomawar, line 280: @mcember 
>Should we also optionally accept a version?

> Should namespace here go after *args? It feels weird to have positional arguments in the order [table, namespace], since people will think of it as the hierarchy from left to right [namespace, table]. At least in create table I thought to demote it from positional argument into keyword only argument for this reason

anshumankomawar: Added an optional TableVersion parameter. Also decided to ensure consistency with param ordering. 
```
positional params,
*args,
optional params,
**kwargs
```

deltacat/catalog/v2/catalog_impl.py
```diff
@@ -76,129 +143,360 @@ def create_table(
...
+    stream_format: StreamFormat = StreamFormat.DELTACAT,
```
anshumankomawar, line 281: As suggested by @mattcember 
> In the future when we have multiple streams, how are users going to specify which stream they want in get / list / etc?

anshumankomawar: Added an explicit stream_format parameter to specify the stream. It defaults to the Deltacat stream.

pdames (approved):

PR #507, opened 2025-03-14
Add enforce_primary_key_uniqueness support to converter

Zyiqin-Miranda:
## Summary

Records with same primary key/identify fields will be deduped with this PR, the purpose is that we can enforce primary key uniqueness as well as finding primary keys to be deleted based on equality delete files within same operation(No need to download into memory same files again). All records meant to be either deduped or deleted will be output in form of position deletes.

## Rationale

## Changes

## Impact

## Testing

## Regression Risk

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/compute/converter/model/converter_session_params.py
+++ b/deltacat/compute/converter/model/converter_session_params.py
@@ -23,6 +23,9 @@ class ConverterSessionParams(dict):
         ), "iceberg_namespace is a required arg"
         result = ConverterSessionParams(params)
 
+        result.enforce_primary_key_uniqueness = params.get(
+            "enforce_primary_key_uniqueness", False
+        )
         result.compact_small_files = params.get("compact_small_files", False)
 
         # For Iceberg v3 spec, option to produce delete vector that can establish 1:1 mapping with data files.
@@ -50,6 +53,14 @@ class ConverterSessionParams(dict):
     @property
     def iceberg_namespace(self) -> str:
         return self["iceberg_namespace"]
+
+    @property
+    def enforce_primary_key_uniqueness(self) -> bool:
+        return self["enforce_primary_key_uniqueness"]
+
+    @enforce_primary_key_uniqueness.setter
+    def enforce_primary_key_uniqueness(self, enforce_primary_key_uniqueness) -> None:
+        self["compact_small_files"] = enforce_primary_key_uniqueness
 
     @property
     def compact_small_files(self) -> bool:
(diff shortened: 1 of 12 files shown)


deltacat/compute/converter/steps/convert.py
```diff
@@ -71,43 +115,34 @@ def filter_rows_to_be_deleted(
             data_file_table[identifier_column],
             equality_delete_table[identifier_column],
         )
-        positional_delete_table = data_file_table.filter(equality_deletes)
-        logger.info(f"positional_delete_table:{positional_delete_table.to_pydict()}")
+        position_delete_table = data_file_table.filter(equality_deletes)
+        logger.info(f"positional_delete_table:{position_delete_table.to_pydict()}")
```
pdames, line 119: This seems like a pretty expensive operation for a logger message... can we either (1) replace this with something cheaper (like just a length of the position delete table) or (2) only execute lines like this if some debug/dev/trace flag is enabled?

deltacat/compute/converter/utils/converter_session_utils.py
```diff
@@ -56,7 +60,52 @@ def construct_iceberg_table_prefix(
...
-    return partition_value_str
+def group_all_files_to_each_bucket(
+    data_file_dict, equality_delete_dict, pos_delete_dict
+):
+    convert_input_files_for_all_buckets = []
+    files_for_each_bucket_for_deletes = defaultdict(tuple)
+    if equality_delete_dict:
+        # files_for_each_bucket contains the following files list:
+        # {partition_value: [(equality_delete_files_tuple_list, data_files_tuple_list, pos_delete_files_tuple_list)]
+        for k, v in data_file_dict.items():
```
pdames, line 71: I think we also have some debug logs used for testing to clean up / remove here?

deltacat/compute/converter/model/converter_session_params.py
```diff
@@ -51,6 +54,14 @@ def iceberg_warehouse_bucket_name(self) -> str:
     def iceberg_namespace(self) -> str:
         return self["iceberg_namespace"]
 
+    @property
+    def enforce_primary_key_uniqueness(self) -> bool:
+        return self["enforce_primary_key_uniqueness"]
+
+    @enforce_primary_key_uniqueness.setter
+    def enforce_primary_key_uniqueness(self, enforce_primary_key_uniqueness) -> None:
+        self["compact_small_files"] = enforce_primary_key_uniqueness
```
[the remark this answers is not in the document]
Zyiqin-Miranda: Good catch, typo here

deltacat/compute/converter/steps/convert.py
Zyiqin-Miranda: Yeah, these are actually debug logs, will change to table length to log out how many records we dropped in this step in follow-up PR

Zyiqin-Miranda:
Since I'm quite close to publish another PR with additional code clean-ups, will merge this PR first and address the comments in that PR

PR #508, opened 2025-03-17
Fallback janitor job two

SashankBalusu:
Secondary Janitor Job to Handle Timed-Out and Failed Transactions

## Description:
This PR enhances the janitor job by implementing the handling of timed-out and failed transactions. It adds logic to:
- Move transactions that exceed the timeout threshold from running/ to failed/.
- Clean up associated metafiles and locator files for failed transactions.
- Perform a brute-force search to remove lingering metafiles tied to failed transactions.
- These changes improve the reliability of the janitor cleanup process.

### Janitor_delete_timed_out_transaction:
- Goes through the in progress transaction folder and checks to see if the heartbeat timeout (stored in the filename) has been reached for any files
- If the heartbeat timeout has been reached, the files are moved to the failed directory with TIMEOUT_TXN constant appended to the file name
- Uses brute_force_search_matching_metafiles to delete orphaned metafiles.
### Janitor_remove_files_in_failed:
- Reads failed transactions and removes their associated metafiles.
- Renames cleaned-up failed transactions to append a SUCCESSFULLY_CLEANED tag. This ensures that when iterating through the failed transaction directory we don't attempt to delete already cleaned failed transactions. 
- While iterating through the failed transaction directory to check for unsuccessful cleanings we also look for  a transaction that was in progress that gets marked as failed because their heartbeat timeout passed but the file was unsuccessfully cleaned
- For these files, we simply call brute_force_search_matching_metafiles like usual. 
### Brute_force_search_matching_metafiles:
- Recursively searches directories to delete any metafiles matching failed transaction id
### Added Transaction Status Tags:
- Introduced constants:
- OPERATION_TIMEOUTS = {
        "create": 5,
        "update": 3,
        "delete": 4,
        "read_siblings": 2,
        "read_children": 2,
        "read_latest": 3,
        "read_exists": 1,
    }
   - sets constant limits for how long each operation is expected to take
- SUCCESSFULLY_CLEANED = "cleaned"
- CURRENTLY_CLEANING = "working"
- TIMEOUT_TXN = “timedout”
- Appended these tags to transaction filenames to track their status:
   - Timed out: f"{id}{TXN_PART_SEPARATOR}{TIMEOUT_TXN}"   
        - Identifies case where transaction times out and is caught by heartbeat failed
   - Failed but associated files not cleaned:f"{id}{TXN_PART_SEPARATOR}{CURRENTLY_CLEANING}"   
        - Identifies case where process crashes causing transaction to be moved to failed but not properly deleted 
   - Deleted: f"{id}{TXN_PART_SEPARATOR}{SUCCESSFULLY_CLEANED}"
        - Identifies when a failed transaction is successfully deleted   
***This allows us to identify failed transactions without having to read the transaction file***


Zyiqin-Miranda (commented):
Thanks for putting together the implementation for the fallback janitor! Looking forward to see how/where this janitor is called and integrated with existing failed transaction cleanup logic inside Transaction.commit() functions. 

One high-level comment is the placement of the janitor, whether it should be part of the deltacat/storage/model/transaction, since it's a cleaning up job for transaction specifically instead of a general compute job? Curious to learn your opinion here.

pdames:
> Thanks for putting together the implementation for the fallback janitor! Looking forward to see how/where this janitor is called and integrated with existing failed transaction cleanup logic inside Transaction.commit() functions.
> 
> One high-level comment is the placement of the janitor, whether it should be part of the deltacat/storage/model/transaction, since it's a cleaning up job for transaction specifically instead of a general compute job? Curious to learn your opinion here.

I _think_ compute should be the right place for it, since it should hold all "stored-procedure-like-jobs" that are expected to be periodically invoked by an external orchestrator (i.e., the simplest orchestrator may just be a daily cron job run against a local catalog instance). Over time, I expect the janitor job to also evolve to distribute its work using Ray/Daft/etc. to more efficiently cleanup potentially vast, exabyte-scale, distributed catalogs stored in S3 or other cloud storage services.

Zyiqin-Miranda:
> 
> I _think_ `compute` should be the right module for it, since I expect the `compute` module to hold all DeltaCAT "stored-procedure-like-jobs" that we expect to periodically invoke w/ an external orchestrator (i.e., manually invoked by the end-user, automatically invoked by daily cron jobs for local catalogs, a distributed AWS Lambda invoked by a Step Function workflow, etc.). Over time, I expect the janitor job to also evolve to distribute its work using Ray/Daft/etc. to more efficiently cleanup potentially vast, exabyte-scale, distributed catalogs stored in S3 or other cloud storage services.

Thanks for the context, putting under `compute` makes sense with janitor invoked as scheduled jobs/distributed jobs.

deltacat/storage/README.md
```diff
@@ -78,7 +78,7 @@ Note that (except **Immutable ID**) this is the same format used by **Metadata R
 
 ### Transaction Log Directory
 The **Transaction Log Directory** (`${CATALOG_ROOT}/txn`) is a special directory in the **Catalog Root Directory** which holds all successfully committed transactions. It contains one **Transaction Log File** per successful transaction recording transaction details.
-```
+``
```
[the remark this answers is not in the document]
025rhu: Looks like it! Fixed.

deltacat/constants.py
```diff
@@ -102,3 +102,19 @@
...
 DEFAULT_PARTITION_ID = "partition"
 DEFAULT_PARTITION_VALUES = ["default"]
+
+# Transaction Status constants
+SUCCESSFULLY_CLEANED = "cleaned"
+CURRENTLY_CLEANING = "working"
+TIMEOUT_TXN = "timedout"
+
+#operation timeout constants
+OPERATION_TIMEOUTS = {
```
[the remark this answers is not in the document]
025rhu: These are fixed values that we decided to use. We picked arbitrarily large values per Patrick's advice, but @SashankBalusu is currently working on something to get more accurate values, that he could give more information on in a bit.

It is difficult to determine a timeout time based on the size of the data, however, so these values will be constant across all transaction operations once they are finalized.

deltacat/storage/model/transaction.py
```diff
@@ -577,11 +596,20 @@ def _commit_write(
...
-        )
+        #TODO: move this dict into a different more suiting place
+        
+
+        total_time_for_transaction = 0
+        for operation in self.operations:
+            total_time_for_transaction += OPERATION_TIMEOUTS.get(operation.type, 0)
+
+        start_time = float(self.id.split(TXN_PART_SEPARATOR)[0])
+        final_time_heartbeat = start_time + (total_time_for_transaction * 1_000_000_000)  # Convert seconds to nanoseconds
```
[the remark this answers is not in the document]
025rhu: Makes sense, done!

deltacat/constants.py
```diff
@@ -102,3 +102,19 @@
 DEFAULT_STREAM_ID = "stream"
 DEFAULT_PARTITION_ID = "partition"
 DEFAULT_PARTITION_VALUES = ["default"]
+
+# Transaction Status constants
+SUCCESSFULLY_CLEANED = "cleaned"
+CURRENTLY_CLEANING = "working"
```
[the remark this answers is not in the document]
025rhu: That makes sense, changed to "cleaning."

deltacat/storage/model/transaction.py
```diff
@@ -591,6 +619,7 @@ def _commit_write(
         locator_write_paths = []
         try:
             for operation in self.operations:
+                total_time_for_transaction += OPERATION_TIMEOUTS[operation.type]
```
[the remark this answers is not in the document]
025rhu: Yes, it is not used anywhere downstream. Since we have the for loop above on line 603 and 604 summing the heartbeat timeout, we will remove this one.

deltacat/compute/janitor.py
```diff
@@ -0,0 +1,174 @@
...
+                for dirty_file in dirty_files_names:
+                    parts = dirty_file.split(TXN_PART_SEPARATOR)
+                    if len(parts) < 2:
+                        continue
+                    transaction_id = parts[1]
+                    # Look for transaction_id in the filename
+                    if transaction_id in base_name:
+                        try:
+                            filesystem.delete_file(entry.path)
+                            print(f"Deleted file: {entry.path}")
```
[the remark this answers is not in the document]
025rhu: Yes, thanks for catching that. We've changed all print statements in this file to loggers, setting levels as we saw fit, but please double check us on that!

deltacat/tests/compute/test_janitor.py
```diff
@@ -0,0 +1,185 @@
...
+        os.makedirs(running_txn_dir, exist_ok=True)
+        os.makedirs(failed_txn_dir, exist_ok=True)
+
+        # Create multiple timed-out transaction logs
+        txn_ids = ["txn_one", "txn_two", "txn_three"]
+        start_time = time.time_ns() - 1_000_000_000  # 1 second in the past
+
+        txn_filenames = []
+        for txn_id in txn_ids:
+            txn_filename = f"{start_time}{txn_id}{TXN_PART_SEPARATOR}{TXN_PART_SEPARATOR}{time.time_ns()}"
```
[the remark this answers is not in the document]
025rhu: Yes! We fixed it, and the tests are still passing.

deltacat/storage/model/transaction.py
```diff
@@ -410,6 +415,20 @@ def end_time(self) -> Optional[int]:
         Returns the end time of the transaction.
         """
         return self.get("end_time")
+    
+    def _mark_status(self, status: str) -> None:
```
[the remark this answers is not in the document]
025rhu: This function was used in a previous implementation of the janitor job, but was not removed when we changed it. We have deleted it now as it is no longer in use. All tests still pass!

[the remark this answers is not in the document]
pdames: Did you mean to delete the `_mark_start_time` method or just create a new `_mark_status` method? The `_mark_start_time` method is still used by transaction `commit`, so it seems like the current code will be broken until we add it back.

deltacat/storage/model/transaction.py
```diff
@@ -577,11 +583,21 @@ def _commit_write(
...
-            running_txn_log_dir,
-            self.id,
-        )
+        # TODO: move this dict into a different more suiting place
+
+        total_time_for_transaction = 0
+        for operation in self.operations:
+            total_time_for_transaction += OPERATION_TIMEOUTS.get(operation.type, 0)
+
+        start_time = float(self.id.split(TXN_PART_SEPARATOR)[0])
```
pdames, line 592: Should this just be a call to `self.start_time()`?

deltacat/storage/model/transaction.py
```diff
@@ -577,11 +583,21 @@ def _commit_write(
...
+        final_time_heartbeat = start_time + (
+            total_time_for_transaction * NANOS_PER_SEC
+        )  # Convert seconds to nanoseconds
```
pdames, line 595: Please add all logic for converting the heartbeat timeout time to a `timeout_time` method in `TransactionSystemTimeProvider`, since this method of calculation is only relevant for transaction time providers that use system clock nanos.

deltacat/constants.py
```diff
@@ -102,3 +102,19 @@
 DEFAULT_STREAM_ID = "stream"
 DEFAULT_PARTITION_ID = "partition"
 DEFAULT_PARTITION_VALUES = ["default"]
+
+# Transaction Status constants
+SUCCESSFULLY_CLEANED = "cleaned"
+CURRENTLY_CLEANING = "cleaning"
+TIMEOUT_TXN = "timedout"
```
pdames, line 109: Can we move these to a new `class TransactionState(str, Enum)` type in `types.py`, and change the supported values to simply:
`FAILED` (equivalent to `CLEANING`), `PURGED` (equivalent to `CLEANED`), `TIMEOUT`, `RUNNING`, and `SUCCESS`?

deltacat/storage/model/transaction.py
```diff
@@ -610,10 +626,9 @@ def _commit_write(
                     )
         except Exception:
             # write a failed transaction log file entry
-            failed_txn_log_file_path = posixpath.join(
-                failed_txn_log_dir,
-                self.id,
-            )
+            path_ending = f"{self.id}{TXN_PART_SEPARATOR}{CURRENTLY_CLEANING}"
```
pdames, line 629: Instead of attaching the "CLEANING" and "CLEANED" states to the transaction file directly (which requires subsequent file movements/renames to apply), what do you think about just writing a `transaction.state()` helper method that retrieves the current transaction state? 

We can change the `CLEANING` state to `FAILED`, and infer it as the transaction ID being present in both the failed and running directories, then change `CLEANED` to `PURGED` and infer it as the transaction ID ONLY being present in the failed directory. `SUCCESS` would be inferred as the transaction ID being present in the success directory, and `RUNNING` would be inferred as the transaction ID being present in the running directory (unless it is also present in the failed directory or, in other words, we should have an if/else check that checks if the directory is in the failed dir first, and otherwise falls back to see if it's running).

pdames (changes requested):

deltacat/storage/model/transaction.py
andrewtabbert: Yes, just made this change

deltacat/storage/model/transaction.py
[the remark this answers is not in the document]
025rhu: Oops, I did not mean to delete that, I meant to delete _mark_status instead, since it is not used anywhere else in the code. 

I've fixed this change!

deltacat/storage/model/transaction.py
andrewtabbert: Hey Patrick, I’ve gone ahead and implemented it. I created a transaction.state() helper method that determines the current state based on the presence of the transaction ID in the appropriate directories.

deltacat/constants.py
andrewtabbert: Hi Patrick, I added this to types.py

deltacat/storage/model/transaction.py
andrewtabbert: Just made this edit.

deltacat/storage/model/transaction.py
```diff
@@ -639,12 +706,19 @@ def _commit_write(
             # delete the in-progress transaction log file entry
             filesystem.delete_file(running_txn_log_file_path)
             # failed transaction cleanup is now complete
+            old_path = failed_txn_log_file_path
+            #new_path = posixpath.join(failed_txn_log_dir, f"{self.id}{TXN_PART_SEPARATOR}{SUCCESSFULLY_CLEANED}")
```
pdames, line 710:
```suggestion
```

deltacat/tests/compute/test_janitor.py
```diff
@@ -0,0 +1,241 @@
...
+            txn_path
+        ), "Transaction file still exists in running directory."
+
+        # Run the janitor job that should:
+        # 1. Move the running txn file to the failed directory with TIMEOUT_TXN appended.
+        # 2. Invoke brute force search which deletes the metafile and renames the txn log file to use SUCCESSFULLY_CLEANED.
+        janitor_delete_timed_out_transaction(temp_dir)
+
+        # Expected name: original txn_filename with TIMEOUT_TXN replaced by SUCCESSFULLY_CLEANED.
+       # new_txn_file_name = f"{txn_filename}{TXN_PART_SEPARATOR}{SUCCESSFULLY_CLEANED}"
```
pdames, line 73:
```suggestion
```

deltacat/tests/compute/test_janitor.py
```diff
@@ -0,0 +1,241 @@
...
+            )
+            with open(test_metafile_path, "w") as f:
+                f.write("mock metafile content")
+
+            txn_filenames.append((txn_filename, txn_path, test_metafile_path))
+
+        # Run the janitor function to move timed-out transactions to the failed directory
+        janitor_delete_timed_out_transaction(temp_dir)
+
+        # Verify that all transactions were moved to the failed directory with SUCCESSFULLY_CLEANED appended
```
pdames, line 160:
```suggestion
        # Verify that all transactions were moved to the failed directory
```

Issue #500, opened 2025-03-08
[Bug] Catalog re-initialization in every DeltaCAT Catalog API call


MingshiPeng:
## Issue
For each DeltaCAT Catalog API call (e.g. `deltacat.create_table`), the underlying catalog implementation will be re-initialized instead of re-used. 

This is especially troublesome for local dev with Iceberg in-memory db, since the in-memory db will be purged and re-initialized every time the catalog itself re-initialized.

## Code trace
1. for each Catalog API call, DeltaCAT will try to retrieve the underlying catalog implementation, [link](https://github.com/ray-project/deltacat/blob/ce66736526250719a88a4f8ab92d0bca35f879d9/deltacat/catalog/delegate.py#L118) 
2. a ray-remote method will be called to retrieve the underlying catalog implementation, [link](https://github.com/ray-project/deltacat/blob/ce66736526250719a88a4f8ab92d0bca35f879d9/deltacat/catalog/model/catalog.py#L127)
3. since the native_object (i.e. the catalog impl) is excluded from the `reduce` method, the Catalog will be re-initialized everytime a method is called. [link](https://github.com/ray-project/deltacat/blob/ce66736526250719a88a4f8ab92d0bca35f879d9/deltacat/catalog/model/catalog.py#L42-L45)

pdames:
We may want to break the retrieval apart into serializable and unserializable catalogs. Since the DeltaCAT native catalog should be serializable, it should be easy to avoid this step. However, since the Iceberg catalog is not, we will likely need to keep this implementation until we can deprecate it and rely purely on DeltaCAT -> Iceberg metadata synchronizers.

mcember:
Is there any way in Ray to cache the results of deserialization on the worker node? 

For unserializable catalogs - as long as the catalog can be re-constructed deterministically from input configuration, it isn't a correctness problem. For instance, you can construct an Iceberg glue or SQL catalog from input config deterministically across multiple worker nodes, but InMemoryCatalog will not work. There is a small performance hit in that every single catalog operation needs to re-construct the iceberg catalg

mcember:
I like Patrick's suggestion of splitting catalogs into serializable vs non-serializable so this re-construction path will only happen for non-serializable catalogs

PR #517, opened 2025-03-28
More catalog refactors

mcember:
Changes
0. Carry over changes from previous catalog refactor PR
1. Change name of "catalog" in delegate to "inner". The "catalog" name was super confusing, especially since the user is passing in "catalog" to delegate as the name of their catalog to look up from initialized catalogs
2. Added pattern for different Catalog impls to have different strongly typed config objects (see: IcebergCatalogConfig), and initializing them via factory methods in Catalog. Refactor CatalogProperties class
3. Fix some bugs in Catalog/Catalogs class. Allow `put_catalog` to accept either Catalog object or module
4. merge some bugs in iceberg/DC object translations from Mingshi's uncommitted read interface prototype
5. unit test catalog initialization
6. modify unit tests to use catalog initialization / delegate rather than call catalog impl directly
7. upgrade pyiceberg dependencies to 0.9 and add additional dev dependencies (for unit tests which require in memory catalog)


deltacat/tests/local_deltacat_storage/db_test.sqlite
mcember, line 1: Accidentally checked this in...will remove

deltacat/catalog/main/impl.py
```diff
@@ -34,10 +34,32 @@
 from deltacat.compute.merge_on_read import MERGE_FUNC_BY_DISTRIBUTED_DATASET_TYPE
 from deltacat import logs
 from deltacat.constants import DEFAULT_NAMESPACE
-from deltacat.storage.main import impl as storage_impl
+from deltacat.storage import metastore as storage_impl
 
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
+"""
+This is the default implementation for the Catalog interface, using DeltaCAT native storage
```
MingshiPeng, line 42: nit: should we rename the directory from `main` to something like `default`? The `main` is pretty confusing.

deltacat/catalog/model/catalog.py
```diff
@@ -2,21 +2,38 @@
...
+    def __init__(self, impl: ModuleType = DeltacatCatalog, *args, **kwargs):
+        """
+        Constructor for a Catalog.
+
+        The args and kwargs here will be plumbed through to the catalog initialize function, and the results
+        are stored in Catalog.inner. Any state which is required (like: metastore root URI, pyiceberg native catalog)
+        MUST be returned by initialize.
+
+        Note: all initialization configuration MUST be pickle-able. When `Catalog` is pickled, _inner is excluded
+          and instead we only pass impl/args/kwargs, which
```
MingshiPeng, line 35: nit: sentence ending abruptly

deltacat/catalog/model/catalog.py
```diff
@@ -136,46 +212,107 @@ def init(
...
+            f"Available catalogs: {available_catalogs}."
         )
     return catalog
 
 
-def put_catalog(name: str, *args, **kwargs) -> Catalog:
-    from deltacat.catalog.model.catalog import all_catalogs
+def put_catalog(
+    name: str,
+    *,
```
MingshiPeng, line 267: What's this wildcard? is this intended?

deltacat/storage/iceberg/model.py
```diff
@@ -502,10 +503,9 @@ def unmap(
                 final_fields.append(final_field)
             iceberg_schema = IcebergSchema(
                 fields=final_fields,
-                schema_id=obj.linked_schema_ids.get(stream_locator, INITIAL_SCHEMA_ID)
-                if obj.linked_schema_ids is not None
-                else INITIAL_SCHEMA_ID,
-                identifier_field_ids=obj.merge_keys,
```
MingshiPeng, line 508: this is a hack I did to make my POC work, I don't think this is a proper fix though, I'm also not sure what a proper fix should look like due to lack of context.

I'm fine keeping those so at least we have some unit test working comparing to simply breaking, but we should create an issue or TODO for tracking.

deltacat/catalog/main/impl.py
```diff
@@ -72,31 +94,26 @@ def read_table(
     stream_position_range_inclusive: Optional[Tuple[int, int]] = None,
     merge_on_read: Optional[bool] = False,
     reader_kwargs: Optional[Dict[Any, Any]] = None,
-    deltacat_storage_kwargs: Optional[Dict[Any, Any]] = None,
     **kwargs,
```
MingshiPeng, line 97: info: in what scenario are those 3 `kwargs` are used? and why are we removing the `deltacat_storage_kwargs`?

deltacat/tests/catalog/iceberg/test_iceberg_catalog.py
```diff
@@ -0,0 +1,56 @@
...
+            type=CatalogType.IN_MEMORY, properties={"warehouse": self.temp_dir}
+        )
+        # Initialize with the PyIceberg catalog
+        deltacat.put_catalog(
+            catalog_name, impl=deltacat.IcebergCatalog, **{"config": config}
+        )
+
+        table_def = deltacat.create_table(
+            "test_table", catalog=catalog_name, schema=schema_a
+        )
```
MingshiPeng, line 54: did you encounter the catalog re-initialization issue described here? https://github.com/ray-project/deltacat/issues/500

was it fixed in the PR or you find a way to bypass that?

deltacat/catalog/interface.py
```diff
@@ -18,6 +18,20 @@
 from deltacat.types.tables import TableWriteMode
 
 
+# catalog functions
+def initialize(*args, **kwargs) -> Optional[Any]:
```
MingshiPeng, line 22: we have `args` and `kwargs` almost everywhere in DeltaCAT method signature, this provides a lot of flexibility at the cost of strong types, readability and testability. I feel we should move away from those wildcard args but I'm also curious on other folks' opinion on this. 

This comment is not specific to this PR as this is a pre-existing issue.

MingshiPeng (approved):
overall LGTM, left some smaller comments.

deltacat/catalog/model/catalog.py
```diff
@@ -120,11 +178,29 @@ def init(
...
+
+    """
+
+    force_reinitialize = kwargs.get("force_reinitialize", False)
+
+    # get namespace from ray_init_args, then kwargs, then default
+    if ray_init_args and "namespace" in ray_init_args:
+        namespace = ray_init_args["namespace"]
+    else:
+        namespace = kwargs.get("namespace", _default_namespace)
```
pdames, line 197: Since this will give preference to the namespace inside `ray_init_args` anyways, can we just get rid of the ability to specify the namespace in either location, and then update the docstring for `ray_init_args` to note how namespaces will be given special handling (i.e., added to the list of registered catalogs). On the other hand, if we just drop Ray namespaces, then this also just goes away.

deltacat/catalog/model/catalog.py
```diff
@@ -136,46 +212,107 @@ def init(
         Catalog, serializer=Catalog.__reduce__, deserializer=Catalog.__init__
     )
 
-    global all_catalogs
+    global catalog_registry
 
-    all_catalogs = Catalogs.remote(
+    catalog_registry[namespace] = Catalogs.remote(
```
pdames, line 217: Since the parent dictionary is now used by multiple test processes but isn't multi-process safe, this seems like it has the potential to ultimately introduce more problems than it solves.

We also have the implicit constraint that each actor reserves a single CPU by default, meaning that we can only support as many namespaces as CPUs on the host. We also have the catalog `namespace` param starting to bleed into our existing catalog APIs, even though we currently only intend to use it for testing.

Would it be easier and more straightforward to just have all tests assign a UUID name to their catalog within a single shared actor unless they intend to share it with other tests? A suite of tests would then presumably have a single test suite setup method to ensure that the catalog manager actor is initialized before letting it safely synchronize each catalog added by concurrent tests.

deltacat/catalog/model/catalog.py
pdames: This marks the end of positional arguments, and looks right to me in this case (i.e., the minimal signature for invoking this method would look like `put_catalog("my_catalog")`.

deltacat/catalog/main/impl.py
mcember: Going to leave as `main` since it follows convention from the `storage` module. I think `main` is more clear than `default` since it implies that it is the sort of primary implementation, whereas `default` implies it is a peer implementation but set as the default compared with iceberg, etc

deltacat/catalog/model/catalog.py
mcember: fixed

deltacat/tests/catalog/iceberg/test_iceberg_catalog.py
mcember: This test actually didn't fail because it only called a single catalog operation (create_table). I added more catalog operations to the test and it failed.

I replaced the in memory catalog with a local SQL catalog. The only difference is that the metadata which was stored in memory, like namespaces, is instead stored in a local sqlite db file.

After doing so, the tests passed, presumably because the catalog could be re-initialized and operate deterministically based on the shared filesystem state

The bug still exists -the question is do we need to fix it?

deltacat/catalog/main/impl.py
mcember: I removed deltacat_storage_kwargs since they were merged with kwargs. I suspect we may be able to remove reader_kwargs too but keeping it for now to keep scope of refactoring small

deltacat/catalog/interface.py
mcember: This function legitimately does need args and kwargs because there are different catalog types which accept different args/kwargs.

In general, my understanding is that it is "pythonic" to keep function signatures passing args and kwargs. We should ask Patrick who has a better understanding of best practices here. Also note that it is important to keep the library backwards compatible, so using args/kwargs will help in cases where the signature evolves but you don't want to break clients who wrote code against an older signature

deltacat/catalog/model/catalog.py
```diff
@@ -120,11 +178,29 @@ def init(
     *args,
     **kwargs,
 ) -> None:
-    if is_initialized():
+    """
+    Initialize DeltaCAT catalogs.
+
+    :param catalogs: Either a single Catalog instance of a map of string to Catalog instance
```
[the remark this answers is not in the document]
mcember: fixed

deltacat/catalog/model/catalog.py
```diff
@@ -25,22 +42,44 @@ def __init__(self, impl=deltacat_catalog, *args, **kwargs):
...
         self._impl = impl
-        self._native_object = self._impl.initialize(*args, **kwargs)
+        self._inner = self._impl.initialize(*args, **kwargs)
         self._args = args
         self._kwargs = kwargs
 
+    @classmethod
+    def iceberg(cls, config: IcebergCatalogConfig, *args, **kwargs):
+        """
+        !!! ICEBERG SUPPORT IS EXPERIMENTAL !!!
```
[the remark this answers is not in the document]
mcember: done, added annotations.py with ExperimentalAPI / DeveloperAPI / PublicAPI

deltacat/catalog/model/catalog.py
mcember: I left in kwargs for consistency with the fact that namespace is passed via kwargs to get_catalog, etc

deltacat/catalog/model/catalog.py
mcember: (copied from slack) - 

I feel the same way about using Ray namespaces. How about - we will only use namespaces to narrowly target code involving registering and fetching catalog (i.e. the Catalogs class and the Daft/DC shim for creating and registering Catalogs). Since, it is impossible to test this code safely and in parallel if we're using one non-namespaced ray instance
Any and all other tests will just share the global ray instance and use a random id for the catalog name to achieve logical isolation

deltacat/catalog/interface.py
pdames: `*args` and `**kwargs` were overused initially to support easier initial API evolution until they reach a steady state. However, going forward, we generally shouldn't use `**kwargs` as a placeholder for known named variables or `*args` as a placeholder for known positional variables that we plan to keep around - instead they should be broken out into documented/named parameters for clarity. After that, the primary reason to keep `*args` and `**kwargs` around is for APIs that need to pass them into a downstream API that it doesn't control (e.g., all of the given `**kwargs` for some DeltaCAT API will be passed to some downstream pyarrow or pandas API for reading a table, or they will be passed to a Ray Core API, etc.).

deltacat/catalog/model/catalog.py
pdames: The problem here is that the `catalog_registry` itself is not an actor or otherwise multi-process safe object - it's just a plain Python dictionary that's being potentially modified by multiple concurrent processes. If it was in fact a single actor (like it was before), then we only need to sync parallel processes on actor creation/availability, and then the actor itself will synchronize all concurrent processes that try to initialize new catalogs from that point forward.

pdames (approved):

deltacat/tests/catalog/iceberg/test_iceberg_catalog.py
MingshiPeng: this bug will cause issues:
1. when setting up unit tests when everything is in memory
2. maybe some performance issue later down the road due to repeated re-initialization.

Item 2 is low priority atm, if we can get around item 1 then we can deprioritize the fix for now. I guess we can keep going until we hit a blocker that we absolutely have to fix it.

PR #519, opened 2025-03-31
[In progress] Integrate with Daft catalog

mcember:
## Summary

NOTE: this PR also pulls in changes from https://github.com/ray-project/deltacat/pull/517

We can merge that PR first and then review this one

Changes:
1. Add `deltacat_daft_patch.py` which implements Daft catalog API for DeltaCAT tables
2. Test create_table / get_table on DC native and Iceberg tables


deltacat/experimental/daft/deltacat_daft_patch.py
```diff
@@ -0,0 +1,327 @@
...
+        Refer to `deltacat.create_table` for full list of keyword arguments accepted by create_table.
+
+        This implementation is modeled after daft/catalog/_iceberg, in particular that the TableSource
+        input may be either a str/DataFrame/schema
+
+        TODO (questions to discuss with Daft):
+        1. Can we expect that `identifier` is always populated?
+            In that case, what is the purpose of `source` being a str?
+        2. Do we expect end users to call this function? Does it make sense to plumb kwargs through to DeltaCAT createTable?
+        3. What format do we expect identifier strings to be in?
```
rchowell, line 97: > 1. Can we expect that `identifier` is always populated? In that case, what is the purpose of `source` being a str?

Yes the identifier will always be populated. The string source is for creating a table from a filepath which is supported in our SQL, but not yet on this API. That is, today you can do `SELECT * FROM '/path/to/file.parquet'` so ideally we can do `daft.create_table("my_table", "/path/to/file.parquet")` but yes not yet implemented here.

> 2. Do we expect end users to call this function? Does it make sense to plumb kwargs through to DeltaCAT createTable?

Yes we do, and forwarding the args was the intention 👍

 > 3. What format do we expect identifier strings to be in?

Answer: Dot-delimited strings

Details: A fun topic ... we have an `Identifier.from_str` and an `Identifier.from_sql`. The former behaves like pyiceberg and splits on dots `.` so `"a.b" -> Identifier("a", "b")`. The latter calls out to our rust sql parser and will optionally case-normalize LOWER the simple (not-delimited/non-double-quoted) identifiers. The former is what python people are used to, and the latter is compatible with PostgreSQL and DuckDB identifiers.

deltacat/experimental/daft/deltacat_daft_patch.py
```diff
@@ -0,0 +1,327 @@
...
+        """
+        t = self._create_table_from_schema(ident, source.schema(), **kwargs)
+        # TODO (mccember) append data upon creation
+        return t
+
+    def _create_table_from_path(
+        self, ident: Identifier | str, source: str, **kwargs
+    ) -> Table:
+        """Create a table from a path."""
+        raise ValueError("table from path not yet supported")
```
rchowell, line 127: Here's that `CREATE TABLE T AS '/path/to/file.parquet'` which isn't in yet

deltacat/experimental/daft/deltacat_daft_patch.py
```diff
@@ -0,0 +1,327 @@
...
+        elif isinstance(ident, str):
+            # TODO (mccember) implement once format of string confirmed with Daft
+            raise ValueError(
+                "Usage of str identifier for DeltaCAT Daft tables not currently support"
+            )
+        else:
+            raise ValueError(
+                f"Expected identifier for createTable to be either Daft Identifier or str. "
+                f"Provided identifier was class: {type(ident)}"
+            )
```
rchowell, line 224: Format confirmed, but you also have a choice. I've been doing the same as iceberg. So here's my common snippet.

```python
if isinstance(identifier, str):
    identifier = Identifier.from_str(identifier)
```

Which calls out to the split on "." implementation. This is much more sane for python people, but you do have the `Identifier.from_sql(input, normalize: bool =False)` as an option.

deltacat/experimental/daft/deltacat_daft_patch.py
```diff
@@ -0,0 +1,327 @@
...
+            t = DeltaCATTable.__new__(DeltaCATTable)
+            t._inner = obj
+            return t
+        raise ValueError(f"Unsupported DeltaCAT table type: {type(obj)}")
+
+    def read(self, **options) -> DataFrame:
+        raise NotImplementedError("Not implemented")
+
+    def write(self, df: DataFrame | object, mode: str = "append", **options):
+        raise NotImplementedError("Not implemented")
```
rchowell, line 263: Here's where we'll create a deltacat ScanOperator then DataFrame from this ScanOperator impl 🙌

deltacat/experimental/daft/deltacat_daft_patch.py
```diff
@@ -0,0 +1,327 @@
...
+
+            Returns:
+                Table: new daft table instance from the DeltaCAT table object.
+            """
+            return DeltaCATTable._from_obj(table)
+
+        DaftTable.from_deltacat = from_deltacat
+
+    # Also update the Catalog._from_obj method to try DeltaCAT as well
+    original_from_obj = DaftCatalog._from_obj
```
rchowell, line 308: this could be tested with that,

```python
sess = Session()
sess.attach_catalog(my_deltacat_catalog_instance)
sess.list_tables()
```

deltacat/experimental/daft/daft_catalog.py
```diff
@@ -0,0 +1,233 @@
...
+        self, identifier: Identifier | str, source: TableSource | object, **kwargs
+    ) -> Table:
+        """
+        Create a DeltaCAT table via Daft catalog API
+
+        End users calling create_table through the daft table API may provide kwargs which will be plumbed through
+        to deltacat create_table. For full list of keyword arguments accepted by create_table.
+
+        :param identifier: Daft table identifier. Sequence of strings of the format (namespace) or (namespace, table)
+            or (namespace, table, table version). If this is a string, it is a dot delimited string of the same format.
```
rchowell, line 71: nit: perhaps adding that the identifier can be constructed like `Identifier("namespace", "table", "version")` or using `Identifier.from_str("namespace.table.version")`

deltacat/experimental/daft/daft_catalog.py
```diff
@@ -0,0 +1,233 @@
...
+        """Create a new namespace in the catalog."""
+        if isinstance(identifier, Identifier):
+            identifier = str(identifier)
+        self.dc_catalog.impl.create_namespace(identifier, inner=self.dc_catalog.inner)
+
+    def create_table(
+        self, identifier: Identifier | str, source: TableSource | object, **kwargs
+    ) -> Table:
+        """
+        Create a DeltaCAT table via Daft catalog API
```
rchowell, line 65: I just simplified create_table to only take `Schema | DataFrame` - the str/path will be handled on the rust side internally and will be sent to the catalog as a DataFrame like `DataFrame.read_parquet`. The TableSource is also not necessary on the python side, only the rust side. It has been removed. https://github.com/Eventual-Inc/Daft/commit/95fb910bb9ba264366ef4ccea600a22c144355d5

rchowell: I see you already have your own kwargs which is nice, I will prioritize this minor addition as well, https://github.com/Eventual-Inc/Daft/issues/4195

MingshiPeng:
FYI - you can run `make lint` before sending out the PR to resolve any auto-formatting issues to make sure the github checks can pass.

deltacat/experimental/daft/daft_catalog.py
```diff
@@ -0,0 +1,233 @@
...
+
+    TODO (mccember) in follow up PR we need to consider how to keep the DeltaCAT Catalogs class and Daft session in sync,
+      and the user-facing entrypoint to get a Daft catalog
+
+    This class itself expects a `Catalog` and will invoke the underlying implementation
+    similar to `deltacat.catalog.delegate.py`, like:
+      catalog.impl.create_namespace(namespace, inner=catalog.inner)
+
+    We cannot route calls through the higher level catalog registry / delegate since this wrapper class is at a lower
+     layer and does not manage registering catalogs.
```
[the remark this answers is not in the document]
mcember: Having that interface would help to keep the daft/DC catalogs in sync. 

I'm not sure we necessarily need it, since it feels like a pre-mature abstraction or not generally useful. Let's wait until we implement the synchronization to see how bad / if it becomes clear that a better abstraction daft-side could help

deltacat/experimental/daft/daft_catalog.py
mcember: sounds good

PR #505, opened 2025-03-12
Bugfix: `s3_file_to_parquet` function doesn't override content encoding if provided with the override kwarg

pfaraone:
## Summary
This change no longer raises ContentTypeValidationError from  `s3_file_to_parquet` if appropriately overridden

## Rationale
-Although the ["Allow overriding content encoding" bugfix](https://github.com/ray-project/deltacat/pull/462/files)
was merged into DeltaCAT and released in [1.1.30](https://github.com/ray-project/deltacat/releases/tag/1.1.30), the overriding does not properly happen during the compaction session which results in in a raised [ContentTypeValidationError](https://github.com/ray-project/deltacat/blob/79475284f2539f8b1b82fe80878ab593da930d0b/deltacat/utils/pyarrow.py#L483)
when the underlying file is gzip encoded - even if the `override_content_encoding_for_parquet` is provided. 

## Changes
- Added check for override_content_encoding_for_parquet
- Adding unit testing coverage for `s3_file_to_parquet` from no coverage including the default case

## Impact
- Small blast radius - this should allow for additional tables to be onboard if they are gzip encoded but overriden

## Testing
- Added unit testing coverage

## Regression Risk
- Low as the control flow is only changed if `override_content_encoding_for_parquet` kwarg is provided

## Checklist

- [x] Unit tests covering the changes have been added
  - [ ] If this is a bugfix, regression tests have been added

- [ ] E2E testing has been performed

## Additional Notes

Any additional information or context relevant to this PR.

PR #521, opened 2025-04-04
Bugfix: Uniform delta creation (compaction path) not propagating the file_reader_kwargs_provider when appending content type params to the deltas

pfaraone:
## Summary
Uniform delta creation (compaction path) not propagating the `file_reader_kwargs_provider` when appending content type params to the deltas

## Rationale
- Fixes the `override_content_encoding_for_parquet` bug for compaction flow described in [505](https://github.com/ray-project/deltacat/pull/505/files) and [462](https://github.com/ray-project/deltacat/pull/462/files). Content encoding is appropriately overridden for *resource estimation* but not building uniform deltas during *compaction* 

## Changes
- `append_content_type_params` signature updated to accept the `file_reader_kwargs_provider` explicitly as an optional parameter
- Additional logging

## Impact
- Optional parameter added to the signature of public method `append_content_type_params`. Could impact callers that invoke `append_content_type_params` using positional arguments. `append_content_type_params` is called within DeltaCAT by keyword - [1](https://github.com/ray-project/deltacat/blob/2bd5c4cc6c58b13db6973060b9e86225b055d0eb/deltacat/compute/resource_estimation/delta.py#L97), [2](https://github.com/ray-project/deltacat/blob/2bd5c4cc6c58b13db6973060b9e86225b055d0eb/deltacat/compute/compactor_v2/utils/io.py#L115)

## Testing
- Tested with internally. Test results will be shared via internal communication

## Regression Risk

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/compute/compactor_v2/utils/content_type_params.py
+++ b/deltacat/compute/compactor_v2/utils/content_type_params.py
@@ -5,6 +5,7 @@ from deltacat.compute.compactor_v2.constants import (
     TASK_MAX_PARALLELISM,
     MAX_PARQUET_METADATA_SIZE,
 )
+from deltacat.utils.common import ReadKwargsProvider
 from deltacat.utils.ray_utils.concurrency import invoke_parallel
 from deltacat import logs
 from deltacat.storage import (
@@ -75,11 +76,16 @@ def _download_parquet_metadata_for_manifest_entry(
     entry_index: int,
     deltacat_storage: unimplemented_deltacat_storage,
     deltacat_storage_kwargs: Optional[Dict[Any, Any]] = {},
+    file_reader_kwargs_provider: Optional[ReadKwargsProvider] = None,
 ) -> Dict[str, Any]:
+    logger.info(
+        f"PDEBUG:_download_parquet_metadata_for_manifest_entry-{deltacat_storage_kwargs=}"
+    )
     pq_file = deltacat_storage.download_delta_manifest_entry(
         delta,
         entry_index=entry_index,
         table_type=TableType.PYARROW_PARQUET,
+        file_reader_kwargs_provider=file_reader_kwargs_provider,
         **deltacat_storage_kwargs,
     )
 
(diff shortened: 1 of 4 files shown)


deltacat/compute/compactor_v2/utils/content_type_params.py
```diff
@@ -75,11 +76,16 @@ def _download_parquet_metadata_for_manifest_entry(
...
+    file_reader_kwargs_provider: Optional[ReadKwargsProvider] = None,
 ) -> Dict[str, Any]:
+    logger.info(
+        f"Downloading the parquet metadata for Delta with locator {delta.locator} and entry_index: {entry_index}"
+    )
     pq_file = deltacat_storage.download_delta_manifest_entry(
         delta,
         entry_index=entry_index,
         table_type=TableType.PYARROW_PARQUET,
+        file_reader_kwargs_provider=file_reader_kwargs_provider,
```
raghumdani, line 88: should we remove `file_reader_kwargs_provider` from `deltacat_storage_kwargs` to prevent any conflicts here?

pfaraone: Should we check and pop it out? That seems reasonable

raghumdani: Yeah.

pfaraone: Done

deltacat/compute/compactor_v2/utils/content_type_params.py
```diff
@@ -75,11 +76,21 @@ def _download_parquet_metadata_for_manifest_entry(
     entry_index: int,
     deltacat_storage: unimplemented_deltacat_storage,
     deltacat_storage_kwargs: Optional[Dict[Any, Any]] = {},
+    file_reader_kwargs_provider: Optional[ReadKwargsProvider] = None,
 ) -> Dict[str, Any]:
+    logger.info(
+        f"Downloading the parquet metadata for Delta with locator {delta.locator} and entry_index: {entry_index}"
+    )
+    if deltacat_storage_kwargs.get("file_reader_kwargs_provider") is not None:
```
raghumdani, line 84: How about using in operator here? For example, if `deltacat_storage_kwargs={"file_reader_kwargs_provider": None}`, you would still see the keyword arg conflict runtime error.

pfaraone: Updated

raghumdani (approved):
LGTM

PR #523, opened 2025-04-07
Remove ray namespace support in Catalogs & minor test refactors

mcember:
## Summary
Follow on from PR https://github.com/ray-project/deltacat/pull/517

Remove Namespace support in Catalogs. Also refactor compute tests to use a sqlite DB in a temporary directory, since the tests were not properly cleaning up the sqlite file that previously used the test directory

### DIFF
--- a/deltacat/catalog/model/catalog.py
+++ b/deltacat/catalog/model/catalog.py
@@ -257,7 +257,6 @@ def put_catalog(
     Args:
         name: name of catalog
         catalog: catalog instance to use, if provided
-        impl: catalog module to initialize. Only used if `catalog` param not provided
         default:  Make this the default catalog if multiple catalogs are available.
             ignored if this is the only catalog available, since it will always be the default catalog.
         ray_init_args: ray initialization args (used only if ray not already initialized)
@@ -288,7 +287,7 @@ def put_catalog(
             pass
         if catalog_already_exists:
             raise ValueError(
-                f"Failed to put catalog {name} because it already exists and fail_if_exists=True"
+                f"Failed to put catalog {name} because it already exists and fail_if_exists={fail_if_exists}"
             )
 
     # Add the catalog (which may overwrite existing if fail_if_exists=False)
--- a/deltacat/catalog/model/catalog.py
+++ b/deltacat/catalog/model/catalog.py
@@ -103,14 +103,13 @@ class Catalogs:
     def __init__(
         self,
         catalogs: Union[Catalog, Dict[str, Catalog]],
-        default_catalog_name: Optional[str] = None,
+        default: Optional[str] = None,
         *args,
         **kwargs,
     ):
-        if default_catalog_name and default_catalog_name not in catalogs:
+        if default and default not in catalogs:
             raise ValueError(
-                f"Catalog {default_catalog_name} not found "
-                f"in catalogs to register: {catalogs}"
+                f"Catalog {default} not found " f"in catalogs to register: {catalogs}"
             )
         if not catalogs:
             raise ValueError(
@@ -123,8 +122,8 @@ class Catalogs:
             catalogs = {DEFAULT_CATALOG: catalogs}
 
         self.catalogs: Dict[str, Catalog] = catalogs
-        if default_catalog_name:
-            self.default_catalog = self.catalogs[default_catalog_name]
+        if default:
+            self.default_catalog = self.catalogs[default]
         elif len(catalogs) == 1:
             self.default_catalog = list(self.catalogs.values())[0]
         else:
@@ -152,35 +151,35 @@ def is_initialized(*args, **kwargs) -> bool:
     """
     Check if DeltaCAT is initialized
     """
-    import deltacat.catalog.model.catalog as catalog_module
+    global all_catalogs
 
     # If ray is not initialized, then Catalogs cannot be initialized
     if not ray.is_initialized():
         # Any existing actor reference stored in catalog_module must be stale - reset it
-        catalog_module.all_catalogs = None
+        all_catalogs = None
         return False
 
-    return catalog_module.all_catalogs is not None
+    return all_catalogs is not None
 
 
 def init(
     catalogs: Union[Dict[str, Catalog], Catalog],
-    default_catalog_name: Optional[str] = None,
+    default: Optional[str] = None,
     ray_init_args: Dict[str, Any] = None,
     *args,
+    force_reinitialize=False,
     **kwargs,
 ) -> None:
     """
     Initialize DeltaCAT catalogs.
 
     :param catalogs: Either a single Catalog instance or a map of string to Catalog instance
-    :param default_catalog_name: The Catalog to use by default. If only one Catalog is provided, it will
+    :param default: The Catalog to use by default. If only one Catalog is provided, it will
         be set as the default
     :param ray_init_args: kwargs to pass to ray initialization
+    :param force_reinitialize: if True, force the reinitialization of Ray. If false, will do nothing if ray already initialized
     """
-    import deltacat.catalog.model.catalog as catalog_module
-
-    force_reinitialize = kwargs.get("force_reinitialize", False)
+    global all_catalogs
 
     if is_initialized() and not force_reinitialize:
         logger.warning("DeltaCAT already initialized.")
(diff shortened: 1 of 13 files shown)


deltacat/tests/compute/resource_estimation/test_delta.py
```diff
@@ -21,21 +20,6 @@
...
-    )
-    # see deltacat/tests/local_deltacat_storage/README.md for documentation
-    kwargs_for_local_deltacat_storage = {
-        DATABASE_FILE_PATH_KEY: DATABASE_FILE_PATH_VALUE,
-    }
-    yield kwargs_for_local_deltacat_storage
-    if os.path.exists(DATABASE_FILE_PATH_VALUE):
-        os.remove(DATABASE_FILE_PATH_VALUE)
-
-
```
MingshiPeng, line 38: info: I assume the reason to remove those method is because they are moved to `conftest.py` file, however I don't see the `conftest/py` version of this method getting imported anywhere. How is the importing configured?

deltacat/catalog/model/catalog.py
```diff
@@ -16,9 +16,7 @@
 from deltacat.catalog.iceberg import IcebergCatalogConfig
 from deltacat.constants import DEFAULT_CATALOG
 
-# registry of namespace to ActorHandle for Catalogs class
-catalog_registry: Dict[str, ray.actor.ActorHandle] = {}
-_default_namespace = "DEFAULT"
```
MingshiPeng, line 21: I'm having trouble understanding the use case of Ray Namespace in DeltaCAT (but since we are removing that, and the code doesn't seem to show up in many places, which probably makes it not so important.) Will ask during 4/9 PythonSDK standup parking lot.

Will approve for now to not block code merge

MingshiPeng (approved):

deltacat/tests/compute/resource_estimation/test_delta.py
MingshiPeng: got the answer by searching `conftest.py`.

deltacat/catalog/model/catalog.py
```diff
@@ -213,62 +196,57 @@ def init(
...
+                f"{available_catalogs}."
+            )
+        return catalog
 
 
 def put_catalog(
     name: str,
+    catalog: Catalog = None,
     *,
     impl: ModuleType = DeltacatCatalog,
```
pdames, line 248: Do we need to provide two different ways of adding the catalog here? Why not standardize on just one? Also, since this is to be a user-facing API, I don't think we should to commit to supporting the ability to provide custom catalog implementations long-term, and thus shouldn't expose this as an explicit parameter.

deltacat/catalog/model/catalog.py
```diff
@@ -277,44 +255,32 @@ def put_catalog(
 
     You may explicitly provide an initialized Catalog instance, like from the Catalog constructor or
     from factory methods like Catalog.default or Catalog.iceberg
-
     Otherwise, this function initializes the catalog by using the catalog implementation provided by `impl`.
```
pdames, line 258: I think we need to drop the ability to provide or instantiate separate catalog implementations as top-level/user-facing APIs here, since it also implies a long-term dedication to supporting multiple separate implementations of the DeltaCAT catalog API as thin wrappers over other catalog APIs vs. supporting them via synchronization. Subsequently dropping the `default` vs. `iceberg` Catalog factory methods should also be part of this (we can move the factory method directly to the Iceberg catalog module itself, but it shouldn't be exposed in a common/core catalog location like this).

pdames (changes requested):

deltacat/catalog/model/catalog.py
```diff
@@ -277,44 +255,32 @@ def put_catalog(
...
+        ray.get(catalog_module.all_catalogs.put.remote(name, catalog, set_as_default))
     else:
-        init({name: catalog}, ray_init_args=ray_init_args, **{"namespace": namespace})
+        # NOTE - since we are initializing with a single catalog, it will be set to the default
+        if not set_as_default:
+            logger.warning(
+                f"Calling put_catalog with set_as_default=False, "
+                f"but still setting Catalog {catalog} as default since it is the only catalog."
+            )
+        init({name: catalog}, ray_init_args=ray_init_args, **kwargs)
```
pdames, line 284: Overloading `**kwargs` as input to both `init` and `Catalog` seems a little dangerous/confusing. Also, we're not actually using `**kwargs` in this case, and they could simply be omitted here.

deltacat/catalog/model/catalog.py
mcember: that's fine - I will remove the ability to provide impl and init from kwargs and just require users to provide the Catalog

deltacat/catalog/model/catalog.py
mcember: Discussing offline on slack

deltacat/catalog/model/catalog.py
```diff
@@ -189,20 +178,14 @@ def init(
         be set as the default
     :param ray_init_args: kwargs to pass to ray initialization
     """
+    import deltacat.catalog.model.catalog as catalog_module
```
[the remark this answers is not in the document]
mcember: these import statements weren't from an auto-refactor, it was from modifying the scope of all_catalogs from global to module. If it were a global variable, you would still need a line `global all_catalogs` to effectively signal that you're working in the global scope (whereas this is just getting a handle to the module scope for later reference of all_catalogs). IMO it's more proper encapsulation for this variable to be in the module scope, so leaving as is

deltacat/catalog/model/catalog.py
```diff
@@ -213,108 +195,101 @@ def init(
...
     ray_init_args: Dict[str, Any] = None,
+    fail_if_exists: bool = False,
     **kwargs,
-) -> Catalog:
+) -> None:
     """
     Add a named catalog to the global map of named catalogs. Initializes ray if not already initialized.
 
     You may explicitly provide an initialized Catalog instance, like from the Catalog constructor or
     from factory methods like Catalog.default or Catalog.iceberg
```
pdames, line 254: Can we remove lines 253-255 since `Catalog.iceberg` and `impl` no longer exist?

deltacat/catalog/model/catalog.py
```diff
@@ -213,108 +195,101 @@ def init(
...
-
     Otherwise, this function initializes the catalog by using the catalog implementation provided by `impl`.
 
     Args:
         name: name of catalog
         catalog: catalog instance to use, if provided
         impl: catalog module to initialize. Only used if `catalog` param not provided
```
pdames, line 260:
```suggestion
```

deltacat/catalog/model/catalog.py
```diff
@@ -213,108 +195,101 @@ def init(
...
-    return catalog
+    else:
+        catalog = ray.get(all_catalogs.default.remote())
+        if not catalog:
+            available_catalogs = ray.get(all_catalogs.all.remote()).values()
+            raise ValueError(
+                f"Call to get_catalog without name set failed because there is no default Catalog set. Available catalogs: "
+                f"{available_catalogs}."
+            )
+        return catalog
```
pdames, line 238: We can reduce code duplication here by taking advantage of the fact that the only divergent logic in each branch involves raising a different ValueError based on whether `name` is set or not (we should also wrap the error message to 80 chars per line: https://peps.python.org/pep-0008/):
```python
    catalog = (
        ray.get(all_catalogs.get.remote(name))
        if name
        else ray.get(all_catalogs.default.remote())
    )
    if not catalog:
        available_catalogs = ray.get(all_catalogs.all.remote()).values()
        if name:
            raise ValueError(
                f"Catalog '{name}' not found. Available catalogs: "
                f"{available_catalogs}."
            )
        else:
            raise ValueError(
                f"Call to get_catalog without name set failed because there "
                f"is no default Catalog set. Available catalogs: "
                f"{available_catalogs}."
            )
```

deltacat/catalog/model/catalog.py
```diff
@@ -213,108 +195,101 @@ def init(
...
+            get_catalog(name)
+            # Note - need to set state catalog_already_exists and throw ValueError later, or else it will be
+            # caught in the except block which is meant to catch the ValueError from get_catalog
+            catalog_already_exists = True
+        except ValueError:
+            pass
+        if catalog_already_exists:
             raise ValueError(
-                f"Catalog {name} already exists in namespace '{namespace}'."
+                f"Failed to put catalog {name} because it already exists and fail_if_exists=True"
```
pdames, line 291: For both a sanity check of this error message and to (hopefully) draw attention of future devs who may change the `fail_if_exists` param name, can we change this to:
```suggestion
                f"Failed to put catalog {name} because it already exists and fail_if_exists={fail_if_exists}"
```

pdames (changes requested):
Looks pretty good overall - just a few minor changes to merge.

PR #525, opened 2025-04-09
create_scan_plan interface & Iceberg implementation

MingshiPeng:
## Summary

- add `create_scan_plan` API in DeltaCAT Catalog interface, this is part of Daft-DeltaCAT read path integration.
- add basic class for `ScanPlan`, `ScanTask`, `Pushdown`
- add dummy Iceberg implementation for this API - which just calls iceberg `scan` API and return all iceberg data files into DeltaCAT ScanPlan object.

## Testing
- I tested Iceberg implementation of this API using an existing Glue Table preloaded with test data, the response look like this:
```
ScanPlan(scan_tasks=[FileScanTask(data_file_list=[DataFile(file_path='s3://mingshi-test-andes-local/iceberg/warehouse/test_namespace.db/test_table_bucketed/data/symbol_bucket=1/0fcd8894-1f9a-43fa-bd11-e45a5cc7c62a-0.parquet')]), FileScanTask(data_file_list=[DataFile(file_path='s3://mingshi-test-andes-local/iceberg/warehouse/test_namespace.db/test_table_bucketed/data/symbol_bucket=0/80a9262d-4e75-48d0-90f9-cde7ed8f3545-0.parquet')])])
```
- I couldn't setup unit test, because the DeltaCAT `write_to_table` API only supports S3 destinations atm. (will work on this in follow up stories as this is crucial to setup e2e local testing)


MingshiPeng:
@mcember @pdames @anshumankomawar Could you review when you get time? Thanks.

deltacat/catalog/delegate.py
```diff
@@ -67,6 +69,19 @@ def read_table(
         **kwargs,
     )
 
+def create_scan_plan(
```
pdames, line 72: Can we move this down from `catalog` to `storage`?

Everything in `catalog` is exposed to DeltaCAT end-users, but this should be serving compute framework integrations and the typical DeltaCAT end-user shouldn't ever need to know about it. 

As an example, the APIs in `catalog` should provide roughly equivalent functionality to the Iceberg Catalog API: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/catalog/Catalog.java.

deltacat/__init__.py
```diff
@@ -93,6 +94,7 @@
     "default_namespace",
     "write_to_table",
     "read_table",
+    "create_scan_plan",
```
pdames, line 97: Can we move this from `deltacat/__init__.py` to `deltacat/storage/__init__.py` (the average DeltaCAT end-user shouldn't need to invoke `create_scan_plan` directly).

deltacat/storage/model/push_down.py
```diff
@@ -0,0 +1,13 @@
+class RowFilter:...
```
pdames, line 1: Can we move this file and the other scan models into dedicated `storage/model/scan/*.py` files?

deltacat/catalog/iceberg/impl.py
```diff
@@ -9,6 +9,10 @@
 from deltacat.exceptions import TableAlreadyExistsError
 from deltacat.storage.iceberg.model import PartitionSchemeMapper, SchemaMapper
 from deltacat.storage.model.partition import PartitionScheme
+from deltacat.storage.model.push_down import Pushdown
+from deltacat.storage.model.scan_plan import ScanPlan
+from deltacat.storage.model.scan_task import FileScanTask, DataFile
+from deltacat.storage.iceberg.impl import _try_load_iceberg_table
```
pdames, line 15: Instead of bringing in this protected method from `storage`, can we move `create_scan_plan` down to the `storage` layer?

pdames (changes requested):

deltacat/catalog/delegate.py
pdames: Or, since we're requiring a `table` to always be specified anyways, attaching a `scan` method to `TableDefinition` may also make sense for now.

deltacat/storage/model/scan_task.py
```diff
@@ -0,0 +1,32 @@
...
+class FileScanTask(ScanTask):
+    """A unit of data in the form of data files
+    """
+    data_file_list: list[DataFile]
+
+    def data_files(self) -> list[DataFile]:
+        return self.data_file_list
+
+
+class ShardedScanTask(ScanTask):
```
mcember, line 27: How do you plan to use ShardedScanTask?

I had thought that the create_scan_plan API would accept some sort of shard parameter (e.g. you say "this is shard 1 out of 10, and the table has some ShardingStrategy that identifies that sharding is based on e.g. row count or primary key or whatever). 

I would think though that when you ask for a sharded scan plan, it just returns the normal ScanTask class.

deltacat/storage/model/scan_task.py
```diff
@@ -0,0 +1,32 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+
+@dataclass
+class DataFile:
+    """Represents a data file, e.g. a S3 object, or a local file.
+    """
+    file_path: str
+
+class ScanTask(ABC):
```
mcember, line 10: In general, have you found any comparable interfaces to inform whether you will need anything other than just a list of data files? I took a quick look at Iceberg ScanTask (https://iceberg.apache.org/javadoc/1.5.2/org/apache/iceberg/ScanTask.html). They have a lot more complexity but it looks like Scan Task is used more broadly than just reading data in their case.

I'm fine starting small here but want to sanity check that you've thought about how this might grow in the future

deltacat/catalog/main/impl.py
```diff
@@ -172,6 +174,13 @@ def read_table(
         params=merge_on_read_params, **kwargs
     )
 
+def create_scan_plan(
+    table: str,
+    namespace: Optional[str] = None,
+    pushdown: Optional[Pushdown] = None,
```
mcember, line 180: Does this need to be list of pushdowns? 

I forget what we discussed with Daft - did we say we were going to use or heavily imitate their pushdown interface? Or are you making your own and you will see what the connector code between pushdown interfaces looks like when you implement the daft scan operator?

deltacat/catalog/interface.py
```diff
@@ -71,6 +73,20 @@ def read_table(
     """
     raise NotImplementedError("read_table not implemented")
 
+def create_scan_plan(
+    table: str, namespace: Optional[str] = None, pushdown: Optional[Pushdown] = None
```
mcember, line 77: General comment: make sure that there are consistent conventions between how parameters are passed in versus comparable functions. At least in the catalog interface we have the convention of making most of the non-mandatory args keyword only args. I would have to look at the storage interface to see if this follows convention

mcember (approved):

deltacat/storage/model/scan_task.py
MingshiPeng: > I had thought that the create_scan_plan API would accept some sort of shard parameter

I'm aligned with that, the `create_scan_plan` API will accept certain configuration args, which will inform the API whether it should return list of FileScanTask or ShardedScanTask.

The ShardedScanTask will just contain basic information like "shard_num" and "world_size", the actual resolving to data files will be done at the worker nodes.

For the worker node compute process reading data, it will just treat it as any other ScanTask, and the process will depend on the `data_files()` method to get the actual data files.

deltacat/storage/model/scan_task.py
MingshiPeng: yes - there are tons of attributes that will be needed in the ScanTask and even in the DataFile class - I know some file level metadata will definitely be needed in order for Daft to properly prune the DataFiles. We will need to expand the class in the future.

deltacat/catalog/main/impl.py
MingshiPeng: > Does this need to be list of pushdowns?

There are scenarios where we need a collection of pushdowns, and we will end up in a situation where we need to define the relationship between different pushdowns, e.g. an AND relationship between predicate_1 and predicate_2 and then OR relationships between predicate_3 and predicate_4. List is not a good choice for modeling this relationship, so my current mental model is to use one object to model all pushdowns, where the interaction/relationship between the pushdowns can be modeled internally.

> did we say we were going to use or heavily imitate their pushdown interface?

I'm still on that path - below is the Daft interface, you will see that the fields are exactly the same. I've created another abstraction for column_filter, row_filter, partition_filter because I want the keep the door open for anything that we might want to do differently than what Daft is already doing today.
```
class Pushdowns:
    """Pushdowns from the query optimizer that can optimize scanning data sources."""

    columns: list[str] | None
    filters: PyExpr | None
    partition_filters: PyExpr | None
    limit: int | None
```

deltacat/catalog/delegate.py
MingshiPeng: Sounds good - I'll move the API to under `deltacat.catalog` as per suggestion.

deltacat/catalog/interface.py
MingshiPeng: for this API, I'll move it to storage interface and adapt Catalog Interface api conventions. I'll not include kwargs for now as I don't have a use case for it at the moment, and it's issue for strong typing.

MingshiPeng:
## Rev2
- refactor `create_scan_plan` API into `ScanPlanner` interface as per offline discussion
- move scan_plan related models under a `scan` folder

deltacat/catalog/model/table_definition.py
```diff
@@ -38,3 +43,12 @@ def stream(self) -> Stream:
...
         return self.get("nativeObject")
+
+    @property
+    def scan_planner(self) -> Optional[ScanPlanner]:
+        return self.get("scan_planner")
+
+    def create_scan_plan(self, pushdown: Optional[Pushdown] = None) -> ScanPlan:
+        if not self.scan_planner:
+            raise RuntimeError(f"ScanPlanner is not initialized for table '{self.table.table_name}' of namespace '{self.table.namespace}'")
+        return self.scan_planner.create_scan_plan(table_name=self.table.table_name, namespace=self.table.namespace, pushdown=pushdown)
```
pdames, line 54: nit/codestyle: Try to keep line lengths to 80 characters (in accordance with PEP8: https://peps.python.org/pep-0008/... primary reasons IMO are to assist with code readability of large text for individuals who may struggle to read small text, and to allow side-by-side vertical viewing of 2 codefiles on a single monitor for those that have no problem reading smaller text). 

Also prefer to separate 2+ input args on multiple lines to aid with diff resolution on subsequent PRs. (i.e., to facilitate fast visual resolution of which params were updated/added/removed in a multi-param function invocation or delcaration in subsequent PRs). Note that you'll need to end the argument list with a trailing comma to prevent Black from auto-formatting this back down to a single line.

deltacat/storage/iceberg/iceberg_scan_planner.py
```diff
@@ -0,0 +1,21 @@
...
+from deltacat.storage.model.scan.scan_task import FileScanTask, DataFile
+from deltacat.storage.util.scan_planner import ScanPlanner
+from deltacat.storage.iceberg.impl import _try_load_iceberg_table
+
+
+class IcebergScanPlanner(ScanPlanner):
+    def __init__(self, catalog: Catalog):
+        self.catalog = catalog
+
+    def create_scan_plan(self, table_name: str, namespace: Optional[str] = None, pushdown: Optional[Pushdown] = None) -> ScanPlan:
```
pdames, line 15: Same nit here - please separate declaration args onto separate lines and put a trailing comma behind the last arg.

deltacat/storage/iceberg/iceberg_scan_planner.py
```diff
@@ -0,0 +1,21 @@
...
+        iceberg_table = _try_load_iceberg_table(self.catalog, namespace=namespace, table_name=table_name)
```
pdames, line 16: Same nit here - please separate declaration args onto separate lines and put a trailing comma behind the last arg.

pdames (changes requested):
Overall structure of the code looks good - thanks! Just have some minor codestyle comments to address prior to merging.

deltacat/catalog/model/table_definition.py
MingshiPeng: thanks for providing the context - I was able to format the code by calling `make lint` which I missed earlier

deltacat/storage/iceberg/iceberg_scan_planner.py
MingshiPeng: addressed

deltacat/storage/iceberg/iceberg_scan_planner.py
MingshiPeng: addressed

pdames (approved):
LGTM!

PR #526, opened 2025-04-11
Add bucketing spec validation for existing compacted table

raghumdani:
## Summary

This change adds a bucketing spec validation to assert and identify tables whose records have drifted from their original hash bucket. 

## Rationale

We observed a small number of records which drifted from their original hash bucket in one of the production tables.  Hence, adding the validation check will detect any future drift quickly and help confirm blast radius. 

## Changes

This PR brings two features which can be enabled by setting below environment variables:
1. ENABLE_BUCKETING_SPEC_COMPLIANCE_LOGGING: Ensures that data is compliant with bucketing spec, just logs a message without failing the job. 
2. ENABLE_BUCKETING_SPEC_COMPLIANCE_ASSERTION: Ensures that data is compliant with bucketing spec, throws an error otherwise.

## Impact

The functionality, when enabled has negative impact on cost. Hence, should be enabled 

## Testing

- [x] Enabled the checks for every functional tests we had so far. 
- [x] Add explicit failure and success functional test cases.

## Regression Risk

None, as this change is gated by an environment variable. 

## Checklist

## Additional Notes

None

### DIFF
--- a/deltacat/compute/compactor_v2/constants.py
+++ b/deltacat/compute/compactor_v2/constants.py
@@ -92,3 +92,16 @@ DEFAULT_NUM_ROUNDS = 1
 SHA1_HASHING_FOR_MEMORY_OPTIMIZATION_DISABLED = env_bool(
     "SHA1_HASHING_FOR_MEMORY_OPTIMIZATION_DISABLED", False
 )
+
+# Whether to enable logging if any partition is found
+# to be non-compliant with the bucketing spec.
+ENABLE_BUCKETING_SPEC_COMPLIANCE_LOGGING = env_bool(
+    "ENABLE_BUCKETING_SPEC_COMPLIANCE_LOGGING", False
+)
+
+# Whether to fail the job with ValidationError if the
+# current compacted partition is found to be non-compliant
+# with bucketing spec.
+ENABLE_BUCKETING_SPEC_COMPLIANCE_ASSERTION = env_bool(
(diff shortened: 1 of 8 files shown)


deltacat/compute/compactor_v2/constants.py
```diff
@@ -92,3 +92,16 @@
 SHA1_HASHING_FOR_MEMORY_OPTIMIZATION_DISABLED = env_bool(
     "SHA1_HASHING_FOR_MEMORY_OPTIMIZATION_DISABLED", False
 )
+
+# Whether to enable logging if any partition is found
+# to be non-compliant with the bucketing spec.
+ENABLE_BUCKETING_SPEC_COMPLIANCE_LOGGING = env_bool(
```
pfaraone, line 98: Any reason to not default this to `True`? Cost perhaps?

raghumdani: Yes, cost.

pfaraone, line 98: What do you think about encapsulating this behavior within an `Enum`? Something like `BUCKETING_SPEC_COMPLIANCE_PROFILE`. It looks like logging and assertion are levels of severity on finding non-compliance with the bucketing spec? So logging would be on the side of a less profile that the user can select while assertion is more strict

raghumdani: done.

pfaraone (approved):
LGTM!

Issue #513, opened 2025-03-24
[V2][API] Persist DeltaCAT User Session Configuration


pdames:
DeltaCAT users typically work with the same catalogs every time they initialize DeltaCAT, but right now the only place they have to save configuration related to the catalogs they routinely work with is in code, where we require them to dynamically re-instantiate their catalogs every time DeltaCAT is initialized at runtime. A more natural experience allows users to pick up a new DeltaCAT session where the last one left-off, meaning that all catalogs (and related durable state) they had initialized in their last session will be automatically re-initialized in their next session unless the last session's configuration is explicitly overridden with a new set of catalogs to initialize.

A session starts when DeltaCAT is initialized by a user via `dc.init` and ends when it is terminated (e.g., by terminating the Python process that owns it). Zero to N catalogs are associated with each session. The session catalogs and their config should be persisted to a file on disk so that it can be restored into the all_catalogs Ray Actor the next time DeltaCAT is initialized without explicit overrides of registered ctalogs (e.g., all of a user's registered catalogs and their CatalogProperties could be saved in a location like `~/.dc/config`). Note that `~/.deltacat` should _not_ be used as the parent directory, since this directory name is already reserved for storing catalog metadata (assuming a user initializes a DeltaCAT catalog stored in their home directory).

pdames:
This should be resolved as part of https://github.com/ray-project/deltacat/pull/569.

PR #527, opened 2025-04-15
DeltaCAT ScanOperator implementation

MingshiPeng:
## Summary
- implement DeltaCAT ScanOperator for DeltaCAT-Daft interoperability in `daft_scan.py`
  - the implementation is incomplete due to missing schema conversion part. I hard-coded those for the e2e example to work. Sending out this PR for review first to make the review size manageable. I'll work in the schema conversion part in parallel.
- implement `deltacat.read_table()` to return a Daft DataFrame
- fix PyIceberg write compatibility issue due to recent version bump from 0.6 to 0.9 in `iceberg/override.py`
- repurpose `iceberg_bucket_writer.py` into an e2e DeltaCAT read & write example

## Rationale

This is required as part of DeltaCAT and Daft inter-operability story.

## Testing
- e2e example working in `iceberg_bucket_writer.py`
- unit test pending as deltacat writer only supports s3 location at the moment


deltacat/daft/daft_scan.py
```diff
@@ -0,0 +1,92 @@
...
+    def _infer_partition_keys(self) -> list[PartitionField]:
+        # TODO: remove hard coding
+        partition_field = Field.create(name="symbol_bucket", dtype=DataType.int32())
+        source_field = Field.create(name="symbol", dtype=DataType.string())
+        return [
+            make_partition_field(
+                field=partition_field,
+                source_field=source_field,
+                transform=PartitionTransform.iceberg_bucket(2),
+            )
```
MingshiPeng, line 91: the schema part is hardcoded to make e2e example work - sending this PR out for early review while I work in parallel for the schema conversion part.

deltacat/examples/iceberg/iceberg_bucket_writer.py
```diff
@@ -49,11 +72,13 @@ def run(warehouse="s3://my-bucket/my/key/prefix", **kwargs):
...
-                    "type": "glue",
-                    "region_name": "us-east-1",
-                    "warehouse": warehouse,
-                },
+                config=IcebergCatalogConfig(
+                    type=CatalogType.GLUE,
+                    properties={
+                        "warehouse": warehouse,
+                        "region_name": "us-east-1",
+                    },
```
pdames, line 80: Seeing this in action like this makes me wonder - do we foresee any benefit from having our own separate IcebergCatalogConfig class? If so, are these internal DeltaCAT developer benefits, or user-facing benefits?

The user-facing initialization here has only become more nested, and potentially less straightforward to understand since it's not a straight passthrough to the existing PyIceberg Catalog properties. Would it be more straightforward to just pass the parameters straight through to the Iceberg catalog like before, so that we don't risk continuous extension of our config class for feature parity going forward?

deltacat/daft/daft_scan.py
MingshiPeng: thanks for the suggestion - I'll do that

deltacat/daft/daft_scan.py
```diff
@@ -0,0 +1,92 @@
...
+    FileFormatConfig,
+    ParquetSourceConfig,
+)
+from daft.io.scan import ScanOperator, make_partition_field
+from daft.logical.schema import Field
+
+from deltacat.catalog.model.table_definition import TableDefinition
+
+
+class DeltaCATScanOperator(ScanOperator):
```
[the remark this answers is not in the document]
MingshiPeng: There are two naming options as you pointed out: 1) DaftScanOperator; 2) DeltaCATScanOperator.

ScanOperator is a base class vended by Daft to provide connection between Daft and catalog/datasource. Other implementations of this base class includes: 1) `IcebergScanOperator`; 2) `HudiScanOperator`; 3) `LanceDBScanOperator`. The existing implementations all prefix the class name with the integrating catalog/datasource, which is why I'm also following the convention here.

For future integration with other compute framework, I suspect we won't be creating an another implementation of `ScanOperator`. For example, for Ray implementation, we will likely implement a Ray Datasource Implementation.

Please let me know if that addresses your concern.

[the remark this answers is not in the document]
pdames: I'm not too opinionated here, since the naming makes sense if the imagined/ideal future of this operator is for it to live within the Daft project (since the naming makes sense from their perspective of catalog agnosticism vs. our perspective of compute agnosticism, and you already have it within the `daft` module so that can be considered implicit). At some point we should probably at least fix the inconsistent casing of `DeltaCAT` in our various class/function names... it seems like `DeltaCat` has a slight lead in terms of occurrences across our code base for now, so `DeltaCatScanOperator` would be my vote for now.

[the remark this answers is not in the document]
MingshiPeng: Thanks - I'll change the class name to `DeltaCatScanOperator` instead.

deltacat/examples/iceberg/iceberg_bucket_writer.py
MingshiPeng: Synced offline - Patrick suggests we should move all config loading from config file stored in disk instead of dynamically set them in code. Tracking issue: https://github.com/ray-project/deltacat/issues/513 .

deltacat/examples/iceberg/iceberg_bucket_writer.py
```diff
@@ -30,6 +34,24 @@
...
+
+    To run the script:
+        1. prepare an AWS Account
+            1. prepare a S3 location where the data will be written to, which will be used in Step 3.
+            2. prepare an IAM Role that has access to the S3 location and Glue
+        2. retrieve the IAM Role AWS Credential and cache locally in ~/.aws/credentials
+        3. run below command to execute the example
+        ```
+        make venv && source venv/bin/activate
+        python -m deltacat.examples.iceberg.iceberg_bucket_writer --warehouse=s3://<YOUR_S3_LOCATION>
```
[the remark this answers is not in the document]
MingshiPeng: Synced offline - created a tracking issue https://github.com/ray-project/deltacat/issues/528 to make the example to run locally - after which the setup requirements can be simplified.

deltacat/catalog/iceberg/overrides.py
```diff
@@ -24,11 +23,10 @@ def append(table: Table, paths: List[str]) -> None:
...
     data_files = write_file(table, paths)
-    merge = _MergingSnapshotProducer(operation=Operation.APPEND, table=table)
-    for data_file in data_files:
-        merge.append_data_file(data_file)
-
-    merge.commit()
+    with table.transaction() as txn:
+        with txn.update_snapshot().fast_append() as snapshot_update:
+            for data_file in data_files:
+                snapshot_update.append_data_file(data_file)
```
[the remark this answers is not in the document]
MingshiPeng: Synced offline - created an issue https://github.com/ray-project/deltacat/issues/529 for evaluating how to replace `overrides` logic with Iceberg native APIs.

deltacat/catalog/iceberg/impl.py
```diff
@@ -144,7 +147,18 @@ def read_table(
     table: str, *args, namespace: Optional[str] = None, **kwargs
 ) -> DistributedDataset:
     """Read a table into a distributed dataset."""
-    raise NotImplementedError("read_table not implemented")
+    # TODO: more proper IO configuration
+    io_config = context.get_context().daft_planning_config.default_io_config
+    multithreaded_io = not context.get_context().is_ray_runner
```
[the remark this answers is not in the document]
MingshiPeng: This implementation references Daft read_iceberg() implementation [here](https://github.com/Eventual-Inc/Daft/blob/ecc2970db440a8cb827ecc2f2f255b2481d7aa2d/daft/io/_iceberg.py#L98-L99).

My understanding is that if it's not running on Ray, then use single node multithreaded IO. Wondering @rchowell if can comment on this?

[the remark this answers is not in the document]
rchowell: I checked with the team, and we set multithreaded_io to false for ray since we only want to occupying one worker _thread_ per task to not overwhelm the machine runner multiple workers, one on each core.

When you have the whole machine, so native runner, you should set multithreaded_io to true.

MingshiPeng:
Rev2 changes:
- implement model mapper between Daft and DeltaCAT models for Schema, Partition Key, Field and Transform
- rename `DeltaCATScanOperator` to `DeltaCatScanOperator`

Test
- iceberg_bucket_writer e2e example works as expected: `python -m deltacat.examples.iceberg.iceberg_bucket_writer --warehouse=s3://mingshi-test-andes-local/iceberg/warehouse`
- added unit tests for model mapping logic

deltacat/daft/daft_scan.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if not (
+            self.table and self.table.table_version and self.table.table_version.schema
+        ):
+            raise RuntimeError(
+                f"Failed to infer schema for DeltaCAT Table "
+                f"{self.table.table.namespace}.{self.table.table.table_name}"
+            )
+
+        return Schema.from_pyarrow_schema(self.table.table_version.schema.arrow)
```
MingshiPeng, line 82: Per Patrick's earlier comment about raise not supported error for multi-schema - I see DeltaCAT will merge all schemas into one if multiple schema exists [link](https://github.com/MingshiPeng/deltacat/blob/scan_operator/deltacat/storage/model/schema.py#L435-L438), so curious on:
1. in what scenario will multi-schema be used in DeltaCAT or PyArrow
2. should the Daft mapper still raise error even now that the DeltaCAT schema is already merged into 1 PyArrow schema?

deltacat/daft/model.py
```diff
@@ -0,0 +1,258 @@
...
+            obj: The DaftTransform to convert
+            **kwargs: Additional arguments
+
+        Returns:
+            Converted Transform object
+        """
+
+        # daft.PartitionTransform doesn't have a Python interface for accessing its attributes,
+        # thus conversion is not possible.
+        # TODO: request Daft to expose Python friendly interface for daft.PartitionTransform
```
MingshiPeng, line 93: @rchowell can you help validate my comments here? Please let me know if there is any workaround that I can use to unblock this. If not, what's the process for raising feature request to Daft? Should I directly create an Issue in Daft repo?

deltacat/daft/model.py
```diff
@@ -0,0 +1,258 @@
...
+            obj: The DaftPartitionField to convert
+            schema: The Daft schema containing field information
+            **kwargs: Additional arguments
+
+        Returns:
+            Converted PartitionKey object
+        """
+        # Daft PartitionField only exposes 1 attribute `field` which is not enough
+        # to convert to DeltaCAT PartitionKey
+        # TODO: request Daft to expose more Python friendly interface for PartitionField
```
MingshiPeng, line 159: @rchowell same, can you help validate my comments here?

deltacat/daft/model.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @staticmethod
+    def get_daft_partition_field(
+        partition_field_name: str,
+        daft_source_field: Optional[DaftField],
+        # TODO: replace DeltaCAT transform with Daft Transform for uniformality
+        # We cannot use Daft Transform here because Daft Transform doesn't have a Python interface for us to
+        # access its attributes.
+        # TODO: request Daft to provide a more python friendly interface for Daft Tranform
+        dc_transform: Optional[Transform],
```
MingshiPeng, line 217: @rchowell same here, can you help validate my comments here?

deltacat/daft/daft_scan.py
pdames: This should be fine for now (at worst, we lose track of which fields mapped to each subschema, but this only becomes relevant if we need to rehydrate a DeltaCAT schema losslessly from a downstream written by Daft and is not an immediate concern).

deltacat/daft/model.py
rchowell: Yes, related to all three feature requests could please open Daft issue with any additional context? Feel free to group these on one issue since these are all related. I can then work on a proposal to get you some nice APIs. Thanks!

requirements.txt
```diff
@@ -14,6 +14,7 @@ pandas == 1.3.5
 # TestCompactionSession::test_compact_partition_when_incremental_then_rcf_stats_accurate to fail
 # due to input_inflation exceeding 1e-5
 pyarrow == 17.0.0
+pyarrow-stubs==17.0.0.0
```
pdames, line 17: Do we need this?

MingshiPeng: @pdames - pyarrow lib itself doesn't expose any type information, which makes the dev work super annoying and voids all the compile time check benefits. Adding the stubs helps surfacing a user-friendly python interface for pyarrow.

Please let me know if this addresses your concern.

pdames: Hmm... I think this comes down to whether or not your IDE supports, or can be configured to support, autocomplete against Cython .pyx, .pxd, and .pxi files (I haven't bumped into this issue with PyCharm). If you require it for your dev setup, I'd recommended just creating a personal dev requirements.txt in your workspace like `requirements-mingshi.txt` that imports `dev-requirements.txt` and adds these stubs, but wouldn't currently add it to `requirements.txt` (`pyarrow-stubs` looks like a personal project for now).

This may also go into creating better IDE-specific setup instructions for developers in `README-development.md`.

deltacat/daft/model.py
MingshiPeng: Created an Issue on Daft side: https://github.com/Eventual-Inc/Daft/issues/4228

requirements.txt
MingshiPeng: Yes I’m using VSCode which doesn't seem to have good support for Cython files. I see using PyCharm automatically gets the type information even without the stubs. 

I didn’t realize that the `pyarrow-stubs` was a personal project. Have removed the dependency in the latest commit.

pdames (approved):

PR #535, opened 2025-04-23
chore: add Manifest.from_json() method

MingshiPeng:
## Summary

Implementing `Manifest.from_json()` method to support deserialization from Manifest json string.

## Rationale

This is needed to support parsing Manifest that is returned from Glue partition.

## Testing
- Added unit test and passed

### DIFF
--- a/deltacat/storage/model/manifest.py
+++ b/deltacat/storage/model/manifest.py
@@ -11,6 +11,8 @@ from deltacat import logs
 
 from deltacat.storage.model.schema import FieldLocator
 
+import json
+
 logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 
@@ -193,6 +195,20 @@ class Manifest(dict):
         return manifest
 
     @staticmethod
+    def from_json(json_string: str) -> Manifest:
+        parsed_dict = json.loads(json_string)
+        return Manifest.of(
+            entries=ManifestEntryList.of(
+                [
+                    ManifestEntry.from_dict(entry)
+                    for entry in parsed_dict.get("entries", [])
+                ]
+            ),
+            author=ManifestAuthor.from_dict(parsed_dict.get("author", {})),
+            uuid=parsed_dict.get("id"),
+        )
+
+    @staticmethod
     def merge_manifests(
         manifests: List[Manifest], author: Optional[ManifestAuthor] = None
     ) -> Manifest:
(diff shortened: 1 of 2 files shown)


anshumankomawar (approved):
LGTM!

deltacat/storage/model/manifest.py
```diff
@@ -358,6 +389,16 @@ def from_s3_obj_url(
         manifest_entry = ManifestEntry.of(url, manifest_entry_meta)
         return manifest_entry
 
+    @staticmethod
+    def from_dict(obj: dict) -> ManifestEntry:
+        return ManifestEntry.of(
+            url=obj.get("url"),
+            uri=obj.get("uri"),
+            meta=ManifestMeta.from_dict(obj.get("meta", {})),
```
pdames, line 397: This behavior isn't fully consistent with the default constructor behavior of omitting the `meta` key completely if it doesn't exist (this will instead set the `meta` key to an empty dictionary).

```suggestion
            meta=ManifestMeta.from_dict(obj["meta"]) if "meta" in obj else None,
```

deltacat/storage/model/manifest.py
```diff
@@ -192,6 +194,20 @@ def of(
...
+    def from_json(json_string: str) -> Manifest:
+        parsed_dict = json.loads(json_string)
+        return Manifest.of(
+            entries=ManifestEntryList.of(
+                [
+                    ManifestEntry.from_dict(entry)
+                    for entry in parsed_dict.get("entries", [])
+                ]
+            ),
+            author=ManifestAuthor.from_dict(parsed_dict.get("author", {})),
```
pdames, line 207: This behavior isn't fully consistent with the default constructor behavior of omitting the `author` key completely if it doesn't exist (this will instead set the `author` key to an empty dictionary).



```suggestion
            author=ManifestAuthor.from_dict(parsed_dict["author"]) if "author" in parsed_dict else None,
```

pdames (changes requested):
Recommended a couple minor changes - otherwise this looks good to merge.

deltacat/storage/model/manifest.py
MingshiPeng: addressed

deltacat/storage/model/manifest.py
MingshiPeng: addressed

pdames (approved):

PR #536, opened 2025-04-23
Fix DQ issue due to inconsistent grouping of bucketed records

raghumdani:
## Summary

This change fixes a DQ issue that was caused due to inconsistent grouping of bucketed records. The pyarrow `groupby` method *almost* always returns the keys (on which the records were grouped) in the same order as they appear in the input. Algorithm relied on this invariant for correctness. However, we missed adding an assertion for the same. In rare scenarios, the order of the keys changed which led to incorrect assignment of hash buckets to respective record groups and caused DQ issue in the tables.

The ordering of keys in the output will be consistent with the input if pyarrow uses `exactly one thread` to execute the `groupby` operation. However, in some `unknown` scenarios, pyarrow could be using multi-threading to speed up the grouping operation which has resulted in inconsistent order of the keys on which the records were grouped by.  

## Rationale

This change helps stop the further bleeding by ensuring the data is written correctly from now on. 

## Changes

This change ensures that invariant is true by explicitly sorting the output of pyarrow groupby operation. This comes at an added cost which is necessary for correctness. 

## Impact

Fixes the DQ issue. 

## Testing

I downloaded a specific file from our internal catalog on which this issue occurs consistently. Then, tested it by running compaction against real tables to verify the fix. An attempt to reproduce this issue on hand-crafted files did not trigger pyarrow fallback to `multi-threading`. Hence, we will be adding an integration test case using one of the file on which this issue occurred in production.  

## Regression Risk

None

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/compute/compactor_v2/utils/primary_key_index.py
+++ b/deltacat/compute/compactor_v2/utils/primary_key_index.py
@@ -79,7 +79,6 @@ def _append_table_by_hash_bucket(
     )
 
     hb_pk_grouped_by = hb_pk_grouped_by.sort_by(sc._HASH_BUCKET_IDX_COLUMN_NAME)
-
     group_count_array = hb_pk_grouped_by[f"{sc._HASH_BUCKET_IDX_COLUMN_NAME}_count"]
     hb_group_array = hb_pk_grouped_by[sc._HASH_BUCKET_IDX_COLUMN_NAME]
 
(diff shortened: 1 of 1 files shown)


ekaschaw (approved):

pfaraone (approved):
LGTM! Thanks for this fix

raghumdani:
There is a test case in arrow codebase which ensures the ordering is preserved with `use_threads=False`. However, this argument isn't available on pyarrow 12.0.1 which we use today. 

https://github.com/apache/arrow/blob/520ae44272d491bbb52eb3c9b84864ed7088f11a/python/pyarrow/tests/test_exec_plan.py#L326

PR #543, opened 2025-05-02
chore: upgrade daft version

MingshiPeng:
## Summary

bump up daft version to 0.4.13 which is compatible with AL2.

## Test
- e2e Iceberg read example works


anshumankomawar (approved):
LGTM!

PR #544, opened 2025-05-04
bump deltacat version to 2.0.0b9

MingshiPeng:
## Summary

bump deltacat version to 2.0.0b9 so daft version bump to 0.4.13 can be included.


MingshiPeng:
merge to unblock myself as the process to publish deltacat new version to Amazon internal will take some time

PR #552, opened 2025-05-20
[Converter] Equality Deletes Conversion with Enforce Primary Key Uniqueness support

Zyiqin-Miranda:
## Summary

This PR adds support for converting equality deletes while enforcing primary key uniqueness. Test cases added for correctness check.

How it works?
For Iceberg table, within each partition:
1. Fetch all equality deletes and data files that belong to this partition.
2. Group equality deletes and applicable data files together. By applicable, according to Iceberg spec, equality deletes can apply to all data files having strictly smaller sequence number. 
3. Sort equality delete files and data files based off sequence number. When downloaded, data files table get additional file_path and ordered record index column.
4. After converting, we produce position delete table **A** that delete records based off equality deletes. Data table **a** contains all remaining records.
5. Append to Data table **a** for any data files that don't got applied by equality deletes and get data table **b**.
6. Drop duplicates on Data table **b**, append unique record index based off final table to ensure deterministic record get kept. Produced position delete table **B** that represent duplicate primary key meant to be deleted.
7. Final output will be position delete files containing position delete **A + B**

## Rationale

## Changes

## Impact

## Testing

## Regression Risk

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/compute/converter/pyiceberg/update_snapshot_overrides.py
+++ b/deltacat/compute/converter/pyiceberg/update_snapshot_overrides.py
@@ -1,4 +1,5 @@
-from typing import Optional, List
+from typing import List, Dict
+from collections import defaultdict
 import uuid
 from pyiceberg.table.snapshots import (
     Operation,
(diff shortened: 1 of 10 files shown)


deltacat/tests/compute/converter/test_convert_session.py
```diff
@@ -639,3 +645,195 @@ def test_converter_pos_delete_multiple_identifier_fields_success(
 
     # Assert elements are same disregard ordering in list
     assert sorted(pk_combined_res) == sorted(expected_result_tuple_list)
+
+
```
pdames, line 649: Can we add a test case to ensure that the sort is stable when multiple files share the same sequence number?

deltacat/compute/converter/steps/convert.py
```diff
@@ -160,23 +180,51 @@ def convert(convert_input: ConvertInput):
...
+
+        position_delete_table = position_delete_table.drop(
+            [sc._IDENTIFIER_COLUMNS_HASH_COLUMN_NAME]
+        )
+        assert len(position_delete_table) + len(remaining_data_table) == len(
+            data_file_table
+        ), (
+            f"Length of data file table remaining plus length of pos delete table should match origin data file table length"
+            f"But got {len(position_delete_table)} pos delete, {len(remaining_data_table)} equality delete, "
+            f"doesn't equal to original data table length: {len(data_file_table)}"
```
pdames, line 224:
```suggestion
            f"Expected undeleted data file record count plus length of pos deletes to match original data file record count of {len(data_file_table)}, "
            f"but found {len(position_delete_table)} pos deletes + {len(remaining_data_table)} equality deletes."
```

deltacat/compute/converter/steps/convert.py
```diff
@@ -160,23 +180,51 @@ def convert(convert_input: ConvertInput):
 
 
 def get_additional_applicable_data_files(all_data_files, data_files_downloaded):
-    data_file_to_dedupe = all_data_files
+    data_file_to_dedupe = []
+    assert len(set(all_data_files)) >= len(set(data_files_downloaded)), (
+        f"Length of all data files list: {len(set(all_data_files))} should be greater than"
+        f"Length of corresponding data files list: {len(set(data_files_downloaded))}"
```
pdames, line 186:
```suggestion
        f"Length of all data files ({len(set(all_data_files))}) should never be less than "
        f"the length of candidate equality delete data files ({len(set(data_files_downloaded))})"
```

deltacat/compute/converter/pyiceberg/update_snapshot_overrides.py
```diff
@@ -237,15 +258,23 @@ def _deleted_entries(self) -> List[ManifestEntry]:
...
+        with replace_delete_files_override(
+            tx.update_snapshot()
+        ) as replace_delete_snapshot:
             if new_position_delete_files:
                 for data_file in new_position_delete_files:
-                    append_snapshot.append_data_file(data_file)
+                    replace_delete_snapshot.append_data_file(data_file)
+            if to_be_deleted_files:
+                for delete_file in to_be_deleted_files:
+                    print(f"debug_delete_file_snapshot:{delete_file}")
```
pdames, line 279: Should this be removed (or converted to a debug log)?

deltacat/compute/converter/pyiceberg/update_snapshot_overrides.py
```diff
@@ -114,42 +158,19 @@ def _get_entries(manifest: ManifestFile) -> List[ManifestEntry]:
...
+        if iceberg_table.metadata.name_mapping() is None:
+            tx.set_properties(
+                **{
+                    "schema.name-mapping.default": tx.table_metadata.schema().name_mapping.model_dump_json()
+                }
+            )
+        with append_delete_files_override(tx.update_snapshot()) as append_snapshot:
+            if new_position_delete_files:
+                for data_file in new_position_delete_files:
+                    print(f"debug_append_snapshot_data_file:{data_file}")
```
pdames, line 172: Should this be removed (or converted to a debug log)?

pdames (approved):
LGTM! Just a few minor comments to address, then feel free to merge. Cool to see Iceberg equality delete support added on top of the existing ability to enforce unique primary keys!

deltacat/tests/compute/converter/test_convert_session.py
Zyiqin-Miranda: Good catch! Will add a secondary sort key based off file path, I noticed the files Spark write out have some sort of prefix in the file name that represents the order of this file within same snapshot, e.g: s3://xxxx/partitionKey=8/primaryKey_bucket=83/**00000-58**-3229fae0-4316-4ea4-9e09-367a5d1b96f9-00001.parquet, the bolded part is observed to be the order of files added within this transaction.

Zyiqin-Miranda:
Thanks @pdames, could you review the changes in this [commit](https://github.com/ray-project/deltacat/pull/552/commits/7cda3b4b119c9f66d89b0afcf96c54bb65c7c07b)? Passing filesystem in case the `path` doesn't contain info to infer which filesystem it is but user explicitly pass in a filesystem.

pdames (approved):
Good catch on the missing user-specified filesystem assignment in pyarrow utils!

PR #564, opened 2025-07-03
[Converter] Added memory base buffer; Code clean-up

Zyiqin-Miranda:
## Summary

This PR adjust the memory resource estimation for converter() remote tasks based off experiments. Also contains code clean-up regarding determining which snapshot type to commit. 

All existing tests passing.
## Testing

## Regression Risk

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/compute/converter/utils/convert_task_options.py
+++ b/deltacat/compute/converter/utils/convert_task_options.py
@@ -9,7 +9,7 @@ XXHASH_BYTE_PER_RECORD = 8
 MEMORY_BUFFER_RATE = 2
 # Worst case 2 as no duplicates exists across all pk
 PYARROW_AGGREGATE_MEMORY_MULTIPLIER = 2
-# Observed base memory usage at the beginning of each process
+# Observed base memory usage at the beginning of each worker process
 BASE_MEMORY_BUFFER = 0.3 * 1024 * 1024 * 1024
 
 
--- a/deltacat/compute/converter/utils/convert_task_options.py
+++ b/deltacat/compute/converter/utils/convert_task_options.py
@@ -3,13 +3,14 @@ from deltacat.exceptions import RetryableError
 from pyiceberg.manifest import DataFile
 from deltacat.compute.converter.model.convert_input_files import ConvertInputFiles
 
-AVERAGE_FILE_PATH_COLUMN_SIZE_BYTES = 80
+AVERAGE_FILE_PATH_COLUMN_SIZE_BYTES = 160
 AVERAGE_POS_COLUMN_SIZE_BYTES = 4
 XXHASH_BYTE_PER_RECORD = 8
 MEMORY_BUFFER_RATE = 2
-# TODO: Add audit info to check this number in practice
 # Worst case 2 as no duplicates exists across all pk
 PYARROW_AGGREGATE_MEMORY_MULTIPLIER = 2
+# Observed base memory usage at the beginning of each process
+BASE_MEMORY_BUFFER = 0.3 * 1024 * 1024 * 1024
 
 
 def estimate_fixed_hash_columns(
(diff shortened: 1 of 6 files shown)


deltacat/compute/converter/utils/converter_session_utils.py
```diff
@@ -121,3 +123,53 @@ def sort_data_files_maintaining_order(data_files: DataFileList) -> DataFileList:
...
+class SnapshotType(Enum):
+    """Enumeration of possible snapshot types."""
+
+    NONE = "none"
+    APPEND = "append"
+    REPLACE = "replace"
+    DELETE = "delete"
+
+
+def _get_snapshot_description(
```
pdames, line 137: Since this isn't describing the snapshot but instead the action that we'll take based on the snapshot, maybe something like:
```suggestion
def _get_snapshot_action_description(
```

deltacat/compute/converter/utils/convert_task_options.py
```diff
@@ -3,13 +3,14 @@
...
 
-AVERAGE_FILE_PATH_COLUMN_SIZE_BYTES = 80
+AVERAGE_FILE_PATH_COLUMN_SIZE_BYTES = 160
 AVERAGE_POS_COLUMN_SIZE_BYTES = 4
 XXHASH_BYTE_PER_RECORD = 8
 MEMORY_BUFFER_RATE = 2
-# TODO: Add audit info to check this number in practice
 # Worst case 2 as no duplicates exists across all pk
 PYARROW_AGGREGATE_MEMORY_MULTIPLIER = 2
+# Observed base memory usage at the beginning of each process
```
pdames, line 12:
```suggestion
# Observed base memory usage at the beginning of each worker process
```

deltacat/compute/converter/utils/convert_task_options.py
```diff
@@ -3,13 +3,14 @@
...
+BASE_MEMORY_BUFFER = 0.3 * 1024 * 1024 * 1024
```
pdames, line 13: Should we also document that we need at least (300MB + some buffer) of free memory to run the converter? Maybe just the docs for this function should state that the converter requires a node with at least 512MB of free memory?

deltacat/utils/resources.py
```diff
@@ -169,6 +170,12 @@ def get_current_process_peak_memory_usage_in_bytes():
         return psutil.Process().memory_info().peak_wset
 
 
+def get_current_process_current_memory_usage_in_bytes():
+    process = psutil.Process(os.getpid())
+    mem_info = process.memory_info()
+    return mem_info
+
+
```
pdames, line 178: Since this isn't currently used can we remove it? Or if we're going to use it for a profiler in a follow-up PR we can keep it for now.

pdames (approved):
Looks good overall! Just a few minor comments.

deltacat/compute/converter/utils/convert_task_options.py
Zyiqin-Miranda: Thanks @pdames, added as `**Memory Requirements:**` doc string in `converter_session()` function. Haven't finalize the naming of al params like maybe `primary key` to `merge key` yet, so doc string is still using the current version of converter. 
Additionally, tested locally that even with just test setup, the process memory is using around 300MB to begin with. Just FYI.

Zyiqin-Miranda:
Thanks @pdames, addressed comments and will follow up soon with two PRs: 
1. For make all private functions start with underscore with additional code clean-ups
2. Try-catch OOM failures and retry with additional resources, probably will test with removing custom resources like "coverter_task": 1 to use just CPU resources to make cluster configurations more generic.

Zyiqin-Miranda:
Also, removed the `comment-on-alert: true` when performance testing triggered alert since write permission in general shouldn't be granted with fork, from [github actions security doc](https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication): 
>`You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access`
So probably should setup another way of alerting on action workflow benchmark testing regression like email or slack notification like other repos do. But removing this comment option in CI for now to unblock contribution merging.

PR #566, opened 2025-07-14
Added partition-level predicate pushdown for Iceberg

rnapark:
## Summary

Added partition-level predicate pushdown for Iceberg.

## Rationale

To-do

## Changes

Added a DeltaCAT to PyIceberg expression visitor to utilize PyIceberg APIs for partition-level predicate pushdown, along with corresponding test suites. Updated IcebergScanPlanner to use said APIs if applicable.

## Impact

## Testing

Unit tests, integration tests, etc all pass.

## Regression Risk

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/experimental/storage/iceberg/iceberg_scan_planner.py
+++ b/deltacat/experimental/storage/iceberg/iceberg_scan_planner.py
@@ -1,16 +1,50 @@
-from typing import Optional
+import logging
+from typing import Optional, Any, Set
 
 from pyiceberg.catalog import Catalog
-from deltacat.storage.model.scan.push_down import Pushdown
+from pyiceberg.table import Table
+import deltacat.logs as logs
+
+from deltacat.storage.model.scan.push_down import Pushdown, PartitionFilter
 from deltacat.storage.model.scan.scan_plan import ScanPlan
 from deltacat.storage.model.scan.scan_task import FileScanTask, DataFile
 from deltacat.storage.util.scan_planner import ScanPlanner
 from deltacat.experimental.storage.iceberg.impl import _try_load_iceberg_table
+from deltacat.experimental.storage.iceberg.visitor import IcebergExpressionVisitor
+
+# Initialize DeltaCAT logger
+logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
 
 
 class IcebergScanPlanner(ScanPlanner):
     def __init__(self, catalog: Catalog):
         self.catalog = catalog
+        self.expression_visitor = IcebergExpressionVisitor()
+
+    @classmethod
+    def _collect_filter_fields(cls, expr: Any) -> Set[str]:
+        """
+        Collects all field names referenced in the filter expression.
+
+        Args:
+            expr: The expression to analyze
+
+        Returns:
+            Set of field names referenced in the expression
+        """
+        fields = set()
+        if hasattr(expr, "field"):
+            fields.add(expr.field)
+        if hasattr(expr, "left"):
+            fields.update(cls._collect_filter_fields(expr.left))
+        if hasattr(expr, "right"):
+            fields.update(cls._collect_filter_fields(expr.right))
+        if hasattr(expr, "expr"):
+            fields.update(cls._collect_filter_fields(expr.expr))
+        if hasattr(expr, "values"):
+            for value in expr.values:
+                fields.update(cls._collect_filter_fields(value))
+        return fields
 
     def create_scan_plan(
         self,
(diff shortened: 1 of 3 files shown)


deltacat/experimental/storage/iceberg/visitor.py
```diff
@@ -0,0 +1,107 @@
...
+    Visitor that translates DeltaCAT expressions to PyIceberg expressions.
+    Similar to GlueExpressionVisitor but outputs PyIceberg expressions.
+    """
+
+    def visit_reference(self, expr: Reference, context=None) -> str:
+        return expr.field
+
+    def visit_literal(self, expr: Literal, context=None) -> Any:
+        # Convert PyArrow scalar to Python native type
+        return expr.value.as_py() if hasattr(expr.value, "as_py") else expr.value
```
pdames, line 35: Instead of just looking for an `as_py()` method, it may be safer to just be more explicit about the type we expect the expression type to be:
```suggestion
        return expr.value.as_py() if isinstance(expr.value, pyarrow.Scalar) else expr.value
```

deltacat/experimental/storage/iceberg/visitor.py
```diff
@@ -0,0 +1,107 @@
...
+    GreaterThan,
+    GreaterThanOrEqual,
+    LessThan,
+    LessThanOrEqual,
+    IsNull,
+    In,
+)
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
```
pdames, line 21: For consistency across the project, we should always configure and use the default DeltaCAT system logger:

```suggestion
import deltacat.logs as logs

# Initialize DeltaCAT logger
logger = logs.configure_deltacat_logger(logging.getLogger(__name__))
```

deltacat/experimental/storage/iceberg/iceberg_scan_planner.py
```diff
@@ -21,8 +24,58 @@ def create_scan_plan(
...
+        # Get the partition spec
+        partition_spec = iceberg_table.spec()
+
+        # Check if the table is partitioned
+        is_partitioned = len(partition_spec.fields) > 0
+
+        scan = iceberg_table.scan()
+        if is_partitioned:
+            if pushdown and pushdown.partition_filter:
+                print("There is a partition filter")
```
pdames, line 39: Let's configure the default DeltaCAT system logger and replace this debug `print` statement with a `logger.info` log entry noting that we discovered and are converting the partition filter (ideally, we should at least note the field count and fields discovered in the partition filter as well... I wouldn't recommend recording the values since they may contain sensitive information).

pdames (changes requested):
Looks pretty good overall! Should be good to merge after just a few minor changes (e.g., updating code to use the DeltaCAT system logger).

deltacat/experimental/storage/iceberg/iceberg_scan_planner.py
```diff
@@ -21,8 +24,58 @@ def create_scan_plan(
...
+                scan = scan.filter(iceberg_expression)
+
         file_scan_tasks = []
-        # TODO: implement predicate pushdown to Iceberg
-        for scan_task in iceberg_table.scan().plan_files():
+        for scan_task in scan.plan_files():
             file_scan_tasks.append(FileScanTask([DataFile(scan_task.file.file_path)]))
         return ScanPlan(file_scan_tasks)
+
+    def _convert_partition_filter(self, table: Table, partition_filter):
```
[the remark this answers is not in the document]
rnapark: Addressed!

deltacat/experimental/storage/iceberg/iceberg_scan_planner.py
pdames: Just including any known field/column names in the log statement should be fine. I was just noting that, in any other areas of code where we also know the corresponding field value to filter on, when we should prefer to omit the value (like if we have a partition field of `name` and know that a corresponding partition filter predicate is `name=="Bob"`, then we can log that we're filtering on the `name` field but shouldn't note that we're looking for `"Bob"` specifically).

deltacat/experimental/storage/iceberg/iceberg_scan_planner.py
```diff
@@ -21,8 +24,58 @@ def create_scan_plan(
...
+        """
+        Convert DeltaCAT partition filter to PyIceberg expression,
+        validating that only partition columns are referenced.
+        """
+        partition_cols = set(field.name for field in table.spec().fields)
+
+        # Validate that the filter only references partition columns
+        def validate_partition_references(expr):
```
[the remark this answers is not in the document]
rnapark: Addressed!

pdames (approved):

MingshiPeng:
Have you done e2e testing and assert the partition predicate pushdown is working? E.g. if a Table has 2 partitions, after applying the partition predicate pushdown, only files corresponding to 1 partition is returned as part of the scan planning.

Otherwise LGTM.

rnapark:
Tested e2e, successful.

MingshiPeng (approved):

PR #569, opened 2025-08-17
Rudimentary catalog init refactoring to use a config file

rnapark:
## Summary

Added using config.yaml file as a second choice option to initialize DeltaCAT catalog instead of relying on .init() call passing in parameters explicitly. 

## Rationale

To reduce tasks for users and expose less internal mechanisms.
## Changes

- Created config.yaml parser for initialize()
- Wrote according test
## Impact

Only adds possible functionality, backwards-compatible. 
## Testing

make test

## Checklist

## Additional Notes

Next steps will be to create enums and expand existing catalog types. ex. Iceberg

### DIFF
--- a/deltacat/catalog/model/catalog.py
+++ b/deltacat/catalog/model/catalog.py
@@ -231,7 +231,7 @@ def init(
     if catalogs and config_path is None:
         # Only dump if the catalogs are config-backed
         try:
-            dump_catalogs_to_yaml(catalogs)
+            _dump_catalogs_to_yaml(catalogs)
         except TypeError:
             logger.debug(
                 "Skipping dumping catalogs to YAML: non-CatalogProperties inner"
@@ -432,7 +432,7 @@ def put_catalog(
         existing_catalogs[name] = catalog
 
         # Now write back full merged dictionary
-        dump_catalogs_to_yaml(existing_catalogs, single_if_default=False)
+        _dump_catalogs_to_yaml(existing_catalogs, single_if_default=False)
 
     except Exception as e:
         logger.warning(f"Failed to persist catalog {name} to config file: {e}")
@@ -440,7 +440,7 @@ def put_catalog(
     return catalog
 
 
-def dump_catalogs_to_yaml(
+def _dump_catalogs_to_yaml(
     catalogs: Union[
         Dict[str, Union[Catalog, CatalogProperties]], Catalog, CatalogProperties
     ],
--- a/deltacat/catalog/model/catalog.py
+++ b/deltacat/catalog/model/catalog.py
@@ -2,16 +2,19 @@
 from __future__ import annotations
 
 import logging
+from pathlib import Path
 from types import ModuleType
 
 from typing import Any, Dict, List, Optional, Union
 from functools import partial
 import ray
+import yaml
 
 from deltacat import logs
 from deltacat.catalog.main import impl as dcat
 from deltacat.catalog.model.properties import CatalogProperties
-from deltacat.constants import DEFAULT_CATALOG
+from deltacat.constants import DEFAULT_CATALOG, DELTACAT_CONFIG_PATH
+from deltacat.utils.config_loader import load_catalog_configs_from_yaml
 
 all_catalogs: Optional[ray.actor.ActorHandle] = None
 
@@ -182,6 +185,7 @@ def init(
     ray_init_args: Dict[str, Any] = {},
     *,
     force=False,
+    config_path: Optional[str] = None,
 ) -> Optional[ray.runtime.BaseContext]:
     """
     Initialize DeltaCAT catalogs.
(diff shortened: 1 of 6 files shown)


deltacat/catalog/model/catalog.py
```diff
@@ -202,6 +203,10 @@ def init(
         logger.warning("DeltaCAT already initialized.")
         return None
 
+    # If no catalogs provided but a config_path exists, create a Catalog
+    if not catalogs and config_path is not None:
+        catalogs = Catalog(config_path=config_path)
```
pdames, line 208: To align with the ability to create multiple named catalogs at initialization time, we should also serialize and deserialize a dictionary of name/CatalogProperties key/value pairs into this config file instead of a single unnamed CatalogProperties instance.

deltacat/catalog/model/catalog.py
```diff
@@ -202,6 +203,10 @@ def init(
         logger.warning("DeltaCAT already initialized.")
         return None
 
+    # If no catalogs provided but a config_path exists, create a Catalog
+    if not catalogs and config_path is not None:
```
pdames, line 207: What do we do if a user specifies catalogs AND a config path? I think the easiest initial answer is to just raise a `ValueError`.

deltacat/catalog/model/catalog.py
```diff
@@ -202,6 +203,10 @@ def init(
...
+        catalogs = Catalog(config_path=config_path)
+
```
pdames, line 209: Can we also add some other options here like (1) if catalogs are specified and config_path is not, then create a dictionary to write to a default Path defined by a new `DELTACAT_CONFIG_PATH` env_string in `constants.py` that defaults to `~/.deltacat/config.yaml` with a dictionary value of `{name: catalog.config for name, catalog in catalogs.items()}` (2) if neither catalogs or config path are specified, then load the catalog config from `DELTACAT_CONFIG_PATH` if it exists.

To keep the catalog written to disk consistent with what we have in memory for our current working session, we should also ensure that `put_catalog` updates our existing catalog config at `DELTACAT_CONFIG_PATH` if it exists, and creates it if it does not.

pdames (changes requested):
Looks like a good start! I think this should be ready to merge after making the proposed improvements (esp. ensuring that we support persisting a dictionary of named catalog properties rather than just a single CatalogProperties instance).

deltacat/catalog/model/catalog.py
rnapark: ^ I understand that, but Catalog.initialize() currently returns a single CatalogProperties object, shall I have it return the dict[str, CatalogProperties] created by the config_loader(in the case of multiple configs), or return the "default" catalog's CatalogProperties only?

deltacat/catalog/model/catalog.py
pdames: Merge sounds right, but I think we should also have some handling here to make sure that, after all unit tests are run locally, we don't wind up with a ton of deleted temporary directories in our config file. I think many tests call `pop_catalog` explicitly on teardown, so adding handling there to make sure that we also remove catalogs from config should help, but we need something easier to ignore 100% of test catalogs.

deltacat/catalog/model/catalog.py
```diff
@@ -351,35 +387,116 @@ def put_catalog(
...
+    """
+    from pathlib import Path
+
+    def props_to_dict(inner: Any) -> dict:
+        if isinstance(inner, CatalogProperties):
+            return {
+                "root": getattr(inner, "root", None),
+                "filesystem": None,
+                "storage": None,
+            }
```
pdames, line 466: For consistency with other model serialization methods, can we instead (1) create a new `to_serializable` method on CatalogProperties that returns this dictionary and (2) just call `return inner.to_serializable()` here instead?

deltacat/tests/catalog/test_catalogs.py
```diff
@@ -152,6 +153,49 @@ def test_init_with_default_catalog_name(self, reset_catalogs):
         assert default_catalog.impl == MockCatalogImpl
         assert default_catalog.inner["kwargs"]["id"] == 2
 
+    def test_init_with_config_yaml(self, tmp_path, reset_catalogs):
```
pdames, line 156: Can we also add tests for the modified behavior of `init`, `put_catalog`, and `pop_catalog`?

deltacat/utils/config_loader.py
```diff
@@ -0,0 +1,47 @@
...
+            f"Expected a dict, got {type(config_data)}"
+        )
+
+    # Case 1: single unnamed config
+    # e.g. {"type": "iceberg", "uri": "...", "warehouse": "prod"}
+    if all(
+        isinstance(v, (str, int, float, bool, type(None))) or isinstance(v, list)
+        for v in config_data.values()
+    ):
+        return {"default": CatalogProperties(**config_data)}
```
pdames, line 36: For consistency with other model deserializers and centralize model validation/deserialization logic, can we consolidate CatalogProperties deserialization validation and logic inside a static `CatalogProperties.from_serializable(config)` method?

pdames (changes requested):
Looks pretty good overall! I think this will be ready to merge after a few minor updates.

deltacat/catalog/model/catalog.py
renapark-duke: Addressed!

deltacat/utils/config_loader.py
renapark-duke: Addressed!

deltacat/catalog/model/catalog.py
```diff
@@ -351,35 +387,116 @@ def put_catalog(
...
-                raise
-            # If catalog doesn't exist, continue normally
-            pass
-
-    # Add the catalog (which may overwrite existing if fail_if_exists=False)
-    ray.get(all_catalogs.put.remote(name, catalog, default))
     return catalog
+
+
+def dump_catalogs_to_yaml(
```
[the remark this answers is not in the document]
renapark-duke: Addressed!

deltacat/catalog/model/catalog.py
```diff
@@ -351,35 +387,116 @@ def put_catalog(
...
+) -> None:
+    """
+    Write Catalog configs (Catalogs or CatalogProperties) to YAML.
+
+    - Supports Dict[str, Catalog], Dict[str, CatalogProperties],
+      a single Catalog, or a single CatalogProperties.
+    - If single_if_default=True and only one entry ("default"),
+      emit a flat config instead of mapping.
+    """
+    from pathlib import Path
```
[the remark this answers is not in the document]
rnapark: Addressed!

deltacat/catalog/model/catalog.py
rnapark: Addressed!

deltacat/catalog/model/catalog.py
```diff
@@ -351,35 +387,116 @@ def put_catalog(
...
+                )
+            except ValueError as e:
+                if "not found" not in str(e):
+                    raise
+                # Doesn't exist → safe to add
+
+        # Register in memory
+        ray.get(all_catalogs.put.remote(name, catalog, default))
+
+    # --- Persist to disk (merge behavior) ---
```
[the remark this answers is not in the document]
rnapark: Addressed!

deltacat/catalog/model/catalog.py
rnapark: Created pytest fixture config path to isolate the test config files.

rnapark:
I'm working on more put/pop_catalog tests, but running into some problems with the ray actor carrying over(?) catalogs created from other tests. I confirmed the pytest fixture I created for the config path is working, and not sure what else to isolate.

pdames (approved):

PR #575, opened 2025-09-20
Add Support for Unordered ADD Write Mode, Portable Manifests, Partitioned Revision & Transaction Logs

pdames:
## Summary

This PR adds support for the following new features:

1. **Unordered ADD Write Mode and Deltas**: This is the new default write mode which writes unordered ADD Deltas with randomly assigned INT64 stream positions, while reserving the INT32 stream position space of every table partition for ordered APPEND deltas (the default delta type until now). The primary benefit of ADD deltas is that they don't require serializable isolation on the partition that they're writing to, which allows an arbitrarily large number of ADD deltas to be written concurrently to any table partition. To assist with workflows that prefer to work with recently written deltas `dc.write`/`dc.write_to_table` has also been updated to return the List of deltas written (i.e., previously you could always get the latest APPEND delta written using `get_latest_delta`, but this doesn't work for unordered ADD deltas).

2. **Portable Manifests**: Manifest entries have been updated to always have their urls persisted relative to their catalog root. To support this, `dc.copy` has been updated to perform deep copies by default, and `from_manifest_table` has been promoted to a catalog-level API (since it needs context about the catalog associated with a schemaless table to correctly resolve its file locations).

3. **Partitioned Revision & Transaction Logs**: Improves scale and efficiency (esp. for cloud storage like S3) of working with large transaction and metafile revision counts, by (1) partitioning transaction logs using year/month/day/hour/minute UTC directories based on transaction start time, (2) partitioning revisions in millions/thousands partitions (defaulted to 1000 per partition to ensure that S3, GCP, and Azure can return all revisions in a single list request).

4. **Catalog Version Files**: When a DeltaCAT catalog is first initialized, an empty version file is now created in the
catalog root directory that tracks which version of DeltaCAT wrote to the catalog, and when it started writing to the catalog.  The file format is `${CATALOG_ROOT}/version/<deltacat_version>.<starting_from>` where `<deltacat_version>` is the version of DeltaCAT that wrote the file, like `2.0.0`, and  `<starting_from>` is the epoch time in nanoseconds when that version first initialized the catalog with write access, like `1758178659283916000`.

It also resolves the following issues:

1. **Concurrent Namespace Creation Conflicts During Table Writes**: `dc.write` would previously autocreate the table's parent namespace if it didn't exist. However, this can result in confusing concurrent conflict modification errors if multiple new tables are written to a new namespace concurrently. This is especially true for initial default namespace creation since, from the user's perspective, it looks like they're writing to separate tables without any namespace. To fix this, a catalog's default namespace is now created during CatalogProperties initialization, and `dc.write` will raise a `NamespaceNotFound` error for missing parent namespaces by default, unless `auto_create_namespace=True`. 

2. **https://github.com/ray-project/deltacat/issues/500**: The native catalog's `CatalogProperties` instance is fully serializable by Ray CloudPickle, so the Catalog serializer registered with Ray now passes it directly through to each Catalog it constructs instead of requiring the Catalog to always reconstruct its child `CatalogProperties` instance.

FYI @IvanPartsunev - curious to get your feedback here as well.

## Checklist


deltacat/tests/compute/test_janitor.py
```diff
@@ -130,21 +137,27 @@ def test_janitor_handles_multiple_timed_out_transactions(self, temp_dir):
         os.makedirs(failed_txn_dir, exist_ok=True)
 
         # Create multiple timed-out transaction logs
-        txn_ids = ["txn_one", "txn_two", "txn_three"]
         start_time = time.time_ns() - 1_000_000_000  # 1 second in the past
```
Zyiqin-Miranda, line 140: Do we need to change this value `1_000_000_000` to `500_000_000`as well? Maybe use a constant value to make sure it's consistent instead.

deltacat/tests/catalog/test_default_catalog_impl.py
```diff
@@ -1841,30 +2177,283 @@ def test_append_compaction_with_schema_evolution(self):
...
+            obj for obj in append_objects if Metafile.get_class(obj) == Delta
         ]
+        assert len(append_deltas) == 1, "Should have exactly 1 APPEND delta"
+        append_stream_position = append_deltas[0].stream_position
+
+        # Verify ADD uses random stream position (large number)
         assert (
-            len(explicit_emails) == 2
-        ), "Last two rows should have explicit email values"
+            add_stream_position > 10000
```
Zyiqin-Miranda, line 2419: How about just assert Add Delta stream position is > UNSIGNED_INT32_MAX_VALUE?

Zyiqin-Miranda (approved):

rnapark (approved):
I hadn't realized that unordered and ordered deltas could coexist. Sounds exciting for performance!!

deltacat/storage/model/manifest.py
```diff
@@ -575,11 +586,11 @@ def from_path(
 
     @property
     def uri(self) -> Optional[str]:
-        return self.get("uri")
+        return self.get("uri") or self.get("url")
```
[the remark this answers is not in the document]
pdames: Yeah. I noticed that we were inefficiently storing both "url" and "uri" on disk with our old approach of copying the value of uri over to url or vice-versa during construction. This achieves the same end result of retrieving whichever is actually set (while preferring what was actually requested, in the corner case that somehow both get set) without storing a redundant path for every manifest entry on disk.

deltacat/tests/compute/test_janitor.py
pdames: Agreed - this will make the test more resilient to potential system clock skew on different hosts (although we can't go too far back in time, since we now have sanity checks making sure that the start/end transaction times look like a reasonable recent date since someone may have conceivably started using DeltaCAT, so I've fixed them to a starting point of January 1, 2025).

deltacat/utils/filesystem.py
```diff
@@ -242,13 +250,229 @@ def list_directory(
...
+        if relative.startswith("/"):
+            relative = relative[1:]
         if any(relative.startswith(prefix) for prefix in exclude_prefixes):
             continue
         out.append((file_path, file_.size))
     # We sort the paths to guarantee a stable order.
     return sorted(out)
 
 
+def list_directory_partitioned(
```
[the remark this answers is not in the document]
pdames: Hmm... although only the partition end range is required, it also takes an optional partition start range value. Both of these values can also be set in the future or the past, so it's not strictly prior files. I think I'd prefer to keep this name (or perhaps think about a different name entirely) to separate how we currently use it in practice (i.e., to retrieve metafile revisions prior to some revision number, or transactions prior to some snapshot time) vs. how it can theoretically be used (to retrieve arbitrary files between arbitrary partition start/end values, given any custom directory partitioning scheme).

PR #603, opened 2026-04-10
PlacementGroupManager now supports specifying custom_resources

pfaraone:
## Summary

Add `custom_resources` parameter to `PlacementGroupManager` and the `_config` remote function to support targeting specific Ray node types via custom resource scheduling.

## Rationale

When a Ray cluster has multiple worker node types with different instance sizes (e.g., large compute workers and smaller storage workers), CPU count alone is insufficient to guarantee placement group bundles land on the intended node type. Ray's custom resource scheduling allows declaring arbitrary resources (e.g., `storage_worker: 1`) on specific node types and requesting them in placement group bundles. This change enables callers to pass custom resources that get merged into each bundle, giving deterministic control over node type targeting.

## Changes

- Added optional `custom_resources: Optional[Dict[str, float]]` parameter to `PlacementGroupManager.__init__`
- Threaded `custom_resources` through to the `_config` Ray remote function
- `_config` merges custom resources into each bundle: `{"CPU": cpu_per_node, **(custom_resources or {})}`
- Updated `PlacementGroupManager` docstring with custom resource usage example and corrected arg descriptions
- Added unit tests for bundle construction with and without custom resources

## Impact

Fully backward compatible. The new parameter defaults to `None`, which preserves existing bundle behavior (`{"CPU": cpu_per_node}` only). No existing callers are affected.

## Testing

- Added integration tests via `PlacementGroupManager` constructor with `custom_resources` and `None` (Ray local mode)
- Added unit tests mocking Ray placement group APIs to verify bundle construction:
  - Single bundle with custom resources includes both CPU and custom resource
  - Single bundle without custom resources contains only CPU (backward compat)
  - Multiple bundles each receive the custom resources independently

## Regression Risk

Low. The change is additive — a new optional parameter with a `None` default. The bundle construction path only diverges when `custom_resources` is explicitly provided. Existing callers pass no value, so they hit the same code path as before.

## Checklist

## Additional Notes

E2E validation will be performed after upstream callers are updated to pass custom resources and the autoscaler config declares the corresponding resources on the target node types.

### DIFF
--- a/deltacat/tests/utils/test_placement.py
+++ b/deltacat/tests/utils/test_placement.py
@@ -1,7 +1,9 @@
 import unittest
+from unittest.mock import patch, MagicMock
 import ray
 from deltacat.utils.placement import (
     PlacementGroupManager,
+    _config,
     _get_available_resources_per_node,
 )
 
@@ -23,3 +25,15 @@ class TestPlacementGroupManager(unittest.TestCase):
         result = _get_available_resources_per_node()
 
         self.assertIsNotNone(result)
+
+    def test_placement_group_manager_accepts_custom_resources(self):
+
+        pgm = PlacementGroupManager(1, 1, 1, custom_resources={"storage_worker": 1})
+
+        self.assertIsNotNone(pgm)
+
+    def test_placement_group_manager_none_custom_resources(self):
+
+        pgm = PlacementGroupManager(1, 1, 1, custom_resources=None)
+
+        self.assertIsNotNone(pgm)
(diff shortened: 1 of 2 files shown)


Zyiqin-Miranda:
Do we need to modify the [cluster_resource](https://github.com/ray-project/deltacat/blob/81f3f65624327b0494cee478b7b659698d2e8866/deltacat/utils/placement.py#L323) accordingly to include the `custom_resources`?

pfaraone:
> Do we need to modify the [cluster_resource](https://github.com/ray-project/deltacat/blob/81f3f65624327b0494cee478b7b659698d2e8866/deltacat/utils/placement.py#L323) accordingly to include the `custom_resources`?

Modified accordingly

Zyiqin-Miranda (approved):
LGTM! Feel free to merge after linter is passing.

yankevn (approved):
LGTM! Please fix linting

PR #607, opened 2026-04-15
Handle compacting an empty delta

yankevn:
## Summary

This change handles the case where the delta results in no materialize results. In this case, committing the delta will be skipped in favor of directly writing back the RCF.

## Rationale

Empty deltas may occur, which will result in an assertion error during the merge deltas step while processing the merge results.

## Changes

## Impact

## Testing

## Regression Risk

## Checklist

## Additional Notes

### DIFF
--- a/deltacat/compute/compactor_v2/private/compaction_utils.py
+++ b/deltacat/compute/compactor_v2/private/compaction_utils.py
@@ -350,9 +350,10 @@ def _run_hash_and_merge(
     mutable_compaction_audit.set_records_deduped(total_dd_record_count.item())
     mutable_compaction_audit.set_records_deleted(total_deleted_record_count.item())
     record_info_msg: str = (
-        f"Hash bucket records: {total_hb_record_count},"
-        f" Deduped records: {total_dd_record_count}, "
-        f" Deleted records: {total_deleted_record_count}, "
+        f"Hash bucket records: {total_hb_record_count}, "
+        f"Input records: {total_input_records_count}, "
+        f"Deduped records: {total_dd_record_count}, "
+        f"Deleted records: {total_deleted_record_count}."
     )
     logger.info(record_info_msg)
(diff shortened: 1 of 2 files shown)


deltacat/compute/compactor_v2/private/compaction_utils.py
```diff
@@ -602,12 +605,26 @@ def _process_merge_results(
...
+            deltas,
+            stream_position=params.last_stream_position_to_compact,
+        )
+    else:
+        logger.info("No deltas found to merge, returning an empty delta...")
+        empty_manifest = Manifest.of(ManifestEntryList())
+        merged_delta: Delta = Delta.of(
+            DeltaLocator.of(
+                compacted_partition.locator,
+                None,
```
Zyiqin-Miranda: Is passing `None` as stream_position intended? Can you add test case assert follow-up incremental compaction can pick up this `None` stream position and conduct incremental starting from new delta correctly?

Zyiqin-Miranda (commented):
On a high-level, any particular reason (I assume just for minimal changes) to commit an empty delta with empty manifests?

deltacat/compute/compactor_v2/private/compaction_utils.py
yankevn: That was intended, but I'm realizing now that I got that from the sungate implementation which we shouldn't rely on in here, so I'll change it to use `current_time_ms()` instead.

yankevn:
Yeah, committing an empty delta means that at least there's something committed vs an empty staged partition. When I initially tested by just skipping the committed delta, the follow-up job failed, which is why I added the empty delta.

Zyiqin-Miranda (approved):
Looks good, maybe good idea to add empty->incremental, incremental->empty->incremental test cases to guard the behavior, especially around stream_position/rcf fetching/deriving.

